PR #7, opened 2015-09-29
Skip database 'hcatalog' for physical database operations

jimmida:
\* Code review and testing has already happened for the fix, parking the changes here temporarily until I get commit rights

"hcatalog" as a database object is created solely for the purpose of referencing externally managed data, and it does not actually contain any tables, therefore we should skip the virtual database when calculating dbsize and truncating relation node.


jimmida:
Need to update first commit from HAWQ-13 to HAWQ-18

shivzone:
Looks good.

jimmida:
@cwelton we are aware of the hawq_toolkit view and it is tracked by https://issues.apache.org/jira/browse/HAWQ-10. Lirong mentioned he already has the fix in his local repo and will get to it.

Pushed to git in Apache

PR #8, opened 2015-10-02
Hawq 5

hornn:
Update PXF to use HDP 2.3 components.


shivzone:
Nice catch with the closeForRead in ReadBridge

shivzone:
+1 on the updates to support new version of hdfs,hbase and hive

PR #9, opened 2015-10-08
HAWQ-26. Check cancel signal when Dispatcher waiting for job finish

jianlirong:
Add code to check interrupts when dispatcher waiting job finish.


wangzw:
good catch! +1

huor:
Seems good. +1

PR #11, opened 2015-10-09
HAWQ-23. Fixed building gpfdist error

liming01:


wangzw:
Please enable test case "errortbl" in install check good. It was disabled when gpfdist was disabled.

wangzw:
Others are good to me.

radarwave:
+1

PR #12, opened 2015-10-09
HAWQ-33. Change initdb log location

radarwave:
So it can avoid giving write privileges to gpadmin user on parent hawq data dir.


huor:
Seems good. +1

liming01:
+1.

PR #13, opened 2015-10-10
HAWQ-10. Skip hcatalog database when calculating the sizes of databases

jianlirong:
Update the gp_toolkit.sql.in to skip hcatalog database when calculating the sizes of database.

### DIFF
--- a/src/backend/catalog/gp_toolkit.sql.in
+++ b/src/backend/catalog/gp_toolkit.sql.in
@@ -1768,7 +1768,8 @@ AS
 	WHERE
 		datname <> 'template0' 
 	AND datname <> 'template1'
-	AND datname <> 'postgres';
+	AND datname <> 'postgres'
+	AND datname <> 'hcatalog';
 		
 GRANT SELECT ON TABLE %%JETPACK_PREFIX%%size_of_database TO public;
 


huor:
Looks good. +1

liming01:
+1. Looks good for this defect.

jimmida:
@jianlirong You should also have an upgrade path for existing customer using older HAWQ versions that does not have "hcatalog". I think the file you changed is only run during initdb?

liming01:
@jimmida 
I haven't started to think about 'hcatalog' feature's impact for gpmigrator. Some points need to be thinked:
1) If old db instance already has db named 'hcatalog':  maybe we should rename it?
2) Add a new meta for 'hcatalog' in pg_database: Done.
3) Skipping 'hcatalog'  for any db loop processing. Because this affected by the solution for point 1), so I haven't fixed it. However this behavior works fine in this version's upgrade, it will have problem when next version's upgrade.

But for this defect, this fix is okay, because the function PerformPostUpgrade() in gpMigratorUtil.py calls gp_toolkit.sql.

GodenYao:
@liming01 - we have a story to cover the upgrade scenario for Hcatalog.

jimmida:
@liming01 really good points for (1). I'm not sure what versions you are referring to in (3). What I was getting at was that for users upgrading from 1.3.x to 2.x, the hawq_toolkit view should also be CREATE OR REPLACE'd. But if we have a bigger story for upgrade c.f. @GodenYao, we can be more confident on the use case.

PR #15, opened 2015-10-10
HAWQ-35. Handle a case when a large tuple across split boundary

jianlirong:
For large tuples in a randomly distributed table, if one large tuple is across the split boundary, then the execution would fail and generate core.

The fix here is to handle this specific case. More explanations please read the comments.


liming01:
Looks fine.

huor:
The fix looks good. +1

PR #23, opened 2015-10-15
HAWQ-62. Incorrect calculation of locus describing of hash table intr…

zhangh43:
…oduce core dump in OLAP query


huor:
Below two changes look good. +1
1. In range relations, converting all hash relations instead of only the first hash relation to random distributed relation
2. Correct the calculation of locus description of hash table when there is no resource allocated for it

jianlirong:
+1. The modifications look good to me.

PR #26, opened 2015-10-19
HAWQ-67. Wrong variable type for saving fgetc() return value

jiny2:
This is to fix a bug of processing return value of fgetc(). char variable may cut the return value, int is utilized now.


yaoj2:
Looks good to me. +1

jianlirong:
+1, looks good.

PR #29, opened 2015-10-19
HAWQ-68. Wrong answer when converting hash to random partition table

zhangh43:


jianlirong:
As the JIRA description suggested, the strategy of this fix is to let the partition child table with the largest data size decide whether all other partition child tables should be transformed from hash to random.

According to this definition, this fix is OK to me. +1.

wangzw:
Code is good to me. +1

PR #31, opened 2015-10-20
HAWQ-69. Fix HAWQ init issue when HDFS HA enabled.

radarwave:
This line is useless and will make hawq init error in some cases. So just delete it. As test, after the change, hawq init goes well.


huor:
Removal of dfs_namenode, dfs_namenode_port, and dfs_space loos good. +1
1. In non-HA mode, dfs_url is good enough, so we don't need dfs_namenode, dfs_namenode_port, and dfs_space.
2. In HA mode, we don't need dfs_namenode, dfs_namenode_port, and dfs_space as well.

PR #32, opened 2015-10-21
HAWQ-46. Change data type for Metadata Versioning versions from int64 to uint64

armenatzoglou:
Currently, metadata versioning versions are uint64. However, during version generation, we use an atomic operation that returns int64 and then cast it to unsigned: 

uint64 
mdver_next_global_version()
{ 
       return (uint64) gp_atomic_add_64(mdver_global_version_counter, 1); 
}

But, this may cause an integer overflow. 

Now we have the atomic operations gp_atomic_add_int64 (former gp_atomic_add_64) and gp_atomic_add_uint64, for additions of int64 and unsigned int64 values, respectively. 
All places that handle versions use the gp_atomic_add_uint64 function. 

Moreover, I have created unit tests, which run on both OS X and Linux.

### DIFF
--- a/src/backend/utils/cache/sharedcache.c
+++ b/src/backend/utils/cache/sharedcache.c
@@ -1104,7 +1104,7 @@ void
 Cache_UpdatePerfCounter64(int64 *counter, int64 delta)
 {
 	Assert(counter + delta >= 0);
-	gp_atomic_add_64(counter, delta);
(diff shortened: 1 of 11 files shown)


gcaragea:
Looks good! 
+1

karthijrk:
+1

PR #35, opened 2015-10-23
HAWQ-72. Introduce ADDON_DIR macro in configure/makefile to allow customization

jianlirong:
1. Introduce ADDON_DIR macro in configure/makefile to make users customize Apache HAWQ easily without major modification of the code base, such as adding some proprietary features to override the built-in ones.
2. Provide an example showing how to achieve (1): override a built-in compression algorithm.


liming01:
+1, looks good.

huor:
The improvement looks good. The example using quicklz to show the usage of ADDON_DIR is great. +1

PR #37, opened 2015-10-23
HAWQ-78. Fixed parallel build error(using "make -j8")

liming01:
The error is caused by the dependency not set correctly in Makefile. 

Note: This fix only correct 2 occurrences. If we encounter the same problem in the future,  we need to fix them using same solution. Because there are a lot of Makefile, I don't look through all of them.


wangzw:
looks good. +1

changleicn:
Good fix.

PR #38, opened 2015-10-26
HAWQ-83. Rename resource manager's GUC names in hawq-site.xml.

radarwave:
Since we have changed the GUC design for RM. Now we need to update the related GUC names in hawq-site.xml.
For example:
hawq_resourcemanager_server_type --> hawq_global_rm_type.


yaoj2:
+1 look good to me since it need to comparable with code.

PR #40, opened 2015-10-26
HAWQ-73. Fix bug of 'Fail to unregister in HAWQ resource manager beca…

jiny2:
…use of RPC error'

This bug is caused by wrong self-built sql statement. 

postgres=# create resource queue test_queue_1 with (parent='pg_root', memory_limit_cluster=50%, core_limit_cluster=50%);
CREATE QUEUE
postgres=# alter resource queue test_queue_1 with(memory_limit_cluster=40%, core_limit_cluster=40%);
ALTER QUEUE
postgres=# drop resource queue test_queue_1;
DROP QUEUE
postgres=# \q
jiny2:hawqmain yjin$


huor:
The fix is good to me. +1

liming01:
+1. Looks good.

PR #41, opened 2015-10-26
[HAWQ-55] Adding dependency between optimizer_release_mdcache and gp_metadata_version

karthijrk:
Adding dependency between optimizer_release_mdcache and gp_metadata_version guc. Error out when both guc are set to off which is invalid state


gcaragea:
Karthi, I think you need to specify the jira number in the format [HAWQ-51] no spaces etc for the pull request to be automatically linked to the jira. Otherwise, there's no notification in the jira.

gcaragea:
+1
Looks good.

karthijrk:
@gcaragea  Updated the title :)

armenatzoglou:
+1

PR #42, opened 2015-10-26
HAWQ-80. Support dynamic relation distribution type in gpopt

entong:
If there are 3 hash distributed tables in a query, each has a different buckets number, say 8, 24 and 64, the resource negotiator will allocate max(8,24,64)=64 vSegs. In this case, the tables with buckets number 8 and 24 will need to be regarded as random distributed table during planning.


zhangh43:
remove the forbid_optimizer and enable hash to random in gpopt. 
+1

entong:
Merged. Closing this PR.

PR #43, opened 2015-10-26
HAWQ-81. Make resource manager to reject resource request when there …

jiny2:
…are too many segments unavailable.

This improvement makes HAWQ RM able to reject resource request directly if there are too many segments unavailable. The expect cluster size is read from $GPHOME/etc/slaves. The new guc name is hawq_rm_rejectrequest_nseg_limit. The default value is 2, i.e. if there are more than 2 segments unavailable, HAWQ RM rejects the request directly. 

The reject error message looks as below.

postgres=# insert into t values(1);                                                                  
ERROR:  failed to acquire resource because of too many unavailable segments. (pquery.c:798)


yaoj2:
+! since limited segment will impact performance

zhangh43:
+1 when more than two nodes are down, rm will reject  resource request.

PR #44, opened 2015-10-27
HAWQ-48. Change warning log message to display correct numbers.

jiny2:
Change the log message to display correct cluster size and segment size in allocated resource.


yaoj2:
+1 for more clear warning information

wangzw:
looks good. +1

PR #45, opened 2015-10-27
HAWQ-36. Allow querying pg_locks

jianlirong:
After eliminating the difference between writer gang and reader gang, we only need the status of locks in the master now.


liming01:
Looks good. +1.

PR #48, opened 2015-10-27
HAWQ-87. Make install packages of pljava and pgcrypto.

radarwave:
Fix the makefiles so we can get gppkg installers of pljava and pgcrypto.


wangzw:
looks good. +1

yaoj2:
+1. looks good

PR #49, opened 2015-10-28
HAWQ-88. set default value of guc hash_to_random_flag to  be ENFORCE_…

zhangh43:
…KEEP_HASH
guc hash_to_random_flag has three values: HASH_TO_RANDOM_BASEDON_DATALOCALITY, ENFORCE_HASH_TO_RANDOM and ENFORCE_KEEP_HASH.
It determines the different behaviour of whether to convert a hash table to random when vseg number equals to table bucket number(since when vseg num doesn't equal to bucket number, it must be converted to random)
We now set this guc to ENFORCE_KEEP_HASH by default.


ztao1987:
+1

yaoj2:
According to test result, it should be use hash as default. +1

PR #51, opened 2015-10-28
HAWQ-90. Fixed gpmigrator crash caused by pg_resqueue table change

liming01:
This bug is caused by the fix for HAWQ-25 and HAWQ-83.


jiny2:
+1 The change looks good to me. Thank you

radarwave:
Looks good. +1

PR #52, opened 2015-10-28
HAWQ-92. Add a copy function for SegFileSplitMapNode

jianlirong:


huor:
The addition of copy function for SegFileSplitMapNode looks good. +1

PR #53, opened 2015-10-28
HAWQ-93. Remove quicklz test case of InputFormat

jianlirong:


yaoj2:
+1 for quicklz is not suported

radarwave:
Looks good to me.

PR #54, opened 2015-10-29
HAWQ-104. HAWQ RM resource request timeout mechanism does not work

jiny2:
When yarn is not available or yarn has no enough resource to support query running, the resource request of that query should be timed out based on design. User should use guc to control the timeout value.
This fix is to make this work as before.


jianlirong:
+1, looks good to me.

amyrazz44:
looks good to me . +1

PR #56, opened 2015-10-29
HAWQ-108. Allocated memory space for saving preferred host informatio…

jiny2:
…n is not freed

This is a memory leak bug.

When YARN is not available, in resource broker, the temporarily allocated space is not freed.


huor:
The fix looks good. +1

yaoj2:
+1 . good for me.

PR #57, opened 2015-10-29
HAWQ-82. Allocate query resource for EXPLAIN statement

jianlirong:


zhangh43:
set query type of explain statement as QRL_ONCE to allocate resource for it.
+1

huor:
The fix looks good to allocate resource for explain. +1

entong:
Hi Lirong, does it mean we can support hash to random conversion in 'explain' now?

jianlirong:
Yes, I think so. Please have a try, and let me know if anything wrong.

PR #59, opened 2015-10-29
HAWQ-109. Fix exit code of hawq command line tools.

radarwave:
If hawq command fails to perform a specific operation, we should ensure that the return code is non-zero.

Already ran sanity and management tools for this change.

### DIFF
--- a/tools/bin/hawq
+++ b/tools/bin/hawq
@@ -6,6 +6,7 @@ import os
 import sys
 import subprocess
 from hawqpylib.HAWQ_HELP import *
+from hawqpylib.hawqlib import exit_with_return_code
 
 global hawq_home, source_hawq_env
 
@@ -19,13 +20,6 @@ def local_ssh(cmd):
     '''Execute shell command on local machine.'''
     result = subprocess.Popen(cmd, shell=True).wait()
     return result
-
-def check_error_code(result):
-    '''Check shell command exit code.'''
-    if result != 0:
-        sys.exit(1)
-    else:
-        sys.exit(0)
 
 def main():
     cluster_type_list = ('cluster', 'segment', 'master', 'standby', 'allsegments')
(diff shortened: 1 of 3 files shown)


tools/bin/hawq_ctl
```diff
@@ -886,24 +897,47 @@ def hawq_activate_standby(opts, hawq_dict):
...
-        local_ssh(cmd)
+        result = local_ssh(cmd)
+        if result != 0:
+            logger.error("Stop cluster failed, exit.")
+            sys.exit(1)
     cmd = "%s; hawq config -c hawq_master_address_host -v %s --skipvalidation -q" % (source_hawq_env, hawq_dict['hawq_standby_address_host'])
-    local_ssh(cmd)
+    result = local_ssh(cmd)
+    if result != 0:
+        logger.error("Set hawq_master_address_host  failed.")
```
hornn: extra space before "failed"

tools/bin/hawq
```diff
@@ -51,71 +57,69 @@ def main():
...
         cmd = "%s; hawq_ctl %s %s" % (source_hawq_env, hawq_command, sub_args)
-        local_ssh(cmd)
+        result = local_ssh(cmd)
     elif hawq_command == "restart":
         if second_arg not in cluster_type_list:
             print START_HELP
             sys.exit(1)
         cmd = "%s; hawq_ctl stop %s" % (source_hawq_env, sub_args)
-        local_ssh(cmd)
+        result = local_ssh(cmd)
```
hornn: In restart, if stop fails I think start should not be performed.
I can open a separate JIRA if you think it's not relevant to this one.

liming01:
Looks fine. +1.

tools/bin/hawq
radarwave: Thanks Noa, for this one I intend to do so, I want restart can continue if some of the segments already been shutdown before this restart. Sure we can have another Jira to track and discuss of such behavior.

PR #62, opened 2015-10-29
HAWQ-45. PXF Package Namespace change

shivzone:
This change will go hand in hand with other dependent changes listed below
1. Update to the internal automation tests
2. Update to internal cluster tools (singlecluster)
3. Doc updates


pxf/build.gradle
```diff
@@ -154,7 +154,7 @@ project('pxf-service') {
         }
 
         from("src/main/resources") {
-            into("/etc/gphd/pxf-${project.version}/conf")
+            into("/etc/pxf-${project.version}/conf")
```
hornn: good catch

hornn:
+118
Just a note - a change like that is a regression, because tables defined with the old fully qualified plugins names (e.g. com.pivotal.pxf.plugins.hdfs.StringPassResolver) won't work. This should probably added to release notes.

shivzone:
@hornn I completely agree. I think Goden created a task for it.
Besides, all of our documents will need to be updated along with this change.

pxf/pxf-hdfs/src/test/java/org/apache/hawq/pxf/plugins/hdfs/ChunkReaderTest.java
```diff
@@ -1,18 +1,15 @@
...
-import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mockito;
 import org.mockito.stubbing.*;
 import org.mockito.invocation.*;
-import org.powermock.api.mockito.PowerMockito;
 import org.powermock.core.classloader.annotations.PrepareForTest;
-import org.powermock.core.classloader.annotations.SuppressStaticInitializationFor;
```
GodenYao: I assume you did change through IDE - any reason IDE would remove these name spaces with this refactory?

hornn:
@shivzone Perhaps we should have a user friendly error message in case the plugin in question starts with "com.pivotal.pxf", recommending to reconstruct the table with a plugin that starts with "org.apache.hawq.pxf". We can put it in the code where the plugin names are instantiated (Utilities.createAnyInstance)

pxf/pxf-hdfs/src/test/java/org/apache/hawq/pxf/plugins/hdfs/ChunkReaderTest.java
shivzone: Those were unused imports.

shivzone:
@hornn Will do this post namespace change. The goal of the first change is to make sure we are feature parity including automation tests

hornn:
@shivzone Perhaps we should have a user friendly error message in case the plugin in question starts with "com.pivotal.pxf", recommending to reconstruct the table with a plugin that starts with "org.apache.hawq.pxf". We can put it in the code where the plugin names are instantiated (Utilities.createAnyInstance)

PR #64, opened 2015-10-30
HAWQ-86. Fix and re-enable unit test for pxf and external storage

huor:
Please review and comment.


src/backend/access/external/test/hd_work_mgr_distribute_work_2_gp_segments_test.c
```diff
@@ -243,18 +244,24 @@ static void test__distribute_work_to_gp_segments(TestInputData *input)
 	buildCdbComponentDatabases(total_segs, array_of_segs, array_of_primaries);	
 	if (enable_print_input_segments)
 		print_segments_list();
-		
-	/* 5. The actual unitest of distribute_work_2_gp_segments() */
+
+    /* 5. Build QueryResource */
+    buildQueryResource(num_hosts_in_cluster*num_segments_on_host, array_of_segs);
```
hornn: indentation seems off

src/backend/access/external/test/hd_work_mgr_mock.c
```diff
@@ -51,3 +51,49 @@ void restoreCdbComponentDatabases()
...
+		segment->alive = true;
+
+		resource->segments = lappend(resource->segments, segment);
+	}
+
+	return resource;
+}
+
+/* Restores the QueryResource for a query */
+void restoreQueryResource()
```
hornn: restore* is used when the previous state is restored (relevant in case the variable is static).
In this case, I think free* is better.

src/backend/access/external/test/hd_work_mgr_mock.c
```diff
@@ -51,3 +51,49 @@ void restoreCdbComponentDatabases()
...
+				ListCell *tmp = cell;
+				cell = lnext(cell);
+
+				pfree(((Segment *)lfirst(tmp))->hostip);
+				pfree(lfirst(tmp));
+				pfree(tmp);
+			}
+		}
+
+		pfree(resource);
```
hornn: add resource = NULL?

src/backend/access/external/test_discard/.gitignore
```diff
@@ -1 +0,0 @@
-*.t
```
hornn: should probably be added to the test directory

hornn:
+1. Thanks for fixing these tests.

src/backend/access/external/test_discard/.gitignore
huor: Yes, this is a good idea.

src/backend/access/external/test/hd_work_mgr_mock.c
huor: Agree. free* is better than restore* here for queryResoruce.

src/backend/access/external/test/hd_work_mgr_distribute_work_2_gp_segments_test.c
huor: Yes. I guess it is because of mixed use of blanks and tabs.

src/backend/access/external/test/hd_work_mgr_mock.c
huor: Right. It is a good practice to set resource to NULL since it is a global variable. This will prevent it becomes a dangling pointer.

jianlirong:
+1. Looks good to me.

PR #65, opened 2015-10-30
HAWQ-113. Fix NONE mode resource broker minus more pending resource w…

linwen:
Please review


jiny2:
+1 The change is good to me.

huor:
The fix look good. +1

PR #66, opened 2015-10-30
HAWQ-114. Set resource queue default parallel limit to 20

jiny2:
This is to set default active_statements value to 20


yaoj2:
+1

zhangh43:
+1

PR #68, opened 2015-11-02
HAWQ-116. Default active_statements setting blocks the goh_portals test

jiny2:
The parallel count is too small to run this test.


huor:
The fix looks good to me. +1

yaoj2:
+1

PR #69, opened 2015-11-03
HAWQ-118. Hawq treats commented lines as hosts in slaves file

jiny2:
slaves:
sdw1
sdw2
#sdw3
sdw4 # segment
We should get a host list like: [sdw1, sdw2, sdw4].


jiny2:
This is a wrong pull request, I will create another one.

PR #70, opened 2015-11-03
HAWQ-118. Hawq treats commented lines as hosts in slaves file

jiny2:
Now hawq treats commented lines in slaves file as valid hosts.
We should ignore the lines start with "#", and the parts after a "#".
Example:
slaves:
sdw1
sdw2
#sdw3
sdw4 # segment
We should get a host list like: [sdw1, sdw2, sdw4].


jiny2:
I think there must be something wrong, the pull request contains additional commits.

PR #71, opened 2015-11-03
Hawq-119. Log gpcheckhdfs output and Suppress unwanted messages

radarwave:
Enhance local_ssh to support more logs. Remove some useless logs.


hawqcore:
The fix looks good to me. +1

huor:
The fix looks good. +1

liming01:
Looks good. +1

PR #75, opened 2015-11-05
HAWQ-117. Add query dispatching context checking before serialization

jianlirong:


src/backend/cdb/cdbcopy.c
```diff
@@ -207,7 +207,7 @@ cdbCopyStart(CdbCopy *c, char *copyCmd, Oid relid, Oid relerror, List *err_aoseg
 		((CopyStmt *)q->utilityStmt)->scantable_splits = scantable_splits;
 	}
 
-    CloseQueryContextInfo(q->contextdisp);
+  FinalizeQueryContextInfo(q->contextdisp);
```
huor: The indention needs to followed here.

huor:
The fix looks good to me. It would be better if the minor revision regarding indention can be fixed. +1

liming01:
Looks good. +1

PR #77, opened 2015-11-05
HAWQ-128. Cleanup pg_resqueue impact in catalog and dump correspondin…

jiny2:
1) Make pg_dumpall able to dump resource queue DDL statments. Support building tree structured resource queues from pg_dumpall output.

2) pg_stat_resqueues() is expired, pg_resqueue_status can provide the same information.

postgres=# select * from pg_resqueue_status;
  rsqname   | segmem | segcore  | segsize | segsizemax | inusemem | inusecore | rsqholders | rsqwaiters | paused
------------+--------+----------+---------+------------+----------+-----------+------------+------------+--------
 pg_root    | 128    | 0.062500 | 64      | 64         | 0        | 0.000000  | 0          | 0          | F
 pg_default | 128    | 0.062500 | 32      | 64         | 0        | 0.000000  | 0          | 0          | F
(2 rows)

3) Removed some legacy code referencing pg_resourcetype and pg_resqueuecapability.

### DIFF
--- a/src/backend/catalog/caql/caqlaccess.c
+++ b/src/backend/catalog/caql/caqlaccess.c
@@ -939,15 +939,6 @@ caql_getattr(cqContext *pCtx, AttrNumber attnum, bool *isnull)
 	Assert(HeapTupleIsValid(pCtx->cq_lasttup));
 
 	return caql_getattr_internal(pCtx, pCtx->cq_lasttup, attnum, isnull);
-
-	/*
-	 * NOTE: could this be used if caql is extended to support joins, eg
-	 * what would attnum be for
-	 * "SELECT * FROM pg_resqueue, pg_resqueuecapability ..." ?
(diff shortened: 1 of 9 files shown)


liming01:
Hi Yi, 

I think it is better to keep the name of pg_stat_resqueues() unchanged. Otherwise it is hard for user or DBA to keep legacy application/doc/tools working. Thanks.

jiny2:
pg_resqueue_status is another existing UDF/view as well. I think pg_stat_resqueues is redundant.

liming01:
Hi Yi,

The purposes for these 2 views pg_catalog.pg_stat_resqueues and gp_toolkit.gp_resqueue_status are different. 
- The first one is used to monitor a resource queue's workload over time from the aspect of queries' execution time.
- The second one is focus on each resource queue's limitation setting(quota) and the resource has been occupied by active queries.

So you can't delete any of them.

PR #78, opened 2015-11-05
HAWQ-131. check hawq_rm_rejectrequest_nseg_limit should use > instead…

jiny2:
This is bug fix to make logic of checking hawq_rm_rejectrequest_nseg_limit keep align with logic of checking hawq_rm_rejectrequest_nseg_limit.


yaoj2:
+1

linwen:
+1

PR #82, opened 2015-11-06
HAWQ-135. Fixed coredump when explain generate_serial in optimizer off

liming01:
Need to set query as  DISPATCH_SEQUENTIAL when legacy query optimizer doesn't set it as DISPATCH_PARALLEL.


yaoj2:
+1

wangzw:
Looks good. +1

PR #83, opened 2015-11-09
HAWQ-136. Uneven resource allocation request caused queuing queries u…

jiny2:
This fix is to make resource manager always acquire GRM containers (YARN for example) to make HAWQ acquried GRM resource evenly dispatched in the whole cluster. This may cause more resource acquired from GRM(YARN etc.)

If resource manager wants more resource, it always try to level up the minimum water level of the whole cluster HAWQ acquired resource.

### DIFF
--- a/src/backend/resourcemanager/resourcemanager.c
+++ b/src/backend/resourcemanager/resourcemanager.c
@@ -2165,39 +2165,42 @@ int generateAllocRequestToBroker(void)
 					"overall water level.",
 					reqcore);
 
-		addResourceBundleData(&(mctrack->TotalPending), reqmem, reqcore);
-		uint64_t oldtime = mctrack->TotalPendingStartTime;
-		if ( mctrack->TotalPendingStartTime == 0 )
+		if ( reqcore > 0 )
 		{
-			mctrack->TotalPendingStartTime = gettime_microsec();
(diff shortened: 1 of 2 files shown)


jiny2:
There is also a fix due to this new implementation, unavailable segment is not considered when building resource request to libyarn/none resource broker.

src/backend/resourcemanager/resourcemanager.c
```diff
@@ -2177,6 +2213,149 @@ int generateAllocRequestToBroker(void)
...
+							llevel,
+							aclevel-clevel);
+			}
+
+			index++;
+		}
+	}
+
+	/* Adjust total mem and core request. */
+	*reqcore -= reqcoreleft;
```
zhangh43: +1
and just wonder is reqcoreleft must be 0 or can be negative?
and is there a circumstance that "resource manager skips this segment"(line2303) are always true for all the segments, which leads to dead loop?

jiny2: Yes, it is possible to be negative, for example, 1 GRM container is required, but in order to level up the minimum water level, 3 segments have to have 3 more GRM containers added into the request, which should cause reqcoreleft -2.

I will check infinite loop problem and fix it. Thank you!

PR #84, opened 2015-11-09
HAWQ-137. Update hdfs-client.xml and yarn-client.xml configuration.

radarwave:
Passed installcheck-good test and sanity test.


wangzw:
looks good. +1

PR #85, opened 2015-11-09
HAWQ-132. Enrich log information when queuing resource request is tim…

jiny2:
This is to enrich log information in case some request is timed out, it is useful to find the timeout reason.


yaoj2:
+1

zhangh43:
＋1

PR #87, opened 2015-11-09
HAWQ-138. Alter resource quene error when memory set 64mb

jiny2:


yaoj2:
+1

huor:
The fix looks good to me. +1

PR #88, opened 2015-11-09
HAWQ-97. Optimizing metadata versioning event generation

gcaragea:
A DDL command usually generates more than one Nuke Versioning event during its execution. All these nuke events are then pushed to Shared Versioning Queue (SVQ) at commit, and then all of them get propagated to other backends. 

This is extra work and it can be optimized. We should push only one Nuke VE instead of several.

Also, when moving messages from the Current Command Invalidation Queue (CVQ) to the Prior Command Invalidation Queue (XVQ), pre-process the event queue to eliminate any events that will have no effect. If a versioning event is followed by a nuke event, it will have no effect.
 Moreover, when a nuke event is moved to XVQ, everything prior in XVQ will  also have no effect.
 
Solution is  to look for Nuke events in CVQ. Find the last one (in the order of creation), then move all messages from then on to XVQ, overwriting XVQ.


karthijrk:
+1

armenatzoglou:
The code looks good to me. +1

foyzur:
+1

PR #90, opened 2015-11-10
HAWQ-141. memory accessing panic in system test

jiny2:
The idea of the fix is to avoid calling gethostbyname() in multi-thread env.  The required address information are prepared before creating heart-beat thread.

### DIFF
--- a/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
+++ b/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
@@ -1340,11 +1340,12 @@ void *generateResourceRefreshHeartBeat(void *arg)
 				   tharg->HostAddrLength);
 			server_addr.sin_port = htons(rm_master_port);
 
+			int sockres = 0;
 			while(true)
 			{
-				int sockres = connect(fd,
-									  (struct sockaddr *)&server_addr,
-									  sizeof(server_addr));
+				sockres = connect(fd,
+								  (struct sockaddr *)&server_addr,
+								  sizeof(server_addr));
 				if( sockres < 0)
 				{
 					if (errno == EINTR)
@@ -1360,8 +1361,13 @@ void *generateResourceRefreshHeartBeat(void *arg)
 					}
 					break;
 			    }
-
 				break;
+			}
+
+			if ( sockres < 0 )
+			{
+				pg_usleep(1000000L);
+				continue;
 			}
 
 			RMMessageHead phead = (RMMessageHead)messagehead;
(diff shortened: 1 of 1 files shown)


src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -234,12 +245,59 @@ void initializeQD2RMComm(void)
...
+    		if ( rmserver->h_addrtype == AF_INET )
+    		{
+    			tharg->HostAddrSize++;
+    		}
+    	}
+    	elog(DEBUG3, "Resolved resource manager host %s to %d INET addresses.",
+    				 master_addr_host,
+					 tharg->HostAddrSize);
+
+    	tharg->HostAddrs = malloc(sizeof(char *) * tharg->HostAddrSize);
```
zhangh43: I think in line 1433 you also need to free (*arg)->HostAddrs.
but also pay attention to line 276: if h_addrtype != AF_INET then tharg->HostAddrSize =0, which leads to  malloc(0) happens.  In this case, you should not free (*arg)->HostAddrs

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -1263,43 +1324,118 @@ void *generateResourceRefreshHeartBeat(void *arg)
...
+							  "(errno %d)", errno);
 				}
 			}
-			closeConnectionRemote(&fd);
+			else
+			{
+				write_log("ERROR generateResourceRefreshHeartBeat send error "
+						  "(errno %d)", errno);
+			}
+			close(fd);
```
zhangh43: you only wirte log when recv and send error, Does the error affect others?

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -1263,43 +1324,118 @@ void *generateResourceRefreshHeartBeat(void *arg)
...
+						close(fd);
+					}
+					break;
+			    }
+				break;
+			}
+
+			if ( sockres < 0 )
+			{
+				pg_usleep(1000000L);
```
zhangh43: logic here is sleep 1s and reconnect when connect failed?
Is there a need to add max_retry_num?

jiny2: Let me check the memory free problem. Sleep 1s is for reconnecting.

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
jiny2: The logic is to just reconnect and resend no matter currently, the message is sent with or without error.

zhangh43: got it

PR #92, opened 2015-11-10
Analyze HAWQ-44

hornn:
Advanced statistics for PXF table.

PXF sample rows are collected into a temporary table, where statistics are derived of them in the same way ANALYZE works for hawq tables.
Statistics are gathered at 3 stages:
1. Getting general statistics - number of fragments, size of data source, size of first fragment
2. Count of first fragment tuples
HAWQ uses these numbers to determine how many tuples are needed, and these parameters are translated to sampling ratio and number of sampled fragments.
3. Sampling the PXF table based on the sampling ratio and number of fragments to be sampled. The returned tuples are saved in a temporary table.

On the PXF side, a function has been made to the Fragmenter API, to allow gathering the stats of the first stage. In addition, a mechanism to sample rows on the fly was added to the Bridge.


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * FragmentsStats holds statistics for a given path.
+ */
+public class FragmentsStats {
+
+    /**
+     * Default fragment size. Assuming a fragment is equivalent to a block in
+     * HDFS, we guess a full fragment size is 64MB.
+     */
+    public static final long DEFAULT_FRAGMENT_SIZE = 67108864L;
```
shivzone: This should eventually be exposed as a configuration parameter

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
```diff
@@ -0,0 +1,226 @@
...
+        }
+
+        sizeAndUnit.unit = getSizeUnit(orderOfMagnitude);
+        return sizeAndUnit;
+    }
+
+    private SizeUnit getSizeUnit(int orderOfMagnitude) {
+        switch (orderOfMagnitude) {
+            case 0:
+                return SizeUnit.B;
```
shivzone: no break statement

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
```diff
@@ -0,0 +1,226 @@
...
+            sizeAndUnit.size /= 1024;
+            orderOfMagnitude++;
+        }
+
+        sizeAndUnit.unit = getSizeUnit(orderOfMagnitude);
+        return sizeAndUnit;
+    }
+
+    private SizeUnit getSizeUnit(int orderOfMagnitude) {
+        switch (orderOfMagnitude) {
```
shivzone: missing break for all the cases

src/backend/access/external/hd_work_mgr.c
```diff
@@ -216,51 +216,50 @@ static void assign_pxf_port_to_fragments(int remote_rest_port, List *fragments)
...
 /*
- * Fetches statistics of the PXF datasource from the PXF service
+ * Fetches fragments statistics of the PXF datasource from the PXF service
  *
- * The function will generate a delegation token when secure filesystem mode 
+ * The function will generate a delegation token when secure filesystem mode
  * is on and cancel it right after.
  */
-PxfStatsElem *get_pxf_statistics(char *uri, Relation rel, StringInfo err_msg)
+PxfFragmentStatsElem *get_pxf_fragments_statistics(char *uri, Relation rel, StringInfo err_msg)
```
entong: Many functions in this CL have err_msg as an argument, which is not very clean API. Is there a better way? Can we emit the error message at the place the error occurs?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
```diff
@@ -0,0 +1,226 @@
...
+    private long fragmentsNumber;
+    // first fragment size
+    private SizeAndUnit firstFragmentSize;
+    // total fragments size
+    private SizeAndUnit totalSize;
+
+    /**
+     * Enum to represent unit (Bytes/MB/GB/TB)
+     */
+    public enum SizeUnit {
```
shivzone: Good job with the handling of large size

src/backend/commands/analyze.c
```diff
@@ -933,13 +914,12 @@ static void analyzeRelation(Relation relation, List *lAttributeNames, bool rooto
...
+		analyzePxfEstimateReltuplesRelpages(relation, &location, &estimatedRelTuples, &estimatedRelPages, &err_msg);
 		if (err_msg.len > 0)
 		{
-			ereport(WARNING,
-					(errmsg("skipping \"%s\" --- error returned: %s",
-							RelationGetRelationName(relation),
-							err_msg.data)));
+			elog(ERROR,
+					"%s",
+					err_msg.data);
```
entong: If an error occurs, do we abort the entire analyze?

src/backend/commands/analyze.c
```diff
@@ -1436,7 +1431,7 @@ static Oid buildSampleTable(Oid relationOid,
 /**
  * Drops the sample table created during ANALYZE.
  */
-static void dropSampleTable(Oid sampleTableOid)
+void dropSampleTable(Oid sampleTableOid, bool isExternal)
```
entong: I'm confused here. The sample table we created, even for pxf, is a regular AO table. Why we use 'drop external table' here?

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/HdfsDataFragmenter.java
```diff
@@ -43,19 +46,11 @@ public HdfsDataFragmenter(InputData md) {
     @Override
     public List<Fragment> getFragments() throws Exception {
         String absoluteDataPath = HdfsUtilities.absoluteDataPath(inputData.getDataSource());
-        InputSplit[] splits = getSplits(new Path(absoluteDataPath));
+        ArrayList<InputSplit> splits = getSplits(new Path(absoluteDataPath));
```
shivzone: Use Interface List here instead of ArrayList ... just a best practice

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/BridgeOutputBuilder.java
```diff
@@ -216,10 +245,53 @@ void fillText(List<OneField> recFields) throws BadRecordException {
...
+                outputList.add(output); // TODO break output into lines
+            }
         } else { // from QuotedLineBreakAccessor
             String textRec = (String) val;
             output = new Text(textRec + "\n");
+            outputList.add(output);
+        }
+    }
+
+    void convertTextDataToLines(byte[] val) {
```
shivzone: Add a comment to this function explaining the purpose of this either here or when it is invoked for sampling

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/BridgeOutputBuilder.java
```diff
@@ -216,10 +245,53 @@ void fillText(List<OneField> recFields) throws BadRecordException {
...
+                output = new BufferWritable((byte[]) val);
+                outputList.add(output); // TODO break output into lines
+            }
         } else { // from QuotedLineBreakAccessor
             String textRec = (String) val;
             output = new Text(textRec + "\n");
+            outputList.add(output);
+        }
+    }
+
```
shivzone: Add a comment to this function explaining the purpose of this either here or when it is invoked for sampling

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ReadBridge.java
```diff
@@ -14,58 +14,81 @@
...
-    /*
-     * Fetch next object from file and turn it into a record that the GPDB backend can process
+    /**
+     * Fetches next object from file and turn it into a record that the HAWQ
+     * backend can process.
      */
     @Override
     public Writable getNext() throws Exception {
-        Writable output;
+        Writable output = null;
```
shivzone: java by default initializes objects to null

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
hornn: Agreed.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FragmentsStats.java
hornn: Will change.

src/backend/access/external/hd_work_mgr.c
hornn: I will check.
If I remember correctly, the reason for this behaviour is that we don't want to error out in case analyze on external fails, but only to issue a warning, and continue to analyze the next tables.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ReadBridge.java
```diff
@@ -14,58 +14,81 @@
...
+ * record as invalid for HAWQ.
  */
 public class ReadBridge implements Bridge {
     ReadAccessor fileAccessor = null;
     ReadResolver fieldsResolver = null;
     BridgeOutputBuilder outputBuilder = null;
+    LinkedList<Writable> outputQueue = null;
 
-    private Log Log;
+    static private Log Log = LogFactory.getLog(ReadBridge.class);
```
shivzone: best practice is to define log objects as private static final when you initialize it

src/backend/commands/analyze.c
```diff
@@ -1217,18 +1214,18 @@ static float4 estimateSampleSize(Oid relationOid, const char *attributeName, flo
...
-static void spiExecuteWithCallback(
+void spiExecuteWithCallback(
 		const char *src,
 		bool read_only,
 		long tcount,
 		spiCallback callbackFn,
 		void *clientData)
 {
-	bool connected = false;
+	volatile bool connected = false; /* needs to be volatile when accessed by PG_CATCH */
```
entong: thanks!

src/backend/commands/analyze.c
```diff
@@ -1812,9 +1807,9 @@ static float4 analyzeComputeNRepeating(Oid relationOid,
 
 	initStringInfo(&str);
 	appendStringInfo(&str, "select count(v)::float4 from (select Ta.%s as v, count(Ta.%s) as f from %s.%s as Ta group by Ta.%s) as foo where f > 1",
-			quote_identifier(attributeName), 
 			quote_identifier(attributeName),
-			quote_identifier(sampleSchemaName), 
+			quote_identifier(attributeName),
+            quote_identifier(sampleSchemaName),
```
entong: indent

src/backend/access/external/hd_work_mgr.c
hornn: Removed all special error handling. Now if something goes wrong, an error is emitted.
The former behaviour was required for two reasons:
1. ANALYZE on all tables was done as one transaction, but now that we introduced subtransaction per table, it is not needed anymore. 
2. PXF had an implicit ANALYZE run from the planner if no statistics were gathered before. In that case, an error had to be suppressed, which was done by writing the error into a string and ignoring it if necessary. We removed this behaviour as part of this feature, so there is no need to suppress errors anymore.

src/backend/commands/analyze.c
hornn: No, only analyze for this table is aborted - each analyzeRelation is encapsulated inside a subtransaction.

src/backend/commands/analyze.c
hornn: We create several sample tables - one is a PXF table that is used to get the samples, and another is a AO table where the samples are stored.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/HdfsDataFragmenter.java
hornn: Done. Thanks

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/BridgeOutputBuilder.java
hornn: Good point. Added documentation to convertTextDataToLines()

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ReadBridge.java
hornn: http://stackoverflow.com/a/16699632
Apparently this rule excludes local variables. They cannot be returned unless they were initialized.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ReadBridge.java
hornn: Good point. 
Also created JIRA to change it all over the code https://issues.apache.org/jira/browse/HAWQ-165

src/backend/commands/analyze.c
hornn: done

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ReadBridge.java
shivzone: Aah ... you are right

shivzone:
+1

PR #93, opened 2015-11-10
HAWQ-123. Add 'Partition by' description in psql

hsyuan:
When user runs \d or \d+ on a partition table, it will display partition
keys on the current level for the partition table with the clause of
'Partition by: (partition_key_1, ..., partition_key_n)' after
'Distributed by' clause. But if user runs \d or \d+ on a non-partition
table or leaf level partition table, the 'Partition by' clause will not
be displayed.

Haisheng Yuan

### DIFF
--- a/src/bin/psql/describe.c
+++ b/src/bin/psql/describe.c
@@ -28,6 +28,7 @@ static bool describeOneTableDetails(const char *schemaname,
 						const char *oid,
 						bool verbose);
 static int add_distributed_by_footer(const char* oid, PQExpBufferData *inoutbuf, PQExpBufferData buf);
+static int add_partition_by_footer(const char* oid, PQExpBufferData *inoutbuf, PQExpBufferData *buf);
 static void add_tablespace_footer(printTableContent *const cont, char relkind,
 					  Oid tablespace, const bool newline);
 static void add_role_attribute(PQExpBuffer buf, const char *const str);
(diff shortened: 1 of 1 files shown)


entong:
Looks good +1

src/bin/psql/describe.c
```diff
@@ -2676,6 +2686,80 @@ add_distributed_by_footer(const char* oid, PQExpBufferData *inoutbuf, PQExpBuffe
...
+	result = PSQLexec(buf->data, false);
+
+	if (!result)
+		return 1;
+	int nRows = PQntuples(result);
+	int nPartKey = 0;
+
+	PQclear(result);
+
+	if(nRows)
```
hornn: missing space after if

src/bin/psql/describe.c
```diff
@@ -2676,6 +2686,80 @@ add_distributed_by_footer(const char* oid, PQExpBufferData *inoutbuf, PQExpBuffe
...
+		char *partColName;
+		int i = 0;
+		appendPQExpBuffer(inoutbuf, "Partition by: (");
+		for (i = 0; i < nPartKey; i++)
+		{
+			if (i > 0)
+				appendPQExpBuffer(inoutbuf, ", ");
+			partColName = PQgetvalue(result, i, 0);
+
+			if (!partColName)
```
hornn: if something bad happens here, maybe inoutbuf should be cleared. Otherwise a partial message will be printed (e.g. "Partition by: x, y, "). Also perhaps clear result.

hornn:
+1

src/bin/psql/describe.c
hsyuan: Yes, we need to clear the inoutbuf here, I will add it. The result is a local variable, no need to clear.
Thanks for pointing out.

src/bin/psql/describe.c
hsyuan: oops. Will add space. :smile:

atris:
+1

PR #94, opened 2015-11-10
HAWQ-145. Memory leak when free the argument instance of resource lea…

jiny2:
This is to fix a memory leak.

### DIFF
--- a/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
+++ b/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
@@ -1438,6 +1438,7 @@ void freeHeartBeatThreadArg(HeartBeatThreadArg *arg)
 	{
 		free((*arg)->HostAddrs[i]);
 	}
+	free((*arg)->HostAddrs);
 	free(*arg);
 	*arg = NULL;
 }


src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -1438,6 +1438,7 @@ void freeHeartBeatThreadArg(HeartBeatThreadArg *arg)
 	{
 		free((*arg)->HostAddrs[i]);
 	}
+	free((*arg)->HostAddrs);
```
zhangh43: have a look at the code: tharg->HostAddrs = malloc(sizeof(char *) * tharg->HostAddrSize);
where tharg->HostAddrSize maybe zero and malloc(0) happens.
what i mean yesterday is to test whether tharg->HostAddrSize > 0 before free arg->HostAddrs

zhangh43: sorry. It's OK to free(malloc(0)).
+1

linwen:
+1

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
jiny2: In the code, I have logic to prevent this case :)

PR #95, opened 2015-11-11
HAWQ-94. Handle one case when reaching EOF for AO large tuple

jianlirong:
When *blockLimitLen = 0, we should return false to indicate this is the end of file.


yaoj2:
+1 For handle blockLimitLen=0

PR #96, opened 2015-11-11
Hawq 142

linwen:
This fix is to remove fips mode. OpenSSL's FIPS mode is not stable, and not easy to maintain.


yaoj2:
The fix looks good to me

radarwave:
Looks good to me. +1

atris:
Fine for me too. +1

PR #98, opened 2015-11-11
HAWQ-146. Re-enable non-uniform partitioning DDLs.

entong:
Previously we disabled DDLs that could create non-uniform partitioning hierarchy. We are enabling those now.


hsyuan:
The code looks good to me. :+1:

entong:
Pushed

PR #99, opened 2015-11-11
HAWQ-149. Add orafce, gp_cancel_query, pgbench and extprotocol to HAWQ

foyzur:


changleicn:
looks good.

huor:
Is there any particular reason that we move pgbench from contrib to src/bin? 

Other than that, the fix looks good to me. +1

foyzur:
@huor most of the binaries are already in src/bin. So, I found it intuitive to locate pgbench under that. Besides that the make file change is much simpler for src/bin.

huor:
@foyzur That makes sense. Good to know that.

atris:
I think patch is fine. My only qualm is that moving it out of contrib and into src/bin will make merging PostgreSQL commits harder and make it a bit counter intuitive.

foyzur:
@atris are you sure that pgbench is under contrib in postgres? I just checked postgres repo and it seems pgbench is already under src/bin in postgres!

PR #100, opened 2015-11-12
HAWQ-102. Change error message when PXF webapp is down

hornn:


src/backend/access/external/libchurl.c
```diff
@@ -815,42 +815,49 @@ void check_response_code(churl_context* context)
...
+		appendStringInfo(&err, "remote component error (%ld)", response_code);
 
-			addr = get_dest_address(context->curl_handle);
-			if (strlen(addr) != 0)
-				appendStringInfo(&err, " from %s", addr);
-			pfree(addr);
+		addr = get_dest_address(context->curl_handle);
+		if (strlen(addr) != 0)
+		{
+			appendStringInfo(&err, " from %s", addr);
```
shivzone: hawq codebase seems to follow convention of not using braces on single statement action on conditional blocks
if(case)
    code

shivzone:
+1

src/backend/access/external/libchurl.c
hornn: Good point. As far as I know the convention is to add curly brackets for each block, but the code doesn't always follow that convention.
In this case, I prefer to keep it that way to be make it clearer.

shivzone: Sounds good

PR #101, opened 2015-11-12
HAWQ-152. Pass validation test in resource manager for unexactly divi…

jiny2:
unexactly divisible mem core ratio can lead to some small biases in the counters in resource pool. This code fix is to make internal validation logic tolerate this case.


yaoj2:
The fix looks good to me

PR #102, opened 2015-11-12
HAWQ-153. Remove expired code using sigar library in resource manager

jiny2:


huor:
The fix looks good me. +1

yaoj2:
+1

PR #105, opened 2015-11-13
HAWQ-156. Fix hd_work_mgr unit tests.

hornn:
The tests were using an the wrong cast to access segment info. The old struct CdbComponentDatabaseInfo should have been replaced by Segment.


shivzone:
+1. good catch

PR #106, opened 2015-11-13
HAWQ-157. Add refactored implementation of allocating vsegs in resour…

jiny2:
This is a simplified implementation of allocating vsegs from resource pool.


linwen:
looks good.

PR #107, opened 2015-11-13
HAWQ-40. Refact hawq standby init. Changes to standby int and start.

radarwave:
Refact hawq standby init. Change standby start sequence and do init directly on standby host(previously on master node).

### DIFF
--- a/tools/bin/hawq_ctl
+++ b/tools/bin/hawq_ctl
@@ -148,12 +148,12 @@ class HawqInit:
         return result
 
     def set_vsegment_num_per_node(self):
-        cmd = "%s; hawq config -c hawq_resourcemanager_query_vsegment_number_per_segment_limit  \
+        cmd = "%s; hawq config -c hawq_rm_nvseg_perquery_perseg_limit \
               -v %s --skipvalidation -q > /dev/null" % \
               (source_hawq_env, self.vseg_num_per_node)
         result = local_ssh(cmd, logger)
         if result != 0:
-            logger.warn("Set hawq_resourcemanager_query_vsegment_number_per_segment_limit failed")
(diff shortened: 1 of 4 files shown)


tools/bin/hawq_ctl
```diff
@@ -149,7 +153,7 @@ class HawqInit:
               (source_hawq_env, self.vseg_num_per_node)
         result = local_ssh(cmd, logger)
         if result != 0:
-            logger.warn("Set hawq_resourcemanager_query_vsegment_number_per_segment_limit failed.")
+            logger.warn("Set hawq_resourcemanager_query_vsegment_number_per_segment_limit failed")
         return result
```
yaoj2: hawq_resourcemanager_query_vsegment_number_per_segment_limit guc name change, you should use new guc name

tools/bin/hawq_ctl
```diff
@@ -284,18 +308,10 @@ class HawqInit:
             logger.info("Segments init successfully on nodes '%s'" % self.host_list)
 
         if self.standby_host_name not in ('', 'None', 'none', 'NONE'):
```
yaoj2: We should lower hostname first and then compare with none

yaoj2:
Looks good to me.

huor:
The fix looks good. +1

denalex:
!! Please hold off on committing this change until after the Beta. Ambari HAWQ plugin will not be able to accommodate these changes for the Beta release. !!

radarwave:
@denalex  Ok, this commit is in, I will revert it.

radarwave:
@denalex Please let me know if this change can be pushed in now since beta is released. Thanks.

vVineet:
Sure, please go ahead.

PR #108, opened 2015-11-13
HAWQ-159. Fix the bug when hawq disconnect from libyarn, heartbeat th…

linwen:
Segmentation fault happens when Hawq disconnect from libyarn. That is because heartbeat thread hasn't been canceled. Try to fix it.


changleicn:
looks good.

huor:
Looks good to me. +1

PR #109, opened 2015-11-13
HAWQ-160. Fixed "dispatcher encounter error" when cancel query

liming01:
CHECK_FOR_INTERRUPTS() can not be called between  PG_TRY(); and  PG_CATCH(); 
because the cancel request ( by passing SIGINT for ProcessInterrupts() to call elog) will be caught by the PG_CATCH(); clause, and the error will be changed to others.

So we just move this line out.


huor:
The fix looks good to me. +1

changleicn:
good fix.

PR #112, opened 2015-11-14
HAWQ-158. Remove legacy command line tools and help.

radarwave:


tools/Makefile
```diff
@@ -38,13 +38,7 @@ SET_VERSION_SCRIPTS = \
 	bin/gpssh-exkeys \
```
huor: Why not remove gpssh-exkeys as we have hawq ssh now?
Why not remove clsAddMirrors.py as we don't have mirrors in hawq?

huor:
The overall fix is good. +1
Why we have gpssh-exkeys and clsAddMirrors.py in Makefile as they are deprecated?

tools/Makefile
radarwave: gpssh-exkeys still in use. 'hawq ssh' will call it.
For others not in bin or sbin, there still plenty need to be remove, we will review and remove them in the future.

PR #113, opened 2015-11-14
HAWQ-162. YARN containers are required evenly

jiny2:
When there are 3 hosts, 8 YARN containers are required, actually, resource broker should require 3, 3, 3 containers in 3 hosts


changleicn:
looks good.

huor:
The fix is good to me. +1

PR #114, opened 2015-11-16
HAWQ-163. Null container set instance is not checked when allocating …

jiny2:
This is a fix to check null pointer when allocating vseg in resource pool.


huor:
The fix is good. +1

PR #120, opened 2015-11-17
HAWQ-167. Update HAWQ VERSION to 2.0.0.0_beta for make rpm package.

radarwave:
Previously version number "2.0.0.0-beta" will cause rpm package fail, so change '-' to '_' .


yaoj2:
+1

huor:
It is good to follow the same naming convention.

PR #121, opened 2015-11-17
HAWQ-56. Add format option to external protocol struct, and disallow …

hornn:
…HEADER format option to be used in PXF tables


shivzone:
+1

PR #122, opened 2015-11-18
HAWQ-172. Add gpoperation.py back for gppkg use.

radarwave:
gpoperation.py is still in use by gppkg. We should add it back.


yaoj2:
+1

huor:
The fix is good. +1

PR #124, opened 2015-11-19
HAWQ-173. Set parameter dfs.client.use.legacy.blockreader.local to fa…

wangzw:
Set parameter dfs.client.use.legacy.blockreader.local to false in hdfs-client.xml


yaoj2:
+1

huor:
The fix is good. +1

PR #125, opened 2015-11-19
HAWQ-174. Set value of dfs.domain.socket.path as /var/lib/hadoop-hdfs…

wangzw:
Set value of dfs.domain.socket.path as /var/lib/hadoop-hdfs/dn_socket


yaoj2:
The fix looks good

huor:
The fix is good. +1

PR #126, opened 2015-11-21
HAWQ-148. Restructuring optimizer directory

vraghavan78:
Restructuring optimizer directory and bumping up optimizer version number


entong:
+1

PR #127, opened 2015-11-23
HAWQ-177. finalize resource queue ddl error messages

jiny2:
This fix is to finalize all resource queue ddl CREATE/ALTER/DROP RESOURCE QUEUE error messages. Then they will be verified by test cases.
For the specifications of the ddl and its error messages, please check 
HAWQ 2.0 Resource Management Design (+FTS), section 4.4


linwen:
looks good.

src/backend/resourcemanager/resqueuemanager.c
```diff
@@ -1065,16 +1070,27 @@ int checkAndCompleteNewResourceQueueAttributes(DynResourceQueue  queue,
...
 			return res;
 		}
+
+		/* The parent queue cannot have roles assigned. */
+		if ( hasUserAssignedToQueue(parenttrack->QueueInfo) )
+		{
+			res = RESQUEMGR_WRONG_ATTR;
+			snprintf(errorbuf, errorbufsize,
+					 "%s has at least one role assigned",
+					 parenttrack->QueueInfo->Name);
```
yaoj2: What about 
"parent quene %s should not have roles assigned"

src/backend/resourcemanager/resqueuemanager.c
```diff
@@ -1511,9 +1521,7 @@ int createQueueAndTrack( DynResourceQueue		queue,
 			{
 				res = RESQUEMGR_WRONG_PARENT_QUEUE;
 				snprintf( errorbuf, errorbufsize,
-						  ERRORPOS_FORMAT
-						  "The parent queue of can not be pg_default.",
-						  ERRREPORTPOS);
+						  "the parent queue can not be pg_default");
```
yaoj2: it should be cannot here

yaoj2:
I suggest that the error message we should directly tell user what should be. 
For example: 
ERROR: the values of memory_limit_cluster and core_limit_cluster are not identical
->
ERROR: the values of memory_limit_cluster and core_limit_cluster should be identical

ERROR: active_statements is less than 1
->
ERROR: active_statements should be greater than 1

nvseg_upper_limit_perseg is less than nvseg_lower_limit_perseg
->
nvseg_upper_limit_perseg should be greater than nvseg_lower_limit_perseg
2. Is it any reason we remove 64MB  for vseg_resource_quota?

3. drop a queue having at least one child queue.
What about change the message to "resource queue %s has at least one chile queue."

src/backend/resourcemanager/resqueuemanager.c
jiny2: I prefer using the message which explain the error reason.

src/backend/resourcemanager/resqueuemanager.c
jiny2: Fixed Thanks.

PR #128, opened 2015-11-23
HAWQ-175. Update PXF version to 3.1.0

hornn:


shivzone:
Would be useful if we can mention the higher level changes in the jira HAWQ-175 for future reference.

shivzone:
+1

PR #136, opened 2015-11-25
Hawq 189

hornn:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
```diff
@@ -143,7 +146,8 @@ private static Response writeResponse(Bridge bridge,
         } finally {
             inputStream.close();
         }
-        returnMsg = "wrote " + totalWritten + " bulks to " + path;
+        String censuredPath = path.replaceAll("[^a-zA-Z0-9_-]", ".");
```
shivzone: this has be be done as part of a utility. We might commonly use this in other apis and possibly modify it later

shivzone:
+1

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
hornn: Done

PR #139, opened 2015-11-27
Hawq-184. Add license header

linwen:
This change will not add license header to 3rd-party codes, like pljava, postgres, pgcrypto, etc.


jiny2:
+1 The code is good to me.

huor:
The fix looks good. +1

PR #140, opened 2015-11-27
HAWQ-197. Adding a huge number of GRM containers cause socket connect…

jiny2:
This fix is to make resource manager move at most hawq_rm_container_batch_limit GRM containers. And a frequently called validation function is removed.


linwen:
looks good.

huor:
The improvement looks good. +1

PR #142, opened 2015-11-30
HAWQ-41. Add a GUC to set max open AO segemnt files

jianlirong:
The newly added GUC is max_appendonly_segfiles, which determines
the max number of appendonly relation segment files open
concurrently.

In a large cluster, we usually need to increase the value of this
GUC in order to support high concurrencies.


huor:
The fix looks good. +1

radarwave:
The fix looks good. +1

PR #144, opened 2015-12-01
HAWQ-201. Resource uneven or fragment problem may cause resource queu…

jiny2:
This fix is to let resource manager actively raise water level to try to deal with possible resource uneven, resource fragment problem.

### DIFF
--- a/src/backend/cdb/cdbvars.c
+++ b/src/backend/cdb/cdbvars.c
@@ -350,6 +350,8 @@ int		rm_resource_allocation_timeout;	/* How may seconds to wait before
 										   request. */
 int		rm_resource_timeout;		/* How many seconds to wait before returning
 									   resource back to the resource broker. */
+int		rm_request_timeoutcheck_interval;
+
 int		rm_session_lease_heartbeat_interval; /* How many seconds to wait before sending
 										   	    another heart-beat to resource manager. */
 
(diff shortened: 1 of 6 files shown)


src/backend/cdb/cdbvars.c
```diff
@@ -350,6 +350,8 @@ int		rm_resource_allocation_timeout;	/* How may seconds to wait before
 										   request. */
 int		rm_resource_timeout;		/* How many seconds to wait before returning
 									   resource back to the resource broker. */
+int		rm_request_timeoutcheck_interval;
```
huor: A description here for rm_request_timeoutcheck_interval would be great.

src/backend/resourcemanager/resourcemanager.c
```diff
@@ -2168,6 +2169,37 @@ int generateAllocRequestToBroker(void)
...
+			 PRESPOOL->GRMTotal.Core * PQUEMGR->GRMQueueMaxCapacity )
+		{
+			/*
+			 * We only add one more GRM container to acquire, this will trigger
+			 * the following logic to raise the water level.
+			 */
+			reqcore = 1;
+			reqmem = reqcore * mctrack->MemCoreRatio;
+
+			PQUEMGR->hasResourceProblem[RESPROBLEM_FRAGMENT]  = false;
```
huor: Here we are marking fragment, or uneven, or too few segments problem false, is this means that adding one more container would resolve all these three kind of issue?

huor:
The overall fix is good to me. +1

src/backend/resourcemanager/resourcemanager.c
jiny2: Yes, adding more GRM containers can deal with three kinds of issues.

src/backend/cdb/cdbvars.c
jiny2: Fixed.

PR #146, opened 2015-12-02
modify metadata cache read lock logic

wengyanqing:
modify metadata cache lock logic when check item exists

### DIFF
--- a/src/backend/cdb/cdbmetadatacache.c
+++ b/src/backend/cdb/cdbmetadatacache.c
@@ -579,7 +579,7 @@ BlockLocation *GetHdfsFileBlockLocations(const HdfsFileInfo *file_info, uint64_t
     MetadataCacheEntry *cache_entry = NULL;
     BlockLocation *locations = NULL;
 
-    LWLockAcquire(MetadataCacheLock, LW_EXCLUSIVE);
+    LWLockAcquire(MetadataCacheLock, LW_SHARED);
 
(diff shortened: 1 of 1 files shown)


linwen:
+1

src/backend/cdb/cdbmetadatacache.c
```diff
@@ -619,6 +619,9 @@ BlockLocation *GetHdfsFileBlockLocations(const HdfsFileInfo *file_info, uint64_t
             if (cache_entry->block_num <= 1)
             {
                 // only one file, re-fetch 
+                LWLockRelease(MetadataCacheLock); 
+                
+                LWLockAcquire(MetadataCacheLock, LW_EXCLUSIVE);
```
gcaragea: Can this create deadlock situations? Upgrading locks usually is a tricky situation, can lead to a situation where:
- proc A has SHARED lock on foo and bar
- proc B has SHARED lock on foo and bar

- proc A requests EXCLUSIVE lock on foo and gets it
- proc B requests EXCLUSIVE lock on bar and gets it
- proc A requests EXCLUSIVE lock on bar and waits
- proc B requests EXCLUSIVE lock on foo and waits - deadlock

wengyanqing: Hi George,

Thanks for your comment. It only has one lock (MetadataCacheLock) and one resource (MetadataCache) which lock on, so it should not cause dead lock.

PR #148, opened 2015-12-02
HAWQ-204. Long dbname causes core dump

jiny2:
This fix cut the long string of database name. 
And the host info string is moved the end of string which maybe missed if the whole string is too long.

For example:
postgres=# \c database1234567890123456789012345678901234567890123456789012345
You are now connected to database "database1234567890123456789012345678901234567890123456789012345" as user "yjin".

postgres: port  5432, yjin database12... [local] con369 idle [local]


yaoj2:
Looks good to me.

PR #152, opened 2015-12-03
HAWQ-209. Hide detailed log of metadata cache by default.

zhangh43:
we need to hide detailed log of metadata cache by default
and show them when guc debug_print_split_alloc_result is set.


jiny2:
+1 The change looks good to me.

wangzw:
+1 The change looks good to me.

PR #157, opened 2015-12-03
HAWQ-220. Add retry to connect postmaster when handle RUAlive message

linwen:
Add retry to connect postmaster when handle RUAlive message. 
Please review.


huor:
The improvement looks good. +1

jiny2:
The code looks good.

PR #159, opened 2015-12-03
HAWQ-221. Cleanup unused files related to optimizer

entong:
There are three files originally used for running gpopt as a standalone process. These code are #ifdef out and were cleaned up on the gpdb side. I'm cleaning them up for Hawq as well since there's no plan of running gpopt as a standalone process in near future.


vraghavan78:
LGTM!

entong:
@xinzweb could you take a quick look?

PR #160, opened 2015-12-04
HAWQ-223. Improve AOGetSegFileDataLocation performane.

zhangh43:
GpPolicyFetch is called per file, which needs to be changed to be called per relation.


yaoj2:
Looks good

wengyanqing:
+1

PR #161, opened 2015-12-04
HAWQ-225: Adding aggregate with grouping set test case

karthijrk:
Adding new test case for aggregate with grouping sets. This test was failing in gpdb with seg fault but HAWQ was working fine. I am just porting this test that we added in gpdb to HAWQ.


armenatzoglou:
LGTM +1

foyzur:
LGTM

karthijrk:
This change has been merged and here is the commit id : 0ec66d57d6cb8c06037e4a3fbb4daca4b533c9df

PR #163, opened 2015-12-07
HAWQ-226. Core dump caused by data locality.

zhangh43:
total split number is counted twice when hostIDs of a file is null in allocate_random_relation() method


jianlirong:
+1, LGTM.

ztao1987:
+1

PR #164, opened 2015-12-07
HAWQ-227. Data locality downgrade by wrong insert host.

zhangh43:
when we determin the insert host of a block, we should ignore the file contains only one block. because in this case, we cannot know which is the right insert host.  when hosts with same occur times exists we cannot know which is the right insert host either.


yaoj2:
+1

jianlirong:
+1, LGTM.

PR #167, opened 2015-12-08
HAWQ-232: recovery Pass3 do not need to redo all relfile and relation dir create/drop for on flight transaction

liming01:
Because all persistent table state and relation dir and relfile on hdfs are keep in consistent status at recovery PASS2,  recovery PASS3 do not need to redo all relfile and relation dir create/drop for on flight transaction.


changleicn:
Good fix.

linwen:
+1

PR #168, opened 2015-12-08
HAWQ-224. Fix syntax error of make distprep. a missing '\' in line 11.

SuperJDC:
add ' \' to the end of line 11

### DIFF
--- a/contrib/Makefile
+++ b/contrib/Makefile
@@ -8,7 +8,7 @@ WANTED_DIRS = \
 		pgbench \
 		orafce \
 		extprotocol \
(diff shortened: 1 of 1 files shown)


radarwave:
This fix the syntax error but did not fix 'make distprep' option,  we might enable distprep option in the future. So this fix is okay.

The only issue is the commit message, we should follow the format. Please update the commit message before we merge it in. Thanks.

For example:
"HAWQ-224. Fix syntax error of make distprep."

radarwave:
Sorry, I mean the commit message title, it should include the JIRA number.

PR #170, opened 2015-12-09
HAWQ-233. Fix HAWQ initialization failed on Centos7

leoncamel:
Hi,
We have investigated this issue, and we found the root cause is that gcc-4.8 introduced new feature "-faggressive-loop-optimizations". This optimization feature is enabled by default, and it will break variable length array under some condition.

Similar issue can be found from PostgreSQL 9.2. See the following links for detail:
1. http://www.postgresql.org/message-id/14242.1365200084@sss.pgh.pa.us
2. https://gcc.gnu.org/gcc-4.8/changes.html
3. http://pkgs.fedoraproject.org/cgit/postgresql.git/commit/?h=f19&id=2a75d7af27fdf03e875f76f4c272c8435f565d10

So, we just borrow the idea to fix this issue and create a patch. Please review it.
Thanks.

### DIFF
--- a/configure
+++ b/configure
@@ -4538,6 +4538,32 @@ $as_echo "no" >&6; }
 fi
 rm -f core conftest.err conftest.$ac_objext conftest.$ac_ext
 
+  # Disable optimizations that do aggresive-loop-optimization for variable length array; needed for gcc 4.8+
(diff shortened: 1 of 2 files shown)


wangzw:
CI passed. Manually tested on centos7. +1

PR #171, opened 2015-12-09
HAWQ-236. Meta data hit ratio is -0.0 when file len is zero.

zhangh43:
hit ratio should be 0.0 not -0.0 when file len i zero


yaoj2:
Looks good

wengyanqing:
+1

PR #172, opened 2015-12-09
HAWQ-237. failed to acquire resource because of no available resource…

jiny2:
This fix is to avoid wrongly calculating request waiting time when it is put at the head of the waiting queue.


linwen:
looks good.

yaoj2:
Looks good

PR #174, opened 2015-12-09
HAWQ-191. Remove Analyzer plugin from PXF

shivzone:


pxf/build.gradle
```diff
@@ -218,7 +218,7 @@ project('pxf-service') {
 
 project('pxf-hdfs') {
     dependencies {
-        compile(project(':pxf-service')) //Yikes, HdfsAnalyzer is directly accessing the bridge
+        compile(project(':pxf-service'))
```
hornn: Can't we replace the dependency with a dependency on pxf-api (like pxf-hbase does)?

hornn:
Removed code looks good.
+0.5 (waiting for the warning message code)

pxf/build.gradle
shivzone: There are other compile time dependancies unrelated to the analyzer

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/InvalidPathResource.java
```diff
@@ -153,9 +167,39 @@ private String parseVersion(String path) {
     }
 
     /**
+     * Parses the version part from the path.
```
hornn: 1. Please update doc - it parses the part after the version part, right? /pxf/version/endpoint
2. could you please refactor parseVersion() to also use PathSegment? 
Actually, both functions can perhaps be consolidated. [e.g. parsePathSegment(path, VERSION), parsePath(path, ENDPOINT)]

shivzone: These are private methods used only by wrongPath(). Having an enum and separate function seems to be an overkill for this. will do away with helper functions and use pathSegments directly to extract the version and endpoint.

hornn: Sounds good.

pxf/build.gradle
hornn: There are 2 other classes using pxf-service. I will open a JIRA to fix it.

hornn:
+1

pxf/build.gradle
shivzone: Yes, separate jira makes more sense

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/InvalidPathResource.java
shivzone: Fixed this directly on master

PR #175, opened 2015-12-10
HAWQ-239. Reduce the call times of GpPolicy to improve performance on…

zhangh43:
… scan empty table.

in get_block_locations_and_claculte_table_size(), GpPolicy is called twice for each relation, we need to reduce it to zero

This pull request also fix some initialization of blocknum and hitratio


ztao1987:
+1

yaoj2:
+1

PR #178, opened 2015-12-10
HAWQ-241: Remove  MirroredLock related codes

liming01:
MirroredLock is no use in hawq because there is no segment mirroring, and it is not used by master mirroring which uses xlog directly.

The reason why we need to remove it is: in some scenarios the request order of this lock and CheckpointStartLock may cause deadlock.


changleicn:
nice fix.

huor:
The fix looks good. +1

PR #182, opened 2015-12-11
Move ivy.xml for qpopt inside apache-hawq for better dependency management

xinzweb:
This is part 1 of 2 step check-in to have better way managing dependencies between HAWQ and optimizer.

Originally, we have ivy.xml managed separately from apache-hawq, and caused inconsistency between HAWQ and optimizer components.

With this fix, we can change the dependency the same time we change the HAWQ code, hence reduce failures due to inconsistent dependencies.

### DIFF
--- a/src/backend/gpopt/ivy-build.xml
+++ b/src/backend/gpopt/ivy-build.xml
@@ -1,0 +1,69 @@
+<project name="cdb3" default="resolve" xmlns:ivy="antlib:org.apache.ivy.ant">
+
+  <!-- ================================================================ -->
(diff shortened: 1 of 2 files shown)


entong:
@wangzw We are bringing the ivy settings related to Orca inside apache, since the translator changes and version update always go hand in hand. By doing so we don't need to push to both repos for every refresh of Orca. Saves us tons of time.

wangzw:
looks good to me. +1

entong:
Thanks! This is a small step on the way to open source Orca.

PR #186, opened 2015-12-12
HAWQ-184. Add license header for some files and add a exclude file(po…

linwen:
Add license header for some files and add a exclude file(pom.xml) for Apache Rat.
Please review, Thanks!


changleicn:
looks good.

huor:
The fix looks good. +1

PR #188, opened 2015-12-14
HAWQ-251. Add gpfaultinjector back for hawq2.0 using.

radarwave:
We need to add gpfaultinjector back, but will require some changes to make it work for HAWQ2.0.
This ticket will only add it back, we will have another ticket to do modifications.


yaoj2:
Looks good

huor:
The fix looks good. +1

PR #191, opened 2015-12-15
HAWQ-255: change CHECKPOINT_START_LOCK to a fine grained scope so that checkpoint not blocked.

liming01:
In some scenarios the request order of this lock and CheckpointStartLock may cause deadlock, MirroredLock and LockRelationForResynchronize()  are useless in hawq now, and it may cause deadlock in AbortTransaction()/CommitTransaction()/smgrDoDeleteActions(), so we remove them.

MirroredLock was removed at previous checkin. LockRelationForResynchronize() need to be removed here.

### DIFF
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -2317,14 +2317,14 @@ CommitTransaction(void)
 	willHaveObjectsFromSmgr =
 			PersistentEndXactRec_WillHaveObjectsFromSmgr(EndXactRecKind_Commit);
 
-	if (willHaveObjectsFromSmgr)
-	{
-		/*
-		 * We need to ensure the recording of the [distributed-]commit record and the
-		 * persistent post-commit work will be done either before or after a checkpoint.
-		 */
-		CHECKPOINT_START_LOCK;
-	}
+	/* In previous version, we ensured the recording of the [distributed-]commit record and the
+	 * persistent post-commit work will be done either before or after a checkpoint.
+	 *
+	 * However the persistent table status will be synchronized with AOSeg_XXXX
+	 * table and hdfs file in PersistentRecovery_Scan() at recovery PASS2.
+	 * We don't need to worry about inconsistent states between them. So no
+	 * CHECKPOINT_START_LOCK any more.
+	 */
 
 	/* Prevent cancel/die interrupt while cleaning up */
 	HOLD_INTERRUPTS();
@@ -2426,11 +2426,6 @@ CommitTransaction(void)
 	 */
 	AtEOXact_smgr(true);
 
-	if (willHaveObjectsFromSmgr)
-	{
-		CHECKPOINT_START_UNLOCK;
-	}
-	
 	AtEOXact_MultiXact();
 
 	ResourceOwnerRelease(TopTransactionResourceOwner,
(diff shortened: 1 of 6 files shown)


src/backend/access/transam/xact.c
```diff
@@ -2317,14 +2317,14 @@ CommitTransaction(void)
...
 			PersistentEndXactRec_WillHaveObjectsFromSmgr(EndXactRecKind_Commit);
 
-	if (willHaveObjectsFromSmgr)
-	{
-		/*
-		 * We need to ensure the recording of the [distributed-]commit record and the
-		 * persistent post-commit work will be done either before or after a checkpoint.
-		 */
-		CHECKPOINT_START_LOCK;
-	}
```
wangzw: I have some concern about removing this lock. As the comment indicated, what will happen in such case?

1) flush commit record.
2) start a checkpoint.
3) persistent post-commit work (change persistent table)

src/backend/access/transam/xact.c
```diff
@@ -2317,14 +2317,14 @@ CommitTransaction(void)
...
+	/* In previous version, we ensured the recording of the [distributed-]commit record and the
+	 * persistent post-commit work will be done either before or after a checkpoint.
+	 *
+	 * However the persistent table status will be synchronized with AOSeg_XXXX
+	 * table and hdfs file in PersistentRecovery_Scan() at recovery PASS2.
+	 * We don't need to worry about inconsistent states between them. So no
+	 * CHECKPOINT_START_LOCK any more.
+	 */
```
wangzw: Not true. Consider this case.

1) flush commit record.
2) start a checkpoint.
3) checkpoint complete successfully.
4) failed to drop file and or fail to modify persistent table.

Since checkpoint was started (2) and finished (3) successfully after flush commit record (1), the commit record will be truncated during the checkpoint but failure (4) actually happened. I do not think recovery process can handle this case well.

src/backend/storage/smgr/smgr.c
```diff
@@ -1327,28 +1327,6 @@ smgrDoDeleteActions(
...
-
-	/*
-	 * The logic will eventually obtain a CheckpointStartLock in PersistentRelation_Dropped(),
-	 * but functions called from this function my obtain Exclusive locks before the
-	 * CheckpointStartLock is obtained. This could cause a potential deadlock in the future.
-	 * We need to take a CheckpointStartLock here to maintain proper lock ordering
-	 * (i.e. MirrorLock -> CheckpointStartLock ).
-	 */
-	CHECKPOINT_START_LOCK;
-
```
wangzw: Although we do not have a mirror any more, but we still have MirrorLock (should be removed). If we remove this lock, can we still keep the lock ordering and prevent from dead locking?

liming01: I have already removed MirrorLock which can interfere with CHECKPOINT_START_LOCK.

wangzw: Nice work #178

src/backend/access/transam/xact.c
liming01: Now hawq doesn't report error when (4) occurs: I set break in gdb at xact.c:2427, and run below statements, when hang at the "COMMIT", then I run "hadoop dfsadmin -safemode enter" to set hdfs to save mode, then continue in gdb, the commit successfully finished with only warning. Similar problem for ABORT of transaction which includes 'CREATE TABLE'. 

postgres=# begin transaction ISOLATION LEVEL SERIALIZABLE;BEGIN
postgres=# drop table tableinfs2;                                                                                                                             DROP TABLE
postgres=# commit;
WARNING:  could not remove relation directory 24974/16387/24975: Input/output error
CONTEXT:  Dropping file-system object -- Relation Directory: '24974/16387/24975'
COMMIT
postgres=# select * from tableinfs2;
ERROR:  relation "tableinfs2" does not exist

As for your question above, the recovery process will redo (4) again and similarly it report warning info if we failed to drop file and or fail to modify persistent table. Thanks.

src/backend/access/transam/xact.c
liming01: We will add a test with FaultInjection here to cover this scenario, I don't think it would be a problem.

src/backend/access/transam/xact.c
wangzw: If (4) failed, HAWQ should definitely NOT report error since the transaction has already committed (1).

No redo will happen during recovery process since commit log has been truncated after checkpoint.

liming01: The status of current transaction will not be truncated at (4), the truncate related code is:
CreateCheckPoint() --> CheckPointGuts() --> CheckPointMultiXact() -> TruncateMultiXact() -> SimpleLruTruncate()
In this function, cutoffPage is based on the min value of OldestMemberMXactId and OldestVisibleMXactId for all backends' proc. 

And OldestMemberMXactId and OldestVisibleMXactId for current backend's proc are reset in
AtEOXact_MultiXact(). In AbortTransaction() and CommitTransaction()， this function is called after AtEOXact_smgr(). 

So in recovery process, we can know that the current transaction has already committed, and all persistent table and hdfs files should be redo.

linwen:
looks good

PR #192, opened 2015-12-16
HAWQ-218. Core dump while selecting data from partitioned table with where condition that belongs to no partition

huor:


yaoj2:
Looks good

zhangh43:
+1

PR #193, opened 2015-12-16
HAWQ-254. Fix init standby fails if segments in 'slaves' file not incialized

radarwave:
Sanity test passed.


liming01:
+1,  good fix.

yaoj2:
Looks good

PR #195, opened 2015-12-17
HAWQ-218. Core dump while selecting data from partitioned table with where condition that belongs to no partition

huor:


ztao1987:
+1

wangzw:
looks good. +1

PR #197, opened 2015-12-18
HAWQ-259. YARN mode resource manager gets YARN cluster report once be…

jiny2:
The GRM cluster report is triggered only when there is at least one segment registered through FTS heart-beat.

To handle the case, no segment is started in YARN mode, guc hawq_rm_nocluster_timeout is introduced, if when resource manager has waited for the possible segment registeration more than hawq_rm_nocluster_timeout seconds, the resource requests are rejected due to no enough available segment.


linwen:
looks good to me

huor:
The fix looks good. +1

PR #198, opened 2015-12-18
HAWQ-260. Resource manager wrongly calculates number of vseg

jiny2:
This is a bug.

### DIFF
--- a/src/backend/resourcemanager/resqueuemanager.c
+++ b/src/backend/resourcemanager/resqueuemanager.c
@@ -3384,9 +3384,8 @@ int computeQueryQuota(ConnectionTrack conn, char *errorbuf, int errorbufsize)
 		 */
 		if ( conn->MinSegCountFixed != conn->MaxSegCountFixed )
 		{
-			vseglimit = conn->VSegLimit ?
-						conn->VSegLimit :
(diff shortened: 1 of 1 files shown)


huor:
The fix looks good. +1

src/backend/resourcemanager/resqueuemanager.c
```diff
@@ -3384,9 +3384,8 @@ int computeQueryQuota(ConnectionTrack conn, char *errorbuf, int errorbufsize)
 		 */
 		if ( conn->MinSegCountFixed != conn->MaxSegCountFixed )
 		{
-			vseglimit = conn->VSegLimit ?
-						conn->VSegLimit :
-						conn->VSegLimitPerSeg * PRESPOOL->AvailNodeCount;
+			vseglimit = conn->VSegLimitPerSeg * PRESPOOL->AvailNodeCount;
```
huor: I would be clearer using below:
vseglimit = min(conn->VsegLimit, (conn->VSegLimitPerSeg) * (PRESPOOL->AvailNodeCount));

jiny2: I'd like to keep original code.

yaoj2:
+1

PR #203, opened 2015-12-20
HAWQ-266. Check if krb_server_keyfile exist in hawq-site.xml

radarwave:


yaoj2:
+1

liming01:
+1

tools/bin/hawq_ctl
```diff
@@ -111,10 +111,13 @@ class HawqInit:
             self.standby_host_name = ''
 
         if 'enable_secure_filesystem' in self.hawq_dict:
-            self.enable_secure_filesystem=self.hawq_dict['enable_secure_filesystem']
-            self.krb_server_keyfile =self.hawq_dict['krb_server_keyfile']
+            self.enable_secure_filesystem = self.hawq_dict['enable_secure_filesystem']
```
ljainpivotalio: **enable_secure_filesystem** could be set to **OFF** in hawq-site.xml. Current logic is assuming that security is on as long as enable_secure_filesystem parameter is present.

radarwave: Current logic is if 'enable_secure_filesystem' present in hawq-site.xml, then get it's value, the value depend on what user set it. If it does not present, we assume it's off.

ljainpivotalio: Thanks for the clarification.

LGTM

PR #204, opened 2015-12-21
HAWQ-265. Change metadata share memory flush strategy to prevent out of share memory problem when create too many hdfs_file.

stanlyxiang:
please review. thanks.


zhangh43:
+1

PR #206, opened 2015-12-21
Hawq 257

hornn:
When parsing PXF URI, change the port to pxf_service_port only if pxf_isilon is ON.
The parsed port is always used (both Isilon and regular cases), which makes the code simpler.
In case of HA, the port is not specified in the URI, so we use pxf_service_port instead.

### DIFF
--- a/src/backend/access/external/test/pxfuriparser_test.c
+++ b/src/backend/access/external/test/pxfuriparser_test.c
@@ -44,7 +44,8 @@ test__parseGPHDUri__ValidURI(void **state)
 
 	assert_string_equal(parsed->protocol, "pxf");
 	assert_string_equal(parsed->host, "1.2.3.4");
-	assert_string_not_equal(parsed->port, "5678"); /* it should be pxf_service_port */
+	assert_string_equal(parsed->port, "5678");
+	assert_true(parsed->ha_nodes == NULL);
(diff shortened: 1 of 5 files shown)


src/backend/access/external/test/pxfuriparser_test.c
```diff
@@ -76,6 +77,87 @@ test__parseGPHDUri__ValidURI(void **state)
...
+	assert_string_equal(parsed->protocol, "pxf");
+	assert_string_equal(parsed->host, "servername");
+	assert_int_equal(atoi(parsed->port), pxf_service_port); /* it should be pxf_service_port */
+	assert_true(parsed->ha_nodes == NULL);
+	assert_string_equal(parsed->data, "some/path/and/table.tbl");
+
+	freeGPHDUri(parsed);
+
+	/* set pxf_isilon back to false */
+	pxf_isilon = false;
```
shivzone: is this required ?

shivzone:
+!

src/backend/access/external/test/pxfuriparser_test.c
hornn: It's a system variable, so if I change it in the beginning of the test (line 141) I should clean after me.
It can potentially affect other tests that assume it has specific value, although that's not very likely.

PR #207, opened 2015-12-22
HAWQ-269. Change net_disk_ratio from int to double type.

zhangh43:
net_disk_ratio is the punishment factor of the size of remote read compared with local read.
we need to set it to be double type.

### DIFF
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -720,7 +720,7 @@ int			optimizer_log_failure;
 double		optimizer_cost_threshold;
 double		optimizer_nestloop_factor;
 double locality_upper_bound;
-int net_disk_ratio;
+double net_disk_ratio;
 bool		optimizer_cte_inlining;
 int		optimizer_cte_inlining_bound;
 double 	optimizer_damping_factor_filter;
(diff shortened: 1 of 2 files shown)


yaoj2:
+1

wengyanqing:
looks good

src/backend/utils/misc/guc.c
```diff
@@ -720,7 +720,7 @@ int			optimizer_log_failure;
 double		optimizer_cost_threshold;
 double		optimizer_nestloop_factor;
 double locality_upper_bound;
-int net_disk_ratio;
+double net_disk_ratio;
```
huor: The indentation needs be followed here

src/backend/utils/misc/guc.c
```diff
@@ -6857,6 +6846,15 @@ static struct config_real ConfigureNamesReal[] =
 		&locality_upper_bound,
 		1.2, 1.0, 2.0, NULL, NULL
 	},
+	{
+		{"net_disk_ratio", PGC_USERSET, DEVELOPER_OPTIONS,
+			gettext_noop("set the scan volumn ratio of disk against network."),
```
huor: "Sets the scan ..."

huor:

src/backend/utils/misc/guc.c
zhangh43: I'll change it, Thanks

PR #209, opened 2015-12-22
HAWQ-254. Don't register segment if master is running in utility mode

linwen:
Please review.


huor:
The fix looks good. +1

jiny2:
+1 The code change is ok to me

PR #210, opened 2015-12-22
HAWQ-268. Fix HAWQ activate standby fails

radarwave:
FTS tests passed.


huor:
The fix looks good. +1

yaoj2:
The fix looks good.

PR #211, opened 2015-12-23
HAWQ-273. Need re-fetch snapshot at planning stage

liming01:
Because data locality and hdfs metadata cache will fetch pg_aoseg meta data when planning, we need to re-fetch the latest snapshot.

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -1645,7 +1645,7 @@ exec_simple_query(const char *query_string, const char *seqServerHost, int seqSe
 		querytree_list = pg_analyze_and_rewrite(parsetree, query_string,
 												NULL, 0);
 
-		plantree_list = pg_plan_queries(querytree_list, NULL, false, QRL_ONCE);
+		plantree_list = pg_plan_queries(querytree_list, NULL, true, QRL_ONCE);
 
 		/* Done with the snapshot used for parsing/planning */
 		ActiveSnapshot = NULL;
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -850,14 +850,6 @@ pg_plan_query(Query *querytree, ParamListInfo boundParams, QueryResourceLife res
 /*
  * Generate plans for a list of already-rewritten queries.
  *
- * If needSnapshot is TRUE, we haven't yet set a snapshot for the current
- * query.  A snapshot must be set before invoking the planner, since it
- * might try to evaluate user-defined functions.  But we must not set a
- * snapshot if the list contains only utility statements, because some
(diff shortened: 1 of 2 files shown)


src/backend/tcop/postgres.c
```diff
@@ -1645,7 +1645,7 @@ exec_simple_query(const char *query_string, const char *seqServerHost, int seqSe
 		querytree_list = pg_analyze_and_rewrite(parsetree, query_string,
 												NULL, 0);
 
-		plantree_list = pg_plan_queries(querytree_list, NULL, false, QRL_ONCE);
+		plantree_list = pg_plan_queries(querytree_list, NULL, true, QRL_ONCE);
```
wangzw: It's not right
```
A snapshot must be set before invoking the planner, since it
 * might try to evaluate user-defined functions.  But we must not set a
 * snapshot if the list contains only utility statements, because some
 * utility statements depend on not having frozen the snapshot yet.
```

Snapshot for plannable statement was set before, see line 1631.

wangzw:
For PBE query, snapshot is set int PortalRunUtility(), so other changes also cause issues.

wangzw:
```
Because data locality and hdfs metadata cache will fetch pg_aoseg meta data when planning, we need to re-fetch the latest snapshot.
```

That's right, but all plannable statements have already got the snapshot. Did we missing something?

wangzw:
It is the issue of order of locking and getting snapshot. After unlocking, snapshot become outdated and should renew.

src/backend/tcop/postgres.c
liming01: The comment is confusing, it just describes the details about implementation of this function, instead of describing the API parameter constraint. The logic of skipping snapshot for the utility statements is inside pg_plan_queries(). So it is no problem.

liming01:
In fact when executing one normal statement(SELECT/INSERT/UPDATE/DELETE), the snapshot will be fetched several times, at previous version, we fetched one snapshot before transform query, and re-fetched one at the beginning of query execution. 

Your last comments above is exactly the root cause of this defect. ~_~

changleicn:
looks there are still other places that call pg_plan_queries, why don't we re-fetch the snapshot?

liming01:
Yes, only 2 functions:
- _SPI_prepare_plan():  you can find that before calling pg_plan_queries(), it already fetch a new snapshot. And between them, it just some simple code, which is impossible to lock. So we don't need to re-fetch one.  
- exec_parse_message(): From the comment in the code, it just called for syntax check.
                 /**
		 * Even though we do not store the plan, we want to perform planning
		 * to know if SQL syntax is supported.
		 */

changleicn:
cool. good fix.

PR #213, opened 2015-12-24
HAWQ-278. Add min_cost for each query.

zhangh43:
When we run 50 concurrancy insert into values(3) on 16 hosts.
They will run on only one host, not balanced.
This is caused by we pass cost zero to RM for "insert into values" like query.
We need to assign a min_cost for each query. 
One block size(128M) is by default


jiny2:

ztao1987:
+1 looks good

PR #214, opened 2015-12-24
HAWQ-268. Start standby node before master node  while doing hawq init standby

radarwave:
If hawq master start before standby, in some cases standby will do recovery, this will cause master can't connect to standby to do "Synchronized" check. So we will start standby first and sleep 5 seconds, then start master. This should avoid most of the "Synchronized" failues.

Error message is:
[' Not Synchronized | Connection error | 2015-12-23 20:52:20-08 | Master mirroring connect failed for entry db. Master unable to connect to test5 with options host=test5 port=20000 : FATAL:  the database system is in recovery mode \n',

### DIFF
--- a/tools/bin/lib/hawqinit.sh
+++ b/tools/bin/lib/hawqinit.sh
@@ -352,16 +352,7 @@ standby_init() {
     else
         LOG_MSG "[INFO]:-HAWQ master stopped" verbose
     fi
-    
-    ${SSH} -o 'StrictHostKeyChecking no' ${hawqUser}@${master_host_name} \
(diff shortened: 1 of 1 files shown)


liming01:
Good fix.  
BTW, if we can auto detect it in recovery mode, then wait until it finished, instead of hard code to sleep 5 minutes, it would be much better in case of some scenario that the recovery may take more than 5 seconds to finish.

radarwave:
Good suggestion, to accomplish this, we need to do  more investigation to standby recovery process, so we can know how long it will require for different cases and what's the way to check if standby is in recovery mode. Will have another Jira to trace this. Thanks.

PR #217, opened 2015-12-25
HAWQ-281. User can alter any resource queue no matter whether its busy

jiny2:
This provides practicality of manipulating resource queues in real environment. User can alter a resource queue on demand any time without waiting for releasing all corresponding queues' workload.

hawq_rm_force_alterqueue_cancel_queued_request guc variable is introduced, default value is 'true' that means alter resource queue statement can force resource manager to cancel all resource requests that can not be adjusted to fit new resource queue definitions. If it is set 'false', the alter resource queue statement will be cancelled to avoid canceling any queued query resource requests in resource manager.


huor:
This feature looks good. +1

PR #219, opened 2015-12-25
HAWQ-283. Remove the content column from pg_aoseg_xxxxx

liming01:
In HAWQ 2.0, the content column inside the table pg_aoseg.pg_aoseg_xxxxx is meaningless now. We need to remove it.


jianlirong:
+1, LGTM.

huor:
The fix looks good. +1

PR #221, opened 2015-12-28
HAWQ-285. not freed connection track instance

jiny2:
This is a bug, in case altering queue statement is canceled, temporary created connection track instance is not freed in rebuildResourceQueueTrackDynamicStatusInShadow()


yaoj2:
Looks good

radarwave:
Looks good. +1

PR #222, opened 2015-12-28
HAWQ-286. Queued resource requests are not adjusted along with the ch…

jiny2:
With this fix, queued requests are adjusted when cluster size changed.


huor:
The fix looks good. +1

linwen:
Looks good.

PR #223, opened 2015-12-29
HAWQ-219. Rebuild gp_persistent_relation_node when rebuilding persistent tables

liming01:
Fixed bug:  Built-in function gp_persistent_build_all doesn't reshape gp_persistent_relation_node table. No need to add info for heap tables.


jianlirong:
+1, LGTM.

huor:
The fix looks good. +1

PR #224, opened 2015-12-29
HAWQ-287. Control whether to pass data locality prefer list to resour…

zhangh43:
…ce manager.
to achieve workload balance we need to control whether to pass data locality prefer list to resource manager.


jianlirong:
+1, LGTM.

ztao1987:
+1

PR #226, opened 2015-12-29
HAWQ-288. Support hawqfaultinjector as an internal utility for fault injection purpose - client side

huor:
@radarwave , @linwen , please review.


radarwave:
LGTM.
It's great to have a new hawqarray, we can enable more and more functions in hawqarray in the future..

linwen:
looks good.

PR #227, opened 2015-12-29
HAWQ-291. Make resource queue default 256mb as virtual segment resour…

jiny2:
increase default virtual segment resource quota from 128MB to 256MB.


yaoj2:
Looks good

PR #228, opened 2015-12-30
HAWQ-292. Remove debugging information in base.py due to introduction of hawqfaultinjector

huor:


jiny2:
+1 The change is ok to me.

PR #229, opened 2015-12-30
HAWQ-293. Enable writable external table to support on number syntax.

zhangh43:
writable external table is used for export data in hawq to others.
In hawq1.x default is on all, which is deprecated in hawq2.x
we support this feature by using on number instread of on all.

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -370,7 +370,7 @@ static void update_ri_trigger_args(Oid relid,
 					   bool fk_scan,
 					   bool update_relname);
 static Datum transformLocationUris(List *locs, List* fmtopts, bool isweb, bool iswritable);
-static Datum transformExecOnClause(List	*on_clause, int *preferred_segment_num);
+static Datum transformExecOnClause(List	*on_clause, int *preferred_segment_num, bool iswritable);
 static char transformFormatType(char *formatname);
 static Datum transformFormatOpts(char formattype, List *formatOpts, int numcols, bool iswritable);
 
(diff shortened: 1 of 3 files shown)


yaoj2:
+1

jianlirong:
+1, LGTM.

src/backend/commands/tablecmds.c
```diff
@@ -17666,6 +17666,12 @@ static Datum transformExecOnClause(List	*on_clause, int *preferred_segment_num)
 		}
 		else if (strcmp(defel->defname, "master") == 0)
 		{
+			if(iswritable){
+			  ereport(ERROR,
+	 	      (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+			    errmsg("the ON master syntax for writable external tables is deprecated"),
```
hornn: would it be possible to add a hint saying what is the allowed option (ON number)?

zhangh43: Only ON number is allowed, e.g. ON 3 will start 3 virtual segments to run the query.

Here's an example to create a writable external table in Hawq 2.0:
postgres=# create table a (i int);
CREATE TABLE
postgres=# insert into a values(1);
INSERT 0 1
postgres=# insert into a values(2);
INSERT 0 1
postgres=# insert into a values(10);
INSERT 0 1
postgres=# CREATE WRITABLE EXTERNAL WEB TABLE externala (output text)         EXECUTE 'cat > /tmp/externala' on 3 FORMAT 'TEXT'   DISTRIBUTED RANDOMLY;
CREATE EXTERNAL TABLE
postgres=# insert into externala select * from a;
INSERT 0 3

You can check the result file in /tmp/externala:
1
2
10

hornn: Thank you for the clarification. I meant an errhint() in the ereport message.
Something like
   ereport(ERROR,
	 	      (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
			    errmsg("the ON master syntax for writable external tables is deprecated"),
                            errhint("Only ON number is allowed, where number is the number of segments"),
		      errOmitLocation(true)));

PR #231, opened 2015-12-30
HAWQ-295. New metadata flush strategy remove 1 entry every time flush due to flush condition wrong.

stanlyxiang:
please reviews, thanks.

### DIFF
--- a/src/backend/cdb/cdbmetadatacache_process.c
+++ b/src/backend/cdb/cdbmetadatacache_process.c
@@ -443,7 +443,7 @@ ProcessMetadataCacheCheck()
             DestroyHdfsFileInfo(file_info);
             total_remove_files++;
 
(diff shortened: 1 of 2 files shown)


src/backend/cdb/cdbmetadatacache_process.c
```diff
@@ -443,7 +443,7 @@ ProcessMetadataCacheCheck()
             DestroyHdfsFileInfo(file_info);
             total_remove_files++;
 
-            double cache_entry_ratio = hash_get_num_entries(MetadataCache) / metadata_cache_max_hdfs_file_num;
+            double cache_entry_ratio = (hash_get_num_entries(MetadataCache) * 1.0) / metadata_cache_max_hdfs_file_num;
```
huor: It would be better to use:
double cache_entry_ratio = ((double) hash_get_num_entries(MetadataCache)) / metadata_cache_max_hdfs_file_num;

huor:
The fix looks good. +1

PR #233, opened 2015-12-30
HAWQ-294. Rename prefix of logs in pg_log from gpdb-xxx to hawq-xxx

huor:


linwen:
Looks good.

radarwave:
Looks good to me.

PR #234, opened 2015-12-31
HAWQ-297. Double check hdfs and split length equals to metadata logic…

zhangh43:
… length.
for each table, we store file logic length in pg_aoseg metadata table.
when scanning tables, we obtain blocks information from HDFS(or cache)
we need to ensure the file length in HDFS equals to the logic length of file in pg_aoseg.
Again, we also need to ensure after we assign each blocks to virtual segments, the total length of splits equals to the logic length of file in pg_aoseg.


yaoj2:
+1

wengyanqing:
Looks good.

PR #235, opened 2015-12-31
HAWQ-70. Refetch file's all block locations when partly hit metadata …

wengyanqing:
re-fetch file's all block locations, because hdfs will give incorrect result when fetch partly


zhangh43:
+1 LGTM

yaoj2:
+1

PR #236, opened 2015-12-31
Hawq 165

hornn:
Most of the changes are indentations and simple replacement of "private Log" by "private static final LOG".
There are also a few javadoc changes, where I spotted missing documentation.

Notable changes: 
2 unittests where the Log output was tested. The mocking of the Log element had to be changed after it became static final. The interesting classes are all in the second commit.
* RecordKeyAdapterTest - mock was changed to use         Whitebox.setInternalState(RecordkeyAdapter.class, LOG);
* ProfilesConf - the Log variable was not changed to static because this is a singleton element, so all variables will be created once anyway. The problem with changing it to static was that Whitebox.setInternalState fails because the instance is only created on first access (lazy initialization), so at the time of the setup of the test it cannot be set, so I left it as only private.

### DIFF
--- a/contrib/hawq-hadoop/hawq-mapreduce-tool/.gitignore
+++ b/contrib/hawq-hadoop/hawq-mapreduce-tool/.gitignore
@@ -1,2 +1,3 @@
 target/
 test-data/*/output
+lib/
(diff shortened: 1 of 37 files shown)


shivzone:
+1 Really good job on this

PR #237, opened 2015-12-31
HAWQ-298. Double check hdfs length longer than metadata logic length.

zhangh43:
when rollback happens, hdfs block length may increase and longer than metadata logic length.


ztao1987:
+1

wengyanqing:
LGTM

PR #241, opened 2016-01-05
HAWQ-314. Fix AO Read when a segment file larger than 4TB

realdawn:
"real_fileLen" should be int64 as "fileLen" is a int64 number and "bufferedRead->largeReadPosition" is also a int64 number.
According to the code, a int64 number is changed to int32 which make the number wrong.
See more in https://issues.apache.org/jira/browse/HAWQ-314


yaoj2:
+1

jianlirong:
Excellent job!!
+1, LGTM.

PR #244, opened 2016-01-05
HAWQ-318. Infinite loop when adjusting resource request in one resour…

jiny2:


huor:
The fix looks good. +1

yaoj2:
LGTM +1

PR #245, opened 2016-01-05
HAWQ-103. Build PXF RPMs with license and vendor parameters

hornn:
By default the license is "ASL 2.0" and the vendor is left empty.


shivzone:
+1

GodenYao:
checked with @rvs , we should put "Apache HAWQ incubating" in the vendor column for now.
If we graduate to become a top project of Apache, we may use "ASF" in the column.

PR #247, opened 2016-01-06
HAWQ-320. Can't get current user name if \$USER is not set

radarwave:
Tested that if $USER is wrong, we still can get the correct user name.


yaoj2:
+1

huor:
The fix looks good. +1

cwelton:
All checks passed, two +1s, what's preventing this from being merged?

radarwave:
This already been merged, How ever not sure why the pull request not been closed automatically.

PR #248, opened 2016-01-06
HAWQ-300. Fix out of memory during loading large volume data on large cluster with YARN mode

huor:
The fix tries to assign query memory quota to memory intensive operators in a conservative manner.
At present, we give GUC hawq_re_memory_quota_allocation_ratio a default value 0.5. This needs to be tuned later with more tests.


jiny2:

wangzw:
+1

PR #249, opened 2016-01-06
HAWQ-322. Fixed DROP TABLESPACE doesnot check sub-object(database/relation) in some cases

liming01:
We need to check catalog (pg_database/pg_class) instead of related persistent table, because the persistent table doesn't hold lock until transaction end, so concurrency problem will occurs in previous implementation. Here we add code to check catalog.


jianlirong:
+1, LGTM.

linwen:
Looks good.

PR #250, opened 2016-01-06
HAWQ-307. Ubuntu Support

cbaenziger:
This provides code to build HAWQ to pass up to source code unit-tests passing. Further, explanation added on [HAWQ-307](https://issues.apache.org/jira/browse/HAWQ-307).


ivannovick:
nice @cbaenziger!  does it fully build on Ubunutu?  Were you able to run any tests with the system started?

cbaenziger:
Hi @ivannovick it seems to fully build and pass unit-tests. I was looking for some ideas on packaging to add in RPM and DEB build support; then I can try some tests on a [Chef-BACH](https://github.com/bloomberg/chef-bach) cluster.

It's probably best to point to my Chef process for testing. (It ensured CentOS still built and Ubuntu was building.) I ran the following for:
* libhdfs3: https://github.com/cbaenziger/incubator-hawq/blob/test_kitchen/chef/hawq_build/recipes/libhdfs3.rb 
* libyarn: https://github.com/cbaenziger/incubator-hawq/blob/test_kitchen/chef/hawq_build/recipes/libyarn.rb
* HAWQ: https://github.com/cbaenziger/incubator-hawq/blob/test_kitchen/chef/hawq_build/recipes/hawq.rb

I see libyarn's tests were commented out for some reason. I'll add the failure for discussion on the HAWQ-307 JIRA; rebuilding now.

ivannovick:
@changleicn can you give some pointers on what basic set of tests can be run to verify the build?

changleicn:
Please see here for the tests used for validation: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=61320026

ivannovick:
@schubert @d can you help to review this pull request for Ubuntu support please?  I think you have the right experience to give feedback.

depends/libyarn/CMake/Platform.cmake
```diff
@@ -7,23 +7,19 @@ ELSE(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 ENDIF(CMAKE_SYSTEM_NAME STREQUAL "Linux")
 
 IF(CMAKE_COMPILER_IS_GNUCXX)
-    EXECUTE_PROCESS(COMMAND ${CMAKE_CXX_COMPILER} --version  OUTPUT_VARIABLE COMPILER_OUTPUT)
+    EXECUTE_PROCESS(COMMAND ${CMAKE_CXX_COMPILER} -dumpversion OUTPUT_VARIABLE GCC_COMPILER_VERSION)
```
craig-chasseur: Easier to just use the automatic CMake variable `CMAKE_CXX_COMPILER_VERSION` rather than deducing it manually here.

craig-chasseur:
Mostly LGTM, but I had a small suggestion about simplifying GCC version detection.

cwelton:
What is the status of this, is it ready for merge?

src/tools/fsync/test_fsync.c
```diff
@@ -69,7 +69,7 @@ main(int argc, char *argv[])
 	for (i = 0; i < XLOG_SEG_SIZE; i++)
 		full_buf[i] = 'a';
 
-	if ((tmpfile = open(filename, O_RDWR | O_CREAT, S_IRUSR | S_IWUSR, 0)) == -1)
```
ivannovick: What is the purpose of this change?  Are we sure its platform agnostic?

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -2145,7 +2145,7 @@ int findFile(const char *filename)
 int createFile(const char *filename)
 {
 	int fd = 0;
-	fd = open(filename, O_RDWR|O_CREAT);
+	fd = open(filename, O_RDWR|O_CREAT, S_IRUSR | S_IWUSR);
```
ivannovick: What is the purpose of this change? Are we sure its platform agnostic?

ivannovick:
@cbaenziger what is the status of this PR?  Is install check good testing all GREEN on this code?  If yes is it ready for merging, the PR is getting old.

src/backend/resourcemanager/communication/rmcomm_QD2RM.c
cwelton: The S_IRUSR and S_IWUSR flags indicate that the file should be created with the u+rw permission bits set, presumably default permissions were different between Ubuntu and CentOS and switching to the three parameter open() call is more explicit in the behavior.

ivannovick: ok which file is it?  Is it safe to assume read and write permissions is what we want?

src/tools/fsync/test_fsync.c
cwelton: Not sure what the file ,0 would have been for, but the new code is replacement with the code in backend//storage/file/fd.c

cbaenziger: Further, I tested using Chef for CentOS and Ubuntu; but verifying man pages I see the following for [Mac OS X](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man2/open.2.html) (which does not offer a free OSS test bed that I know of) versus [Ubuntu](http://manpages.ubuntu.com/manpages/lucid/man2/open.2.html) which looks to be from upstream Linux.

cbaenziger:
I have tested unit tests only still. I am this week hoping to ramp up on HDFS testing and YARN integration testing to see what Ubuntu-isms are still outstanding there.

I have now added the requested CMake change and rebased. (This is much smaller without the `build_instructions.md` change! Now posted on the [HAWQ-307 JIRA](https://issues.apache.org/jira/browse/HAWQ-307).) Interestingly the test code using `open`(2) failed on both CentOS and Ubuntu for me, for what it is worth.

cwelton:
Hmm interesting that you would be seeing failures on CentOS that we were not seeing.  What version of CentOS were you running?

The CMake changes look much cleaner now.

From what I can tell all of these are good changes.  I'm not sure that it is sufficient to get things good on Ubuntu, but I see no reason to hold off on the proposed patch.

+1

xunzhang:
Building `libyarn` can also not pass with high version gcc which supports C++11, plz check out this [pull request](https://github.com/apache/incubator-hawq/pull/534) for fix. I think it relates to HAWQ-307 building under Ubuntu.

tools/bin/gppylib/commands/unix.py
```diff
@@ -925,7 +925,7 @@ def getDescendentProcesses(pid):
     children = list()
     grandchildren = list()
 
-    for p in psi.process.ProcessTable().values():
+    for p in psutil.process.ProcessTable().values():
```
xunzhang: @cbaenziger This usage here is wrong since `psutil` module has no attribute `process`. Why should you use `psutil` in stead of `psi`?

xunzhang:
@changleicn , this pull request seems good except the above python problem, I have tested this patch under Ubuntu12.04/Mac OSX together with the upstream. We could merge this into master after he fixed that problem.

changleicn:
nice

cbaenziger:
@xunzhang Thanks for catching my `psutil` over sight. Is there a particular unit or functional test I could run to further verify this?

xunzhang:
@cbaenziger Since the test environment is complicated to set up and the test time is long, we use pulse for internal CI(Jenkins CI with github plugin triggered by each commit is only used for simple building test). But we only built centos/redhat linux distribution cluster for build and feature/perf test.

We can merge your pull request immediately(after your rebase) and if the pulse CI do not pass, I will send further pull request and AT you to discuss.

Advice: since hawq is a apache project, you must follow the apache standard to contribute. Specifically, please rebase your commits into one or two commits with obvious meaning. Each commit info must starts with "HAWQ-307. ".

src/test/regress/checkinc.py
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env python
 
 import sys, os, re, subprocess
+try:
+    import sysconfig
+    multiarch_triplet = sysconfig.get_config_var('MULTIARCH')
+except ImportError:
+    multiarch_triplet = sysconfig.get_config_var('MULTIARCH')
```
xunzhang: the code here is wrong... You can define `multiarch_triplet=''` here or try import `distutils.sysconfig` inside and use `distutils.sysconfig.get_config_var('MULTIARCH')` instead. Since HAWQ is near new release, we will fix it for you and remain your name. @cbaenziger

xunzhang:
@cbaenziger , we have already merged into your ubuntu patch into master(https://github.com/apache/incubator-hawq/commit/2b418bf3fdd482a9f66a34f1c6e823de745c7645), please close this pull request and related JIRA Issue. Thanks again for contributing to apache/HAWQ. :beer:

cbaenziger:
Thanks for [merging](https://github.com/apache/incubator-hawq/commit/2b418bf3fdd482a9f66a34f1c6e823de745c7645)!

tools/bin/gppylib/commands/unix.py
cbaenziger: Just to close the loop on this. The psi module does not ship for Ubuntu any more (at least out of the box with `pip` or via standard apt repos) from what I can tell with Ubuntu 14.04; it has not been touched since 2010 either. The psutil module is largely API compatible and maintained: shipping for both CentOS and Ubuntu via pip or packages.

For reference, it looks like GPDB has decided to do the same for these  now too: https://github.com/greenplum-db/gpdb/pull/486

Further, I did not find a test suite, so I tried to migrate this as much as possible; looking at the changes the Greenplum community did, clearly I missed a lot of necessary changes here.

PR #252, opened 2016-01-08
Fix 'hawq init' issue after python is removed

xinzweb:
Remove ext/python/bin from PATH to avoid using the Python version from HAWQ.
Remove the library path lib/python/pygresql from PYTHONPATH.

Otherwise 'hawq init' might not find proper Python libraries.

### DIFF
--- a/src/backend/catalog/Makefile
+++ b/src/backend/catalog/Makefile
@@ -51,6 +51,7 @@ JETPACK_PREFIX = $(JETPACK_SCHEMA).$(JETPACK_OBJECT_PREFIX)
 JETPACK_PREFIX_AUX = $(JETPACK_SCHEMA).__$(JETPACK_OBJECT_PREFIX)
 WORKFILE_PATHNAME = $$libdir\/gp_workfile_mgr
 SESSION_STATE_PATHNAME = $$libdir\/gp_session_state
+OPTUTILS_PATHNAME = $$libdir\/gpoptutils
 
 gp_toolkit.sql: gp_toolkit.sql.in
 	sed \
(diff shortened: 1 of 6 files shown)


vraghavan78:
+1

hornn:
Will this fix work also for rehl? I thought it was needed in order to make sure we run with the right version of python.

xinzweb:
Good question, and I am sorry I don't have resource to test that scenario. I rely on the build automation on all different flavor of OSs to pass.

All the fix is to 'remove' the lingering python references in the Makefiles, so that if we build on OSX, we can have proper reference of the python pre-installed on the box, other than the one comes with the HAWQ.

Again, the build process should now able to install all the python dependencies before the build. Actually, those should be factored in as part of 'configure' on the 'apache-hawq' part.

hornn:
In that case, should we also remove the reference to python in the DYLD_LIBRARY_PATH?

xinzweb:
Yes good catch. I removed, and test it works. Thanks.

hornn:
I have just found out that pygresql (and unittest2) modules are not in the python path.
It is easily reproduced when running the new minirepo command.
So it seems that in addition to installing these components we also need to add them to the path used by python.
```
minirepro 
Traceback (most recent call last):
  File "/Users/nhorn/dev/hawq/hawq-db-devel/bin/minirepro", line 24, in <module>
    from pygresql import pgdb
ImportError: cannot import name pgdb
```

xinzweb:
Let me follow up. Sorry, I don't know what caused 'close' of this PR, maybe due to the upgrade on my fork.

xinzweb:
pygresql should be located under hawq-db-devel/lib/python/pygresql. Let me check the env.

PR #255, opened 2016-01-11
HAWQ-169. Dump filespace syntax changed.

liming01:
Now we don't support specifying different URL for different segments, so CREATE FILESPACE syntax changed, we don't need to print 'dbid:' any more.


radarwave:
Looks good to me. +1

jianlirong:
+1,

LGTM.

PR #256, opened 2016-01-11
HAWQ-325. In 3-node cluster HAWQ RM does not allocate resource to que…

jiny2:
…ries if two nodes have no resource allocated from YARN

This fix is to avoid waiting for pending YARN containers when allocating resource to query resource requests.


yaoj2:
+1

PR #257, opened 2016-01-11
HAWQ-330. Fail to recognize value of activestats in catalog table pg_…

jiny2:
Fail to recognize value of activestats in catalog table pg_resqueue when the value is large


yaoj2:
+1

linwen:
+1

PR #261, opened 2016-01-12
HAWQ-333. Remove MD Versioning code

gcaragea:
We will implement a simpler invalidation-based solution instead of versioning. This will be used by the Query Optimizer to manage Metadata Cache.


gcaragea:
@armenatzoglou @foyzur please take a look.

src/backend/gpopt/translate/CTranslatorUtils.cpp
```diff
@@ -354,21 +353,16 @@ CTranslatorUtils::PmdidWithVersion
 		ullDDLv = 0;
```
foyzur: It's already initialized to 0. Do you need this code?

foyzur:
LGTM. Any reason why we cannot revert the previous commits and let Git handle the cleanup job?

armenatzoglou:
LGTM +1

gcaragea:
I removed the extra GUCs and logging / instrumentation code. Please take a look.

gcaragea:
@foyzur The code is well contained, and this is easier than reverting a few tens of commits (in my opinion). Do you have any specific concern?

foyzur:
If it's 10s of commits, then it makes sense.

PR #262, opened 2016-01-13
HAWQ-336. Set default net_disk_ratio to 1.0.

zhangh43:
After performance test, we decide to change default net_disk_ratio from 3.0 to 1.0. It'll reach better performance on large dataset or complex query.


wengyanqing:
+1

ztao1987:
+1

huor:
The fix looks good. +1

PR #264, opened 2016-01-13
HAWQ-272. Quit resource manager process on segment if postmaster is n…

linwen:
This commit is to fix the bug, segment's postmaster process is killed, but segment's resource manager still alive and send IMAlive message to master resource manager. But actually, this segment can't work.
Please review. Thanks!


jiny2:
+1 Looks good to me.

liming01:
Looks fine. +1

PR #265, opened 2016-01-13
Bump GPOPT version to 1.617

xinzweb:
Handle NULL pointer with partition elimination over full outer join queries.


vraghavan78:
LGTM

PR #267, opened 2016-01-14
HAWQ-341. PL functions core dump due to query resource is not properly allocated for function/table function/sub-query

huor:


src/backend/optimizer/plan/planner.c
```diff
@@ -443,7 +443,7 @@ planner(Query *parse, int cursorOptions,
 	}
 	PG_END_TRY();
 
-	if (plannerLevel == 1)
+	if (plannerLevel >= 1)
 	{
 		resourceNegotiateDone = false;
```
zhangh43: line448: resourceNegotiateDone = false;
is useless and should be deleted

zhangh43:
LGTM +1

linwen:
+1

PR #269, opened 2016-01-14
Fix 'hawq' issue for OSX after Python is removed.

xinzweb:
Fix following environment only for OSX:
Remove ext/python/bin from PATH to avoid using the Python version from HAWQ.
Remove the library path lib/python/pygresql from PYTHONPATH.
Remove ext/python/lib from DYLD_LIBRARY_PATH.

@hornn @radarwave

### DIFF
--- a/tools/bin/generate-greenplum-path.sh
+++ b/tools/bin/generate-greenplum-path.sh
@@ -56,18 +56,18 @@ if [ -h \${GPHOME}/../hawq ]; then
 fi
 EOF
 
-# OSX does not need JAVA_HOME 
+# OSX does NOT have DYLD_LIBRARY_PATH, add it
 if [ "${PLAT}" = "Darwin" ] ; then
-cat << EOF
-PATH=\$GPHOME/bin:\$GPHOME/ext/python/bin:\$PATH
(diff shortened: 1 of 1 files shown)


tools/bin/generate-greenplum-path.sh
```diff
@@ -56,26 +56,26 @@ if [ -h \${GPHOME}/../hawq ]; then
 fi
 EOF
 
-# OSX does not need JAVA_HOME 
+# OSX does NOT have DYLD_LIBRARY_PATH, add it
 if [ "${PLAT}" = "Darwin" ] ; then
-cat << EOF
-PATH=\$GPHOME/bin:\$GPHOME/ext/python/bin:\$PATH
-DYLD_LIBRARY_PATH=\$GPHOME/lib:\$GPHOME/ext/python/lib:\$DYLD_LIBRARY_PATH
+    cat << EOF
```
hornn: remove spaces, it's confusing because the other lines in the if block don't have them.

hornn:
Looks ok, but it's not working on my MAC. Let me know if you want me to push it.

tools/bin/generate-greenplum-path.sh
xinzweb: Good catch, fixed.

PR #271, opened 2016-01-15
HAWQ-289. Support server side and fix issue in client side for hawqfaultinjector

realdawn:


liming01:
Looks good. +1

huor:
The fix looks good. +1

PR #272, opened 2016-01-15
HAWQ-346. Fixed assertion failure when dispatching EXPLAIN ANALYZE wi…

liming01:
…th window function in optimizer off mode


jianlirong:
+1, LGTM.

huor:
The fix looks good. +1

PR #273, opened 2016-01-15
Fallback to planner for command: ALTER TABLE SET DISTRIBUTED BY

vraghavan78:
When the user fires the following query: 
```
 alter table redist_test set with (reorganize=true) distributed  randomly;
```
The intention of the user is to force redistribute the tuple so that there is no skew. 

Under the hood, GPDB first creates a temp table via CTAS and then replaces the table with the temp table. Given that source table and the temp table are both random, Orca says random satisfies random and therefore no tuple relocation via motions is needed. This changes the expected behavior. 

The fix inside Orca is quite extensive, so the immediate stop gap is to fall back to the planner when we execute the CTAS.

@xinzweb @hsyuan  please take a look


hsyuan:
LGTM.

xinzweb:
LGTM

PR #275, opened 2016-01-18
HAWQ-345. Can not drop a resource queue that is busy

jiny2:


huor:
The fix looks good. +1

linwen:
+1

PR #276, opened 2016-01-18
HAWQ-347. Fixed dead lock between pg_filespace lock and PersistentObj…

liming01:
…Lock

In hawq1.x, segment on master node will use local disk, segments on other segment nodes will use hdfs. However in hawq2.0 we make all segment use hdfs. So now only the pg_system file space use local hdfs.


wangzw:
+1

PR #278, opened 2016-01-19
Hawq-353 fix the bug of segment number mismatch in master's RM

linwen:
please review, thanks!


huor:
The fix looks good. +1

jiny2:
The fix looks good. +1

PR #279, opened 2016-01-19
Hawq 317

hornn:
Getting delegation token for PXF directly from pg_filespace_entry which stores the HDFS location of the HAWQ instance. 
This is similar to how delegation is token is retrieved for regular HAWQ table.
The advantage of this approach is that no special treatment is needed for HA or Isilon cases - the location of HDFS in the catalog table already reflects it. (In HA case the location will be the nameservice, and in Isilon it will be the location of the SmartConnect).

@shivzone @sansanichfb


src/backend/access/external/hd_work_mgr.c
```diff
@@ -900,36 +900,30 @@ static void init_client_context(ClientContext *client_context)
  * is created. We cannot use that token because hd_work_mgr.c code gets 
  * executed before a portal is created.
  *
- * The function uses a hdfs uri in the form of hdfs://host:port/
- * where port is 8020. In the case of HA the function uses the form
- * hdfs://nameservice/
- *
- * TODO 8020 is hard-coded. Must find the NameNode port somehow.
+ * The function uses a hdfs uri in the form of hdfs://host:port/path.
```
shivzone: hdfs://host:port/path or hdfs://nameservice/path

src/backend/access/external/pxfutils.c
```diff
@@ -47,6 +52,24 @@ void port_to_str(char **port, int new_port)
...
 /*
+ * get_hdfs_location_from_filespace
+ *
+ * Get hdfs location from pg_filespace_entry
+ * The returned path needs to be pfreed by the caller.
+ */
+void get_hdfs_location_from_filespace(char** path)
+{
+	Assert(NULL != path);
+	Oid dtsoid = get_database_dts(MyDatabaseId);
```
shivzone: what is MyDatabaseId and Where is it set ?

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -2872,31 +2872,27 @@ static char* GetExtTableFirstLocation(Datum *array)
...
-	StringInfoData hdfs_uri;
+	char* dfs_address = NULL;
 
-	initStringInfo(&hdfs_uri);
-	GPHDUri *gphd_uri = parseGPHDUri(uri);
-    if (gphd_uri->ha_nodes)
-        appendStringInfo(&hdfs_uri, "hdfs://%s/", gphd_uri->ha_nodes->nameservice);
-    else
-        appendStringInfo(&hdfs_uri, "hdfs://%s:8020/", gphd_uri->host);
+	if (!enable_secure_filesystem)
```
shivzone: we didn't have this check earlier ?

src/backend/access/external/pxfutils.c
hornn: It holds the OID of the current session's db. It is usually used to get the db's name or any other properties related to it, such as the filespace.
It is defined in globals.c and set in postinit.c's InitPostgres() which is called when a session is opened (see postgres.c's PostgresMain()).

src/backend/cdb/cdbquerycontextdispatching.c
hornn: Surprisingly, no. The check was only done inside prepareDispatchedCatalogFileSystemCredential(), which this function calls.

shivzone:
+1

src/backend/access/external/pxfutils.c
```diff
@@ -47,6 +52,24 @@ void port_to_str(char **port, int new_port)
...
 
 /*
+ * get_hdfs_location_from_filespace
+ *
+ * Get hdfs location from pg_filespace_entry
+ * The returned path needs to be pfreed by the caller.
+ */
+void get_hdfs_location_from_filespace(char** path)
+{
+	Assert(NULL != path);
```
wangzw: If I remember correctly a table can specify a table space that is different from the table space used by the database. Will it cause trouble with this code?

hornn: This code gets the file space of the current DB (MyDatabaseId), so even if a specific table is configured to a different file space it shouldn't affect it.
Do you think it would be safer to get the file space of DB 'template1' instead of the current session's DB? I followed the examples in other places in the code to get the file space, and they either use the current table's DB or MyDatabaseId.

wangzw:
LGTM +1

huor:
Nice fix. It would be perfect if a commit message is provided.

Here is the convention for commit message in [Contributing to HAWQ](https://cwiki.apache.org/confluence/display/HAWQ/Contributing+to+HAWQ). Also pasted below:
```
# Make changes and commit to local. Please format commit message as follows and feel free to add additional comments.
# <jira ticket number>. <Commit Message>
# Example: HAWQ-1002. Add awesome feature
```

cwelton:
This PR has been open for 29 days, and has enough +1 votes to move forward.  Anything blocking it from being resolved?

hornn:
Awaiting verification on Isilon cluster.

shivzone:
As Noa said we are waiting on Jemish's certification of this patch against a secure Isilon cluster.

GodenYao:
Jemish is on paternity leave until Mar 1st. He will pick up the verification and update us.

shivzone:
@hornn this patch is required to fix https://issues.apache.org/jira/browse/HAWQ-462
We did the RCA for the hcatalog failure against secure cluster and this patch fixes that issue as well.
Can you please merge it ?

PR #281, opened 2016-01-20
HAWQ-344. When resource queue capacity is shrunk, deadlock detection …

jiny2:
When resource queue capacity is shrunk, deadlock detection maybe not triggered.


huor:
The fix looks good. +1

linwen:
+1

PR #283, opened 2016-01-21
HAWQ-290. Add faultinjector for test.

realdawn:
As we removed CHECKPOINT_START_LOCK , we need to set fault_injector here, then we can test it.


liming01:
Looks good. +1.

huor:
The fix looks good. +1

PR #285, opened 2016-01-21
HAWQ-155. Fix bug for out of range access to the hdfs file as scan a …

realdawn:
The fileRemainderLen was calculated incorrectly before. 
If the isUseSplitLen is true, it should be the bufferedRead.splitLen-*headerOffsetInFile, 
So the bug will cause the out of range read.


liming01:
Good fix. +1.

ztao1987:
+1

PR #286, opened 2016-01-21
HAWQ-216. Fix bug for built-in function gp_update_global_sequence_entry

realdawn:
Parameter sequenceVal was defined as a int8 number. But it is should be a int64 number.


huor:
The fix looks good. +1

liming01:
Looks good. +1.

PR #287, opened 2016-01-22
HAWQ-229. Forbid altering external table through altering table.

realdawn:
Forbid altering external table through altering table.


huor:
The fix looks good. +1

liming01:
Look good. +1.

PR #288, opened 2016-01-22
HAWQ-274. Fix coverity error

linwen:
Please review.


huor:
The fix looks good. +1

PR #291, opened 2016-01-23
HAWQ-362. Master RM should Recalculate Segment's Allocation Status if Segment's RM is restart

linwen:
Please review, Thanks!

### DIFF
--- a/src/backend/resourcemanager/resourcepool.c
+++ b/src/backend/resourcemanager/resourcepool.c
@@ -896,7 +896,14 @@ int addHAWQSegWithSegStat(SegStat segstat, bool *capstatchanged)
 		uint8_t oldStatus = segresource->Stat->FTSAvailable;
 		bool statusChanged = oldStatus != segstat->FTSAvailable;
 
-		/* Check if RM process is restarted in this segment */
+		/*
+		 * Check if RM process is restarted in this segment.
+		 * If the latest reported RM process startup timestamp doesn't
+		 * match the previous, master RM consider segment's RM process
+		 * has restarted.
+		 * In rare case, the system's time is reset and segment's RM process
+		 * happen to get a same timestamp with previous one.
+		 */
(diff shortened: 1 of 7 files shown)


src/backend/resourcemanager/resourcepool.c
```diff
@@ -893,6 +893,28 @@ int addHAWQSegWithSegStat(SegStat segstat, bool *capstatchanged)
...
+			 * This segment's RM process has restarted,
+			 * we should clean up old status, so mark it down.
+			 */
+			if (oldStatus == RESOURCE_SEG_STATUS_AVAILABLE && !statusChanged)
+			{
+				segstat->FTSAvailable = RESOURCE_SEG_STATUS_UNAVAILABLE;
+				statusChanged = true;
+			}
+			segresource->Stat->RMStartTimestamp = segstat->RMStartTimestamp;
+			elog(LOG, "Master RM finds segment:%s 's RM process has restarted. "
```
jiny2: Suggest to report as a WARNING log.

jiny2:

src/backend/resourcemanager/resourcepool.c
```diff
@@ -893,6 +893,28 @@ int addHAWQSegWithSegStat(SegStat segstat, bool *capstatchanged)
 	else {
 		segresource = getSegResource(segid);
 		Assert(segresource != NULL);
+		uint8_t oldStatus = segresource->Stat->FTSAvailable;
+		bool statusChanged = oldStatus != segstat->FTSAvailable;
+
+		/* Check if RM process is restarted in this segment */
+		if (segresource->Stat->RMStartTimestamp != segstat->RMStartTimestamp)
```
huor: A comment here to illustrate the idea would be better. It is also good to mention the risk that below three condition hold:
1. rmseg reset
2. system time reset on segment node
3. segresource->Stat->RMStartTimestamp == segstat->RMStartTimestamp

huor:

PR #292, opened 2016-01-24
HAWQ-357. Track how many times a segment can not get expected contain…

jiny2:
HAWQ RM will generate WARNING log for each segment that can not get expected quantity of YARN containers.

"Resource manager detects segment %s hasn't gotten expected quantity of resource containers for %d times."


linwen:
Looks good.

huor:
The fix looks good. +1

amyrazz44:
LGTM+1

PR #293, opened 2016-01-25
HAWQ-325. Fix a bug, if the number of allocated containers is less th…

linwen:
Fix the bug, a map reduce job occupied 2 nodes's 12 containers, there is 1 node/6 containers available, hawq failed to allocate resource from yarn.

Please review, thanks!


huor:
The fix looks good. +1

jiny2:

PR #295, opened 2016-01-26
HAWQ-147. Fix a bug of creating database failed due to parquet table …

ztao1987:
Bug HAWQ-147: Create Parquet table and insert data in template1 would cause CREATE DATABASE fail.

Reproduce steps:
$ psql -d template1
psql (8.2.15)
Type "help" for help.

template1=# CREATE TABLE foo (
template1(# a INT)
template1-# WITH (appendonly=true, orientation=parquet);
CREATE TABLE
template1=# INSERT INTO foo VALUES(1);
INSERT 0 1
template1=# CREATE DATABASE gptest;
ERROR: Append-Only relation 'foo' gp_relation_node entry for segment file #1 without an aoseg entry (case #2) (cdbdatabaseinfo.c:1554)
template1=#

### DIFF
--- a/src/backend/cdb/cdbdatabaseinfo.c
+++ b/src/backend/cdb/cdbdatabaseinfo.c
@@ -1037,7 +1037,7 @@ DatabaseInfo_HandleAppendOnly(
 		if (dbInfoRel == NULL)
 			break;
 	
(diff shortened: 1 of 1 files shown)


zhangh43:
LGTM +1

PR #297, opened 2016-01-26
HAWQ-364. Make resource manager dynamically adjust minimum YARN conta…

jiny2:
user can set the minimum YARN container water level through guc hawq_rm_min_resource_perseg. 
1) If the segment capacity is smaller than expected level, this water level is lowered down;
2) If the queue capacity can not make each segment have at least hawq_rm_min_resource_perseg YARN containers, this water level is lowered down.
This improvement can make HAWQ able to dynamically fit the changing YARN cluster and avoid user wrong config caused unexpected behaviors.

If the setting is adjusted by hawq rm, one WARNING log is generated for issue tracking.

"WARNING","01000","Resource manager adjusts minimum global resource manager container count in each segment from 100 to 8.",,,,,,,0,,"resqueuemanager.c",2530,


huor:
The fix looks good. +1

PR #300, opened 2016-01-27
HAWQ-366. Analyse data locality ratio for TPCH query in SFO.

zhangh43:
Data locality ratio for TPCH query(such as Q2) in SFO is low when some machines are down.
The reason for this is that we change net_disk_ratio to 1.0 which leads to we prefer the virtual segment with minimum allocated size plus current split size, but not LOCAL!
To prevent this happens we need to enlarge net_disk_ratio to 1.01. Then if the current size of two vsegs are almost the same, we can choose the LOCAL one.


ztao1987:
+1

yaoj2:
+1

PR #301, opened 2016-01-28
HAWQ-367. Introduce seg_max_connections guc to set segment max_connections setting

jiny2:


src/backend/utils/misc/guc.c
```diff
@@ -4617,6 +4617,20 @@ static struct config_int ConfigureNamesInt[] =
...
+	 */
+	{
+		{"seg_max_connections", PGC_POSTMASTER, CONN_AUTH_SETTINGS,
+			gettext_noop("Sets the maximum number of concurrent connections in a segment."),
+			NULL
+		},
+		&SegMaxBackends,
+		1600, 80, MAX_MAX_BACKENDS, NULL, NULL
+	},
+
```
yaoj2: For default value, according to 1.x  the default is 3 times master value. So we use 200 as master value then 200*3 * 8 we use 4800 as default value.

jiny2: Ok, I fix it.

huor:
LGTM. +1

linwen:
Looks good

PR #302, opened 2016-01-28
HAWQ-178: Add JSON plugin support in code base

tzolov:
Add `pxf-json` and registers it along the pxf-hdfs, pxf-hive and pxf-hbase plugins.  

Note: I've copied the PxfUnit so i can use the pxf-json unit testing. In log term the PxfUnit should be placed in a standalone project.

Plugin has manually been tested against HAWQ-2.0.0-beta.


pxf/build.gradle
```diff
@@ -64,7 +64,7 @@ subprojects { subProject ->
             "-Xlint:finally", "-Xlint:overrides", "-Xlint:path", "-Xlint:processing", "-Xlint:static", "-Xlint:try",
             "-Xlint:-fallthrough", "-Xlint:-deprecation", "-Xlint:-unchecked", "-Xlint:-options"
     ]
-
+
```
hornn: remove white spaces?

pxf/build.gradle
```diff
@@ -286,6 +286,29 @@ project('pxf-hive') {
...
 
+project('pxf-json') {
+	dependencies {
+		compile(project(':pxf-hdfs'))
+		compile "org.apache.commons:commons-lang3:3.0"
+
+		testCompile 'pl.pragmatists:JUnitParams:1.0.2'
+		configurations {
+			// Remove hive-exec from unit tests as it causes VerifyError
+			testRuntime.exclude module: 'hive-exec'
```
hornn: is this necessary?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -128,4 +128,26 @@ under the License.
...
+		<description>A profile for JSON data, one JSON record per line
+		</description>
+		<plugins>
+			<fragmenter>org.apache.hawq.pxf.plugins.hdfs.HdfsDataFragmenter</fragmenter>
+			<accessor>org.apache.hawq.pxf.plugins.json.JsonAccessor</accessor>
+			<resolver>org.apache.hawq.pxf.plugins.json.JsonResolver</resolver>
+		</plugins>
+	</profile>
+	<profile>
+		<name>JSON-PP</name>
```
hornn: what does PP stand for?
Other profiles use CamelCase as their naming standing, like HdfsTextSimple or HiveRc - perhaps change it to something like that?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+			return NullWritable.get();
+		}
+
+		@Override
+		public long getPos() throws IOException {
+			return rdr.getPos();
+		}
+
+		@Override
+		public boolean next(Text key, NullWritable value) throws IOException {
```
hornn: could you please add documentation for this function?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+
+		@Override
+		public float getProgress() throws IOException {
+			return rdr.getProgress();
+		}
+	}
+
+	public static class JsonRecordReader implements RecordReader<Text, NullWritable> {
+
+		private Logger LOG = Logger.getLogger(JsonRecordReader.class);
```
hornn: We use apache.commons.logging.Log in PXF:
import org.apache.commons.logging.Log;
import org.apache.commons.logging.LogFactory;
 private static final Log LOG = LogFactory.getLog(JsonRecordReader.class);

pxf/pxf-json/src/test/resources/tweets-small.json
```diff
@@ -0,0 +1,4 @@
+{"created_at":"Fri Jun 07 22:45:02 +0000 2013","id":343136547115253761,"text":"REPAIR THE TRUST: REMOVE OBAMA\/BIDEN FROM OFFICE. #IRS #DOJ #NSA #tcot","source":"web","user":{"id":"26643566","name":"SpreadButter","screen_name":"SpreadButter","location":"Austin, Texas"},"entities":{"hashtags":["tweetCongress","IRS"]}}
```
GodenYao: better use non-political content for test example.
I think you've fixed it in PDF guide but missed here :)
https://issues.apache.org/jira/secure/attachment/12785045/PXFJSONPluginforHAWQ2.0andPXF3.0.0.pdf

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+
+		private JsonStreamReader rdr = null;
+		private long start = 0, end = 0;
+		private float toRead = 0;
+		private String identifier = null;
+		private Logger log = Logger.getLogger(JsonRecordReader.class);
```
hornn: why having two loggers?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+
+		public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
+			log.info("JsonRecordReader constructor called.  Conf is " + conf + ". Split is " + split);
```
hornn: how often is this ctor called? consider changing all logs to debug level

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+			return (float) rdr.getBytesRead() / toRead;
+		}
+	}
+
+	public static synchronized JsonNode decodeLineToJsonNode(String line) {
+
+		try {
+			return mapper.readTree(line);
+		} catch (JsonParseException e) {
+			e.printStackTrace();
```
sansanichfb: Using log.debug instead of?

hornn: perhaps replace e.printStackTrace() with LOG.warning()?

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,173 @@
...
+	public void testSmallTweets() throws Exception {
+
+		List<String> output = new ArrayList<String>();
+
+		output.add("Fri Jun 07 22:45:02 +0000 2013,343136547115253761,REPAIR THE TRUST: REMOVE OBAMA/BIDEN FROM OFFICE. #IRS #DOJ #NSA #tcot,SpreadButter,tweetCongress,,");
+		output.add("Fri Jun 07 22:45:02 +0000 2013,343136547123646465,@marshafitrie dibagi 1000 aja sha :P,patronusdeadly,,,");
+		output.add("Fri Jun 07 22:45:02 +0000 2013,343136547136233472,Vaga: Supervisor de Almoxarifado. Confira em http://t.co/hK5cy5B2oS,NoSecrets_Vagas,,,");
+		output.add("Fri Jun 07 22:45:03 +0000 2013,343136551322136576,It's Jun 7, 2013 @ 11pm ; Wind = NNE (30,0) 14.0 knots; Swell = 2.6 ft @ 5 seconds....,SevenStonesBuoy,,-6.1,50.103");
+
+		super.assertOutput(new Path(System.getProperty("user.dir") + "/" + "src/test/resources/tweets-small.json"),
```
sansanichfb: Probably we can use File.separator instead of "/".

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
```diff
@@ -0,0 +1,226 @@
...
+			rdr.close();
+		}
+
+		@Override
+		public float getProgress() throws IOException {
+			return (float) rdr.getBytesRead() / toRead;
+		}
+	}
+
+	public static synchronized JsonNode decodeLineToJsonNode(String line) {
```
hornn: why does the mapper have to be static? wouldn't it be better to make it a member of the JsonInputFormat, and then not worry about synchronizing this function?
When handling multiple requests, it probably has a performance implication.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonAccessor.java
```diff
@@ -0,0 +1,66 @@
...
+	public static final String ONERECORDPERLINE_PARAM = "ONERECORDPERLINE";
+
+	private String identifier = "";
+	private boolean oneRecordPerLine = true;
+
+	public JsonAccessor(InputData inputData) throws Exception {
+		super(inputData, new JsonInputFormat());
+
+		if (!StringUtils.isEmpty(inputData.getUserProperty(IDENTIFIER_PARAM))) {
+			identifier = inputData.getUserProperty(IDENTIFIER_PARAM);
```
hornn: is it ok not to have an identifier? what happens if we leave it as ""?
If it's an error, it should probably be handled here by throwing an exception.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonAccessor.java
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		if (!StringUtils.isEmpty(inputData.getUserProperty(ONERECORDPERLINE_PARAM))) {
+			oneRecordPerLine = Boolean.parseBoolean(inputData.getUserProperty(ONERECORDPERLINE_PARAM));
+		}
+	}
+
+	@Override
+	protected Object getReader(JobConf conf, InputSplit split) throws IOException {
+		conf.set(JsonInputFormat.RECORD_IDENTIFIER, identifier);
```
hornn: same comment - if identifier is "", should we set it here or not?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,259 @@
...
+			}
+		}
+
+		return list;
+	}
+
+	/**
+	 * Iterates down the root node to the prior JSON node. This node is used
+	 * 
+	 * @param root
```
hornn: please fill up all the parameters in the javadoc. It causes the build to fail when compiling with java8.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonAccessor.java
GodenYao: Looking at description of JSON-PP plugin:

> A profile for JSON data, where a single JSON record spans multiple lines. Must specify the IDENTIFIER parameter in DDL.

Identifier is required?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,259 @@
...
+					}
+				}
+			}
+		}
+
+		return list;
+	}
+
+	/**
+	 * Iterates down the root node to the prior JSON node. This node is used
```
hornn: unfinished sentence?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,259 @@
...
+
+	/**
+	 * Gets a boolean value indicating if this column is an array index column
+	 * 
+	 * @param projs
+	 *            The array of JSON projections
+	 * @throws ArrayIndexOutOfBoundsException
+	 */
+	private boolean isArrayIndex(String[] projs) {
+		return projs[projs.length - 1].contains("[") && projs[projs.length - 1].contains("]");
```
hornn: perhaps better to check using a regex that element name is `.*\[[0-9]+\]`

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,259 @@
...
+	/**
+	 * Gets the array index from the given String array of JSON projections, parsed from the ColumnDescriptor's name
+	 * 
+	 * @param projs
+	 *            The array of JSON projections
+	 * @return The index
+	 * @throws ArrayIndexOutOfBoundsException
+	 */
+	private int getArrayIndex(String[] projs) {
+		return Integer.parseInt(projs[projs.length - 1].substring(projs[projs.length - 1].indexOf('[') + 1,
```
hornn: perhaps use a variable for `String arrayElem = projs[projs.length - 1]` - will make the code shorter.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,259 @@
...
+			oneField.val = null;
+		} else {
+			switch (type) {
+			case BIGINT:
+				oneField.val = val.asLong();
+				break;
+			case BOOLEAN:
+				oneField.val = val.asBoolean();
+				break;
+			case BPCHAR:
```
hornn: BPCHAR should be treated like VARCHAR and TEXT - it's more than one character.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,89 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+
+public class JsonStreamReader extends BufferedReader {
+
+	private StringBuilder bldr = new StringBuilder();
+	private String identifier = null;
+	private long bytesRead = 0;
+
+	public JsonStreamReader(String identifier, InputStream strm) {
```
hornn: please add javadoc explaining what is the role of the identifier

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,89 @@
...
+		super(new InputStreamReader(strm));
+		this.identifier = identifier;
```
hornn: throw an exception if it's empty or null?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	public String getJsonRecord() throws IOException {
+		bldr.delete(0, bldr.length());
+
+		boolean foundRecord = false;
+
+		int c = 0, numBraces = 1;
+		while ((c = super.read()) != -1) {
```
hornn: better to replace -1 with EOF const

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,89 @@
...
+			++bytesRead;
+			if (!foundRecord) {
+				bldr.append((char) c);
+
+				if (bldr.toString().contains(identifier)) {
+					forwardToBrace();
```
hornn: what if forwardToBrace runs to the end of the stream without finding any '{'?
In that case wouldn't we return a "record" that only consists of "{"?

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
```diff
@@ -0,0 +1,53 @@
...
+		rdr = new JsonStreamReader("menuitem", new FileInputStream(FILE));
+	}
+
+	@Test
+	public void testReadRecords() throws IOException {
+		int count = 0;
+		String record = null;
+		while ((record = rdr.getJsonRecord()) != null) {
+			++count;
+			System.out.println(record);
```
hornn: instead of printing, better to assert the content of each record.

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class JsonStreamReaderTest {
+
+	private JsonStreamReader rdr = null;
+
+	private static final String FILE = "src/test/resources/sample.json";
+
+	@Before
+	public void setup() throws FileNotFoundException {
+		rdr = new JsonStreamReader("menuitem", new FileInputStream(FILE));
```
hornn: please add @After function to close the streams

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+import org.apache.hawq.pxf.plugins.json.JsonStreamReader;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class JsonStreamReaderTest {
```
hornn: Good test - would it be possible to add a negative test with a malformatted json file?

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/PxfUnit.java
```diff
@@ -0,0 +1,666 @@
...
+	 * 
+	 * @param input
+	 *            Input records
+	 * @param expectedOutput
+	 *            File containing output to check
+	 * @throws Exception
+	 */
+	public void assertOutput(Path input, Path expectedOutput) throws Exception {
+
+		BufferedReader rdr = new BufferedReader(new InputStreamReader(FileSystem.get(new Configuration()).open(
```
hornn: close the resources in the end of the function

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/PxfUnit.java
```diff
@@ -0,0 +1,666 @@
...
+				sourceData = "/" + sourceData;
+			}
+			paramsMap.put("X-GP-DATA-DIR", sourceData);
+			paramsMap.put("X-GP-FRAGMENT-METADATA", fragNode.get("metadata").getTextValue());
+			paramsMap.put("X-GP-DATA-FRAGMENT", Integer.toString(i++));
+			inputs.add(new LocalInputData(paramsMap));
+		}
+	}
+
+	private JsonNode decodeLineToJsonNode(String line) {
```
hornn: code duplication? can use the same function in json accessor (perhaps move to JsonUtils class?)

pxf/build.gradle
tzolov: Should we remove all whitespaces in the build.gradle file?  There are quite a few.

pxf/build.gradle
tzolov: @hornn What do you men? do we need the pxf-json project? or do we need some of the dependencies/configurations?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
tzolov: JSON-PP differs from JSON only by the hard coded property `onerecordperline=false`.  
I will remove it though in favor of JSON + explicit parameter in the DDL definition.
About the name shall i change it from `JSON` to `Json`?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
tzolov: As those methods implement `RecordReader` i guess inherit the the doc (`{@inheritDoc}`) should be enough as documentation? Only `next()` seems to deserve additional doc.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
tzolov: yep will change the log level to `debug`

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
tzolov: Will unify them under the apache commons Log.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
tzolov: addressed

pxf/pxf-json/src/test/resources/tweets-small.json
tzolov: sure, i will clean the tweets text content

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonExtensionTest.java
tzolov: yep

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
tzolov: will do

tzolov:
@hornn , @GodenYao 
The `pxf-json` code was implemented by @adamjshook. In this PR i've barely ported it to the HAWQ pxf project structure. My idea was to port the existing code and then improve it if needed.
But the excellent comments above made me review the code and find a significant issue with the JsonRecordReader  - e.g. the multiline JSON objects support (also called Pretty Print - PP). Current implementation will not work when the JSON documents spans multiple HDFS Splits! 
So i will remove the multiline-JSON code from the PR, leaving in only the LineRecordReader version (e.g assuming that json object per line). 
Also i will open a discussion in the dev mailing list about how to handle in PXF documents that spans across Splits.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: addressed!

adamjshook:
I thought I had handled this case in the `JsonRecordReader` code.  The `JsonStreamReader` has no knowledge of the 'end' location of the FileSplit and will read until it completes a JSON record on the call to `getJsonRecord`.  The `JsonInputStream` only returns false on a `next` call when the number of bytes read by the stream reader is greater than or equal to `end - start`.  It will ignore the split boundaries and continue to reading to make a full record.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: will chanage the array name/index retrieval to to use the Pattern.compile("(.*)\\[([0-9]+)\\]") matching groups instead

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: will chanage the array name/index retrieval to to use the Pattern.compile("(.*)\\[([0-9]+)\\]") matching groups instead

adamjshook:
Actually, scratch that, I don't even think you can comfortably build JSON records that span multiple lines.  Unlike XML, there is no cleanly identifiable "end tag" of a JSON record.  I think the intent was to set the isSplitable to false and I never did.  The preferred method of using this was to have one JSON record per line.

tzolov:
@adamjshook, it's nice to hear from you!  
You are right the `JsonRecordReader`,`JsonStreamReader` reads beyond the `Split` boundries. I was missleaded by the `JsonInputFormat` which is never been used. The `HdfsSplittableDataAccessor` (the JsonAccessor parent) doesn't use the InptuFormat and getSplits\getRecordReader methods are not called. 
The `HdfsSplittableDataAccessor` generates exactly **one** split!

tzolov:
Alhtough there are no end-tags (like XLM) one can use the begingin of the next `identifer` as end-tag new-tab signal. If not mistaken this is how current implementation is meant to work? I'm also exploring this project related [json-mapreduce](https://github.com/alexholmes/json-mapreduce) that uses a similar approach.

adamjshook:
Good to hear from you as well Christian :)

I think you're right on the `identifier `as the means to have an end-tag and new-tag.  It's slowly coming back.  The `StreamReader `seeks until it finds the identifier, then flushes the `StringBuilder` and appends characters until the final end brace is met, accounting for any nested begin brace/end brace combinations.  The big problem that happens is if the identifier is embedded as a string inside a JSON record, it would create an incomplete record.  If your identifier was `record` in the below example where the split is the horizontal line, bad stuff would happen.  This is why the expectation was one file = one split.

```
record : {
--------------------------
    text : "this is a record"
}
```

pxf/build.gradle
hornn: sure :)

pxf/build.gradle
hornn: I mean the exclude part - it refers to hive module that is not a dependency for this plugin.
Could you try to remove the configurations block and run make to see if it works?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
hornn: Both JSON and Json is fine, but I like Json better.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonInputFormat.java
hornn: I agree, the rest is more clear. Perhaps add a comment also for the class, explaining the general flow.

tzolov:
Actually I've been thinking about similar use cases that the current stream reader doesn’t handle. For this we can re-implementing the stream reader using the Jackson JSON Stream API or prehaps to resue some parts of the [json-mapreduce](https://github.com/alexholmes/json-mapreduce) project.

hornn:
It sounds very similar to CSV with quoted data, which is not splittable. The way we do it today is by ensuring we process each file by a single accessor, even if it actually consists of multiple splits. (see HdfsTextMulti profile and [QuotedLineBreakAccessor](https://github.com/apache/incubator-hawq/blob/master/pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/QuotedLineBreakAccessor.java)). The problem, of course, is that we lose parallelism and performance.
@tzolov, if it requires too much re-writing I agree that we can make it in two stages - first the splittable case (one record per line), and then the more complex cases.
@adamjshook, good to hear from you :

GodenYao:
@adamjshook , thanks for your contribution to JSON plugin and now everything is open sourced , so feel free to send us your PR if any!

adamjshook:
@GodenYao You're very welcome!  I'll try and find some time to contribute the various prototype connectors I have written.  @tzolov shouldn't have to deal with all of my code!

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/PxfUnit.java
tzolov: The PxfUnit should go in its own standalone project and be used only for testing. So better tackle any code duplications in PxfUnit once it is factored out from the pxf-json.

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
tzolov: So Json it is

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/PxfUnit.java
tzolov: done

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
tzolov: done

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
tzolov: done

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonStreamReaderTest.java
tzolov: addressed

pxf/build.gradle
tzolov: done

pxf/build.gradle
tzolov: done

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonAccessor.java
tzolov: JsonAccessor supports two modes: 
1. json object per line -> LineRecordReader
2. multiline json objects that can span beyond split's boundries. -> JsonRecordReader + JsonStreamReader
Second mode (used to be called JSON-PP e.g. Pretty Print) requires an json element name (e.g. identifier) to be set to identify the boundries of a json record. 
I've removed the JSON-PP instead the Json profile without IDENTIFIER set will default to mode 1 (e.g. json per line) and if the identifier paramter is set switches to mode 2 (multiline json records named by the identifer param).
Will update the documentation to reflect those changes.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonAccessor.java
tzolov: moved the set code to the JsonRecordReader mode only

tzolov:
Updated the PR. Commits should address most of the comments above. I've simplified the code. Replaced the JsonLineRecordReader in favor of the OOB LineRecordReader, removed the onerecordperline property (using the identifier property as a flag instead) and many more. 

Brief [how to use doc](https://issues.apache.org/jira/secure/attachment/12785629/PXFJSONPluginforHAWQ2.0andPXF3.0.0v.2.pdf) is attached to the JIRA ticket. 

Still have not resolved the use case described by @adamjshook above.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,281 @@
...
+		return list;
+	}
+
+	/**
+	 * Iterates down the root node to the prior JSON node. This node is used
+	 * 
+	 * @param root
+	 * @param projs
+	 * @return
+	 */
```
hornn: /Users/nhorn/dev/hawq/apache-hawq/pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java:122: warning - @return tag has no arguments.
1 warning

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,175 @@
+package org.apache.pxf.hawq.plugins.json;
```
hornn: in all tests, the package name is wrong.
should be org.apache.hawq.pxf....

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,121 @@
...
+	 *            JSON document input stream.
+	 */
+	public JsonStreamReader(String identifier, InputStream inputStream) {
+		super(new InputStreamReader(inputStream));
+		this.identifier = identifier;
+	}
+
+	/**
+	 * @return Returns next JSON object identified by the {@link JsonStreamReader#identifier} parameter or Null if no
+	 *         more object are available.
```
hornn: typo objects?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,121 @@
...
+				if (numBraces == 0) {
+					break;
+				}
+			}
+		}
+
+		if (foundRecord && numBraces == 0) {
+			return bldr.toString();
+		} else {
+			return null;
```
hornn: if bldr is not empty, should we perhaps log a warning/error?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
```diff
@@ -0,0 +1,121 @@
...
+
+	/**
+	 * @param identifier
+	 *            JSON object name to extract.
+	 * @param inputStream
+	 *            JSON document input stream.
+	 */
+	public JsonStreamReader(String identifier, InputStream inputStream) {
+		super(new InputStreamReader(inputStream));
+		this.identifier = identifier;
```
hornn: I'm sorry if I already mentioned it, but I think we should check also here that identifier is not null or empty, even if it's checked by the accessor.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.codehaus.jackson.JsonFactory;
+import org.codehaus.jackson.JsonNode;
+import org.codehaus.jackson.map.ObjectMapper;
+
+/**
+ * Utility that helps to convert JSON text into {@link JsonNode} object
+ *
+ */
+public class JsonUtil {
```
hornn: in the other plugins, we usually put the utils classes under their own package org.apache.hawq.pxf.plugins.json.utilities.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
tzolov: yes it makes sense and i will add it. But mind that the current stream reader is not aware of the semantic of the json document and can miss chunks from the documents if later is  malformed. We need to implement a proper stream reading to minimize such issues.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
tzolov: sure

pxf/pxf-json/src/test/java/org/apache/pxf/hawq/plugins/json/JsonExtensionTest.java
tzolov: ops. good spot!. resolved

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: resolved

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -0,0 +1,166 @@
...
+		do {
+			// Exit condition (end of block/file)
+			if (streamReader.getBytesRead() >= (end - start)) {
+				return false;
+			}
+
+			keepGoing = false;
+			String record = streamReader.getJsonRecord();
+			if (record != null) {
+				if (JsonUtil.decodeLineToJsonNode(record) == null) {
```
hornn: We call decodeLineToJsonNode twice for each record - one time here (accessing), and one time in the resolver. I don't know if it is a heavy operation or not, but it is slow because this function is synchronized, so multiple requests will be queued.
Consider removing this logic here - and failing in the resolver. We also have a notion of BadRecordException that can be used when there is a malformatted record.

hornn:
Looks really good. I will try to test the plugins using your doc.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonRecordReader.java
tzolov: Did this and removed the JsonUtil all together. Parsing logic is moved directly in the resolver. Also i think it is better to log an error in case of malformed json object but to continue the processing (e.g. no exception is thrown)

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonUtil.java
tzolov: i've removed the JsonUtil as the parsing logic is now part of the resolver (see below)

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonStreamReader.java
tzolov: makes sense

tzolov:
Hopefully the last committ resolves the remaing remarks

tzolov:
Should be done. the jsonFactory is not required as private field anymore.

tzolov:
@hornn, @adamjshook  
As an experimenti've rempimplemented/replaced the `JsonRecordReader`&`JsonStreamReader` with ideas and code boroed from the [json-mapreduce](https://github.com/alexholmes/json-mapreduce) project.  I really like the result. It comes with a (sort of) JsonLexer/Parser that solves some of the shortcomings in the current implementation. For example having the `identifier` in the nested object(s) values will be handled. 
Furthermore they why how the identifier  is used is slightly differently and more powerful. The identifier refers to a member name which it will use to determine the encapsulating object to return.  This is  superior functionality over  what we have at the moment as you can point to multiline json objects that don't have a parent identifier. 

Shall i add this change to this PR or to a separate one after this has been merged? 
Personally i think it is better to add it now as it changes the `identifier` semantics and would be inconvenient for the potential users to learn/unlearn. Alternatively we can drop the current JsonRecordReader form this PR and introduce the new code as an extension i another one. What do you think?

adamjshook:
@tzolov A quick review of the json-mapreduce project looks good to me.  I think it'd be best to roll it in now to avoid any inconvenience as you have stated.

hornn:
I agree. We can afford to wait a little longer for a better plugin.

tzolov:
Here it comes. The improved JsonRecordReader implementation is committed. Mind the new identifier semantics.  Also as a safeguard you can set the max record length allowed to parse. 
The new implementation should support compression as well (not tested)

tzolov:
@hornn , @GodenYao since we are reusing code from the [json-mapreduce](https://github.com/alexholmes/json-mapreduce) (APL2.0) do we have to mention it in the NOTICE section?

GodenYao:
yes @rvs , coincidentally, we're revamping NOTICE file for HAWQ release.
Roman should be able to incorporate this in if you commit this week. Otherwise, we need to add it later.

tzolov:
In the last commit(s): 
* Added tests to cover corner cases like malformed json records or compressed input files. The _combined_ test coverage is > 95%. @hornn is there an integration tests/environment to itest PXF plugins?
* Improved handling of malformed json records/fields. If the json record is not parsable the Resolver will produce a row with empty fields.  If a field in the record is missing then an empty db field is created.
* Factored out the column metadata computation logic and data in a dedicated `ColumnDescriptorCache` class. Now the cached descriptor metadata is created only once inside the `JsonResolver` constructor. Previously it use to be re-computed on each `JsonResolver.getFields()` method call.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/ColumnDescriptorCache.java
```diff
@@ -0,0 +1,165 @@
...
+
+import org.apache.hawq.pxf.api.io.DataType;
+import org.apache.hawq.pxf.api.utilities.ColumnDescriptor;
+
+/**
+ * Helper class used to retrieve all column details relevant for the json processing.
+ */
+public class ColumnDescriptorCache {
+
+	private static Pattern ARRAY_PROJECTION_PATTERN = Pattern.compile("(.*)\\[([0-9]+)\\]");
```
hornn: just to be safe, I think the first group should be `.+` and not `.*`.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/ColumnDescriptorCache.java
```diff
@@ -0,0 +1,165 @@
...
+	private final int arrayIndex;
+	private final boolean isArray;
+
+	public ColumnDescriptorCache(ColumnDescriptor columnDescriptor) {
+
+		this.columnType = DataType.get(columnDescriptor.columnTypeCode());
+
+		this.projections = columnDescriptor.columnName().split("\\.");
+
+		this.isArray = isArrayIndex(projections);
```
hornn: slight improvement - if we encapsulate the whole array logic in a function, we can call `getMatchGroup` once, and use the results to populate this.isArray, this.nodeName and this.arrayIndex.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonRecordReader.java
```diff
@@ -0,0 +1,176 @@
...
+	public JsonRecordReader(JobConf conf, FileSplit split) throws IOException {
+
+		this.jsonMemberName = conf.get(RECORD_MEMBER_IDENTIFIER);
+		this.maxObjectLength = conf.getInt(RECORD_MAX_LENGTH, Integer.MAX_VALUE);
+
+		start = split.getStart();
+		end = start + split.getLength();
+		final Path file = split.getPath();
+		compressionCodecs = new CompressionCodecFactory(conf);
+		final CompressionCodec codec = compressionCodecs.getCodec(file);
```
hornn: Regarding compression, we support it for hdfs files, and the implementation here looks very similar to the one in LineRecordReader. The issue worth mentioning here is that BZip2 Codec is not thread safe. For map-reduce jobs it doesn't matter, because each job is running in its own JVM, but we are handling multiple concurrent requests through tomcat, so that introduced a problem.
For Hdfs files, we "solved" it by forcing the processing of bzip2 files to be done in a single thread. See HdfsUtilities.isThreadSafe() which is called by HdfsSplittableDataAccessor. I see that the JsonAccessor inherits from this class, so I think it should be fine, but it's worth verifying.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,255 @@
...
+	@Override
+	public List<OneField> getFields(OneRow row) throws Exception {
+		oneFieldList.clear();
+
+		String jsonRecordAsText = row.getData().toString();
+
+		JsonNode root = decodeLineToJsonNode(jsonRecordAsText);
+
+		if (root == null) {
+			LOG.warn("Return null-fields row due to invalid JSON:" + jsonRecordAsText);
```
hornn: another possible improvement: create a final list of null fields for the case where the root is null, and immediately return that list here.
That will save us the iteration over the columns and checking for each one if the root is null.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,255 @@
...
+	 *            node to to start the traversal from.
+	 * @param projs
+	 *            defines the path from the root to the desired child node.
+	 * @return Returns the child node defined by the root and projs path.
+	 */
+	private JsonNode getPriorJsonNode(JsonNode root, String[] projs) {
+
+		// Iterate through all the tokens to the desired JSON node
+		JsonNode node = root;
+		for (int j = 0; j < projs.length - 1; ++j) {
```
hornn: just wondering - why is it `proj.length -1` and not `proj.length`?

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -0,0 +1,198 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+
+import org.apache.hawq.pxf.plugins.json.parser.JsonLexer.JsonLexerState;
+
+/**
+ * A simple parser that can support reading JSON objects from a random point in JSON text. It reads from the supplied
+ * stream (which is assumed to be positioned at any arbitrary position inside some JSON text) until it find the first
```
hornn: typo finds

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	private boolean scanToFirstBeginObject() throws IOException {
+		// seek until we hit the first begin-object
+		char prev = ' ';
+		int i;
+		while ((i = inputStreamReader.read()) != EOF) {
+			char c = (char) i;
+			bytesRead++;
+			if (c == '{' && prev != '\\') {
```
hornn: `{` and `\\` should probably be consts

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
```diff
@@ -0,0 +1,198 @@
...
+ * <p/>
+ * It is not recommended to use this with JSON text where individual JSON objects that can be large (MB's or larger).
+ */
+public class PartitionedJsonParser {
+
+	private static final int EOF = -1;
+	private final InputStreamReader inputStreamReader;
+	private final JsonLexer lexer;
+	private long bytesRead = 0;
+	private boolean endOfStream;
```
hornn: for clarify, perhaps explicitly assign `false` to it?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,288 @@
...
+		columnDefs.add(new Pair<String, DataType>("booleanType", DataType.BOOLEAN));
+		columnDefs.add(new Pair<String, DataType>("charType", DataType.CHAR));
+		// columnDefs.add(new Pair<String, DataType>("byteaType", DataType.BYTEA));
+		columnDefs.add(new Pair<String, DataType>("float8Type", DataType.FLOAT8));
+		columnDefs.add(new Pair<String, DataType>("realType", DataType.REAL));
+		columnDefs.add(new Pair<String, DataType>("integerType", DataType.INTEGER));
+		columnDefs.add(new Pair<String, DataType>("smallintType", DataType.SMALLINT));
+		columnDefs.add(new Pair<String, DataType>("bpcharType", DataType.BPCHAR));
+		columnDefs.add(new Pair<String, DataType>("varcharType", DataType.VARCHAR));
+		columnDefs.add(new Pair<String, DataType>("text", DataType.TEXT));
```
hornn: perhaps add a comment saying this field doesn't exist in the file?
btw, does the order have to match between the table definition (columnDefs) and the json file itself? In other words, if the order between bintType and booleanType will be replaced here, will it still parse the file correctly?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,288 @@
...
+
+	@Test
+	public void testMaxRecordLength() throws Exception {
+
+		// variable-size-objects.json contains 3 json objects but only 2 of them fit in the 27 byte length limitation
+
+		extraParams.add(new Pair<String, String>("IDENTIFIER", "key666"));
+		extraParams.add(new Pair<String, String>("MAXLENGTH", "27"));
+
+		columnDefs.clear();
```
hornn: I think the clear is not necessary since it's also done in the @After function

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,288 @@
...
+		output.add("small object3");
+
+		super.assertOutput(new Path(System.getProperty("user.dir") + File.separator
+				+ "src/test/resources/variable-size-objects.json"), output);
+	}
+
+	@Test
+	public void testDataTypes() throws Exception {
+
+		// TDOO: The BYTEA type is not tested!!! Current oneField.val = val.asText().getBytes(); convention is
```
hornn: Let's open a sub-task for it, what do you think?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,288 @@
...
+
+		output.add("Fri Jun 07 22:45:02 +0000 2013,,text1,SpreadButter,tweetCongress,,");
+		output.add("Fri Jun 07 22:45:02 +0000 2013,,text2,patronusdeadly,,,");
+
+		super.assertOutput(new Path(System.getProperty("user.dir") + File.separator
+				+ "src/test/resources/null-tweets.json"), output);
+	}
+
+	@Test
+	public void testSmallTweetsWithDelete() throws Exception {
```
hornn: what is being tested here?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
```diff
@@ -0,0 +1,288 @@
...
+	}
+
+	@Test
+	public void testWellFormedJsonWithDelete() throws Exception {
+
+		extraParams.add(new Pair<String, String>("IDENTIFIER", "created_at"));
+
+		List<String> output = new ArrayList<String>();
+
+		// output.add(",,,,,,");
```
hornn: remove comment?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
```diff
@@ -0,0 +1,672 @@
...
+
+		List<String> outputLines = new ArrayList<String>();
+
+		String line;
+		while ((line = rdr.readLine()) != null) {
+			outputLines.add(line);
+		}
+
+		assertUnorderedOutput(input, outputLines);
+	}
```
hornn: rdr.close()

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
```diff
@@ -0,0 +1,672 @@
...
+			inputs.add(new LocalInputData(paramsMap));
+		}
+	}
+
+	private JsonNode decodeLineToJsonNode(String line) {
+
+		try {
+			return mapper.readTree(line);
+		} catch (JsonParseException e) {
+			e.printStackTrace();
```
hornn: replace with LOG.debug/warn

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
```diff
@@ -0,0 +1,672 @@
...
+import org.apache.hawq.pxf.api.Fragmenter;
+import org.apache.hawq.pxf.api.OneField;
+import org.apache.hawq.pxf.api.OneRow;
+import org.apache.hawq.pxf.api.ReadAccessor;
+import org.apache.hawq.pxf.api.ReadResolver;
+import org.apache.hawq.pxf.api.WriteAccessor;
+import org.apache.hawq.pxf.api.WriteResolver;
+import org.apache.hawq.pxf.api.io.DataType;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.apache.hawq.pxf.service.FragmentsResponse;
```
hornn: This test file breaks the convention of separating between pxf-service and the plugins. Only pxf-api is supposed to be used by the plugins.
I think it's a really good test mechanism, and we should use it, but it definitely should be moved to another location - perhaps its own package, perhaps under pxf-service.

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class PartitionedJsonParserOffsetTest {
+	/*
+	 * [{"color": "red","v": "vv"},{"color": "red","v": "vv"}]
+	 */
+	public static String json2 = "[{\"color\": \"red\",\"v\": \"vv\"},{\"color\": \"red\",\"v\": \"vv\"}]";
+
+	@Test
+	public void testOffset() throws IOException {
+		InputStream jsonInputStream = createFromString(json2);
```
hornn: close stream

hornn:
@tzolov regarding test coverage, what you did is very impressive.
Normally we have an automation suite where we test end to end query using a specific plugin, but I really like the idea of PxfUnit. 
I think after your commit is done we'll also have to add automation tests, but they will probably not be as exhaustive as what you already did.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/ColumnDescriptorCache.java
tzolov: +1

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/ColumnDescriptorCache.java
tzolov: +1

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: +1

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
tzolov: done

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
tzolov: +1

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParser.java
tzolov: ok, will set it for consistency with the rest preinitialized attributes.

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/parser/PartitionedJsonParserOffsetTest.java
tzolov: +1

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
tzolov: +1. missed this one

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
tzolov: +1

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
tzolov: I am not sure what was the purpose of the deleted json records in the original pxf-json code. Perhaps @adamjshook would remember? 
But i've kept the test because it shows that the identifier will filter out the records not containing the identifier (nothing to do with deletion though).

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
tzolov: Shows that the simple (e.g. LineRecordReader) mode can handle missing or null json attributes. Those are converted into empty fields.

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
tzolov: No the order of the json attributes doesn't affect the parsing. I've modified the test (reversed the columnDefs order) to validate this case as well.

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/JsonExtensionTest.java
tzolov: Sure. let's live it as it is for now. Just as a known issue and address it in a separate (JIRA?) task? 
I don't know what should be the desired value but array ref. doesn't seems right to me. Perhaps @adamjshook  can help.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: Not it points to the element before the last. This is because the last element could either a simple node name or array reference (e.g. name[index]).  Later the JsornResovler#getFields handles both cases separately to resolve the last node element.

tzolov: Update: I've refactored the ColumDescriptorCache to compute a *normalized* projections. In case of array it substitutes the last path element with the name of the array. In this way the JsonResolver#getFields() is simplified. (take a look)

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
tzolov: All the credit for the PxfUnit goes to @adamjshook 
In the pxf-field projects the PxfUnit has its own dedicated project: `pxf-test`. But adding it under the pxf-service might be easiest way to go. Do you want me to move it now or in a follow up task?

tzolov:
Last commit address most issues except the compression thread-safety and the PxfUnit location. Also it assumes that the BYTEA type handling will be evaluated in a separate task. 

@hornn let me know how i can help creating the automation tests. Is there any doc/instructions?

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
hornn: Let's leave it as a follow up.

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -128,4 +128,13 @@ under the License.
             <resolver>org.apache.hawq.pxf.plugins.gemfirexd.GemFireXDResolver</resolver>
         </plugins>
     </profile>
+    <profile>
+        <name>Json</name>
+        <description>A profile for JSON data, one JSON record per line</description>
```
GodenYao: looks like we've consolidate 2 profile - should update the description to not limit it to "one JSON record per line".

pxf/pxf-json/src/test/java/org/apache/hawq/pxf/plugins/json/PxfUnit.java
tzolov: ok

adamjshook: I took some inspiration from MRUnit and wanted to mock the execution framework of PXF.  I imagine it could use some work, but I think it would be very beneficial in the long run for developers to do some testing of their extensions without having to install it on a HAWQ system.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,252 @@
...
+	 * @param line
+	 *            JSON text
+	 * @return Returns a {@link JsonNode} that represents the input line or null for invalid json.
+	 */
+	private JsonNode decodeLineToJsonNode(String line) {
+
+		try {
+			return mapper.readTree(line);
+		} catch (Exception e) {
+			LOG.error("Failed to pars JSON object", e);
```
hornn: typo parse

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
tzolov: Thanks for spotting this @GodenYao! Just fixed it

P.S. as an experiment i've implemented a Json Resolver that leverages [JsonPath](https://github.com/jayway/JsonPath). It allows you to  traverse arbitrary complex json records. (current expression mechanism is quite limited). But this an experimental work and we should leave a possible future improvement. Perhaps as an additional 3rd mode -  the most flexible but potentially slowest.

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
tzolov: +1

cwelton:
What is the current status on this PR?  Are there still outstanding issues to resolve, or is it ready for merge?

GodenYao:
it's pretty much ready - Christian still needs to run automation test (this is still proprietary as we haven't open source automation yet), he's in business trip this week and will pick it up after he returns next week.

hornn:
There are also a few minor fixes to be merged: https://github.com/tzolov/incubator-hawq/pull/1

scottcode:
I have a feature request within the JSON plugin. (Not sure if this is the right venue for this...) Right now, if you specify an element within the JSON that corresponds to further nested objects or arrays, you just get null. I'd like to be able to coerce the nested object as a JSON string. This would be especially useful for structures that have variably-sized nested contents. For example, the Twitter API's `entities` object has an array under the `hashtags` key. It would be nice if specifying `"entities.hashtags" text` within the external table would give you a text representation of the array (e.g. `["tag1", "tag2", "etc"]`). 

If this isn't the right place, let me know where to post such a suggestion. Thanks.

shivzone:
@scottcode the appropriate forum for having questions/discussions such as the one you have is hawq user mailing list. You can find the available mailing lists here http://hawq.incubator.apache.org/#mailing-lists

pxf/pxf-json/src/main/java/org/apache/hawq/pxf/plugins/json/JsonResolver.java
```diff
@@ -0,0 +1,256 @@
...
+			case BIGINT:
+				oneField.val = val.asLong();
+				break;
+			case BOOLEAN:
+				oneField.val = val.asBoolean();
+				break;
+			case CHAR:
+				oneField.val = val.asText().charAt(0);
+				break;
+			case BYTEA:
```
sansanichfb: As far as bytea is not tested and isn't supported, should we remove this section?

sansanichfb:
Merging this PR, will address bytea type in separate story.

GodenYao:
@tzolov , can you close the PR.

changleicn:
@tzolov the code is in, can you please close this PR?

PR #307, opened 2016-02-01
HAWQ-372. Fix single row insert and COPY hang in high concurrent workloads

huor:
Root cause analysis shows that the hang of concurrent workload is in three folds:

1. Most of the queries lock relations at first and then allocate resource for it, while analyze (either manually or automatically triggered) allocate resource at first and then lock relation. This may lead to deadlock between relation lock and query resource with concurrent queries.

2. Some of the queries may do query resource allocation multiple times, i.e., SRI/COPY which triggers automatic statistics collection. They do not return query resource as soon as some of the sub tasks are done for the query.
For example, SRI allocate resource for insert itself, do insertion, return query resource for insertion, allocate resource for automatically triggered analyze, do analyze, return query resource for analyze; while COPY allocate resource for COPY itself, do COPY, allocate resource for automatically triggered analyze, do analyze, return query resource for analyze, return query resource for COPY. This may lead to a lot of query resource for COPY itself is occupied, and they still try to allocate more resource for analyze. Thus, it makes some of the COPY pending to allocate resource for analyze, which seems like "deadlock" on query resource.

3. SRI/COPY do query resource allocation multiple times, while TPC-H do resource allocation only once. Usually TPC-H queries take longer time to complete. Thus, SRI/COPY maybe run in halfway and do second resource allocation for some of the sub-tasks. If meanwhile all resource are busy, SRI/COPY need to wait for TPC-H queries to return resource and then proceed. As a consequence, SRI/COPY run very slow or even hang in user's standpoint.

For address the issue, we do following fix:

For 1, make sure all queries (especially insert queries, create/alter/drop database object queries) follows the pattern that lock relation at first, and then allocate query resource.

For 2, make sure queries follows the pattern that allocate query resource for sub-task1, return query resource for sub-task1, ..., allocate query resource for sub-taskN, return query resource for sub-taskN

For 3, from user practice, separate different workloads in different resource queues, i.e., SRI/COPY in one load queue, while TPC-H in query queue.

### DIFF
--- a/src/backend/commands/analyze.c
+++ b/src/backend/commands/analyze.c
@@ -269,7 +269,7 @@ void analyzeStatement(VacuumStmt *stmt, List *relids, int preferred_seg_num)
 
 	PG_TRY();
 	{
-		analyzeStmt(stmt, relids);
+		analyzeStmt(stmt, relids, target_seg_num);
 		gp_autostats_mode = autostatvalBackup;
 		gp_autostats_mode_in_functions = autostatInFunctionsvalBackup;
 		optimizer = optimizerBackup;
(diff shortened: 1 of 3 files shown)


ztao1987:
+1

jiny2:

huor:
Close it since the fix is refactored and a new PR #321 is created.

PR #309, opened 2016-02-01
HAWQ-373: Fixed QE coredump in pg_encoding_mbcliplen when run concrur…

liming01:
…rent copy in stress

The root cause is:
In the function copy_in_error_callback() we should consider the process data may be incomplete because master maybe quit at any time when error issues.

### DIFF
--- a/src/backend/commands/copy.c
+++ b/src/backend/commands/copy.c
@@ -6243,9 +6243,12 @@ copy_in_error_callback(void *arg)
 			line_start = copy_of_line_buf.data;
 			lineno_delim = memchr(line_start, COPY_METADATA_DELIM, Min(32, cstate->line_buf.len));
 
-			value_len = lineno_delim - line_start + 1;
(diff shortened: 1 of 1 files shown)


src/backend/commands/copy.c
```diff
@@ -6243,9 +6243,12 @@ copy_in_error_callback(void *arg)
 			line_start = copy_of_line_buf.data;
 			lineno_delim = memchr(line_start, COPY_METADATA_DELIM, Min(32, cstate->line_buf.len));
 
-			value_len = lineno_delim - line_start + 1;
-			copy_of_line_buf.data += value_len; /* advance beyond line_buf_converted */
-			copy_of_line_buf.len -= value_len;		
+            if (lineno_delim)
```
huor: The indentation need to be followed.

huor:

wangzw:
The fix looks good. +1

PR #310, opened 2016-02-01
HAWQ-374. Fixed hawq core dump at dispatch_free_result() when run str…

liming01:
…ess test

The root cause is:
The copy error may occurs after  c->executors initiated at cdbCopyEnd(), but before 
c->executors.segment_conns initiated line 1432 at dispatch_statement() .


huor:
The fix looks good. +1

wangzw:
The fix looks good. +1

PR #312, opened 2016-02-01
HAWQ-376. Fixed QE core dump at HdfsCloseFile() when run stress test …

liming01:
…with COPY

The root cause of this defect is:
When the master got some error, it will send signal SIGABRT to all segments and then rollback all works done by all segments. If the segment node haven't finished processing SIGABRT before the master finishing rollback, all segfile created at this segment is deleted at master node in fact. So it will report warning to close/delete that hdfs file. We should clean up hdfs file cache before calling HdfsCloseFile(), so that we will skip second call of HdfsCloseFile() at ShutdownPostgres().

### DIFF
--- a/src/backend/storage/file/fd.c
+++ b/src/backend/storage/file/fd.c
@@ -632,32 +632,54 @@ LruDelete(File file)
 	if (vfdP->seekPos == INT64CONST(-1))
 		elog(ERROR, "could not get the current position of file \"%s\": %m", vfdP->fileName);
 
-	/* close the file */
-	if (IsLocalPath(VfdCache[file].fileName))
-	{
-		if (close(vfdP->fd))
-			elog(ERROR, "could not close file \"%s\": %m", vfdP->fileName);
-	}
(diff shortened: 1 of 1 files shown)


wangzw:
looks good to me. +1

src/backend/storage/file/fd.c
```diff
@@ -632,33 +632,54 @@ LruDelete(File file)
 	if (vfdP->seekPos == INT64CONST(-1))
 		elog(ERROR, "could not get the current position of file \"%s\": %m", vfdP->fileName);
 
-	/* close the file */
-	if (IsLocalPath(VfdCache[file].fileName))
-	{
-		if (close(vfdP->fd))
-			elog(ERROR, "could not close file \"%s\": %m", vfdP->fileName);
-	}
+	// Need to firstly remove file description from VfdCache before closing file,
```
huor: For multiple line comment, use
/**
 * xxx
 */

huor:
LGTM. +1

PR #313, opened 2016-02-01
HAWQ-375. Fixed Fixed hawq core dump at dispatch_free_result():1493 w…

liming01:
…hen run stress test

Same as HAWQ-374, it crashed at different variable, we need to make sure initiate it with NULL, and free it when it is not NULL.


wangzw:
looks good to me. +1

huor:
LGTM. +1

PR #317, opened 2016-02-01
HAWQ-380. Wrongly validate resource pool status when index of allocated resources is not complete

jiny2:
When RM finds appropriate segments to remove GRM containers, the index node of allocated resource maybe partially removed, therefore, we can only validate the resource pool after restoring the index.


huor:
LGTM. +1

PR #318, opened 2016-02-02
HAWQ-382. Adjust combined workload index in resource pool in different situations

jiny2:
This fix it to make combined workload index maintained well when consuming resource, mark segment up/down, etc.


linwen:
Looks good.

huor:
LGTM. +1

PR #319, opened 2016-02-02
HAWQ-383. Fix memory leak when testing local postmaster health

jiny2:
The connection object is not freed if the connecting attempt is failed.


yaoj2:
+1

PR #322, opened 2016-02-02
HAWQ-386. Remove deprecated catalog table gp_fault_strategy

huor:


linwen:
Looks good.

tools/bin/hawqpylib/system/configurationImplHAWQ.py
```diff
@@ -276,12 +271,8 @@ def __updateSystemConfigUpdateSegment(self, conn, hawqArray, seg, originalSeg, t
         # update mode and status
         # when adding a mirror, the replication port may change as well
         #
-        if hawqArray.getFaultStrategy() == hawqarray.FAULT_STRATEGY_NONE:
-            what = "%s: segment hostname and address"
-            self.__updateSegmentAddress(conn, seg)
-        else:
-            what = "%s: segment mode and status"
-            self.__updateSegmentModeStatus(conn, seg)
```
liming01: From the hawq1.x doc, it seems 3 types fault_strategy can be set: nothing / file-based failover / SAN-based failover.
Now we remove this catalog, it means we only have one choice, is it the 'nothing' setting left? 
so here in the code, you may keep the wrong branch in the if-else statement.

PR #323, opened 2016-02-02
[HAWQ-19] Money type overflow fixed by back porting int64 support from Postgres commit 74a4019

0x0FFF:
This commit fixes the issue [HAWQ-19](https://issues.apache.org/jira/browse/HAWQ-19) "Money type overflow" by back porting commit [74a4019](https://github.com/postgres/postgres/commit/74a40190aabd27b052ea7a63e93f7ed47743755d#diff-7ddd78dd264f8cf2a62cf572baf90b84) to HAWQ. Changes internal representation of "money" datatype, input function, cash_words function, adds function to multiply/divide cash by int8. Also changes pg_type to accommodate new internal representation

Before this commit:
```
test=# select '$30000000'::money;
      money      
-----------------
 -$12,949,672.96
(1 row)
```
After this commit:
```
test=# select '$30000000'::money;
     money      
----------------
 $30,000,000.00
(1 row)
```

### DIFF
--- a/src/backend/cdb/cdbparquetcolumn.c
+++ b/src/backend/cdb/cdbparquetcolumn.c
@@ -486,7 +486,7 @@ decodePlain(Datum *value, uint8_t **buffer, int hawqTypeID)
 			 * should be a pointer, see cash.c
 			 */
 			(*value) = PointerGetDatum(*buffer);
-			(*buffer) += 4;
+			(*buffer) += 8;
 			break;
 		}
 
--- a/src/backend/cdb/cdbparquetstoragewrite.c
+++ b/src/backend/cdb/cdbparquetstoragewrite.c
@@ -653,7 +653,6 @@ mappingHAWQType(int hawqTypeID)
 
 	case HAWQ_TYPE_INT2:
 	case HAWQ_TYPE_INT4:
-	case HAWQ_TYPE_MONEY:
 	case HAWQ_TYPE_DATE:
 		return INT32;
 
@@ -661,6 +660,7 @@ mappingHAWQType(int hawqTypeID)
 	case HAWQ_TYPE_TIME:
 	case HAWQ_TYPE_TIMESTAMPTZ:
 	case HAWQ_TYPE_TIMESTAMP:
+	case HAWQ_TYPE_MONEY:
 		return INT64;
 
 	case HAWQ_TYPE_FLOAT4:
@@ -1274,13 +1274,18 @@ encodePlain(Datum data,
 		memcpy(dst_ptr, &val, len);
 		return len;
 	}
+
+
+	/*----------------------------------------------------------------
+	 * Type mapped to 8-bytes INT64/DOUBLE in Parquet
+	 *----------------------------------------------------------------*/
 	case HAWQ_TYPE_MONEY:
 	{
 		/*
-		 * Although money is represented as int32 internally,
+		 * Although money is represented as int64 internally,
 		 * it's passed by reference.
 		 */
-		len = 4;
+		len = 8;
 		if (!ensureBufferCapacity(current_page, len, pageSizeLimit))
 		{
 			return ENCODE_OUTOF_PAGE;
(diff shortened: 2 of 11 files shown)


cwelton:
This change as on disk format implications so we must also consider the impact on upgrade.

GodenYao:
> This change as on disk format implications so we must also consider the impact on upgrade.

Open source community generally have higher tolerance of format changes across versions I suppose?
it'd be vendors who are based on the code base may have to consider upgrade scenarios and format compatibility.

cwelton:
Good topic for the community to discuss.  As an example the Postgres community used to take a stance of tolerance around binary version changes, but has since adopted a stronger stance around binary compatibility.  Its ultimately a question of innovation vs maintainability and strongly effects how quickly people will move from one version of the product to the next.

0x0FFF:
Speaking about this specific case: do we plan an upgrade path from HAWQ 1.x to 2.0? If no, then this PR can be easily merged into 2.0. If yes, we should really care about the upgrade path, and maybe the better solution here would be the following: instead of changing existing data type introduce new one, like "money8" for 8-byte money data type. This would again be a catalog change (for example, in pg_type), but still won't affect the customers using old data type "money" and layout of their existing tables

Generally speaking, I agree that this is a tough question. Postgres is not a very good example here - you can easily do pg_dump and pg_restore there as the data volumes it operates with are usually quite small. As of HAWQ, making dump and restore to accommodate storage format change can't be an option as we might have hundreds of TB there.

changleicn:
First of all, good fix. In 2.0GA, upgrade path from hawq 1.x to 2.0 is via data reloading. Have we tested parquet format, and mapreduce inputformat? The change might have impact on these aspects?

0x0FFF:
Thanks for mentioning, fixed and tested Parquet compatibility

0x0FFF:
As of HAWQ inputformat, it never [supported "money" datatype](https://github.com/apache/incubator-hawq/blob/master/contrib/hawq-hadoop/hawq-mapreduce-common/src/main/java/com/pivotal/hawq/mapreduce/schema/HAWQPrimitiveField.java#L33):

```java
/**
 * Enumeration of all supported primitive type.
 * Note that CHAR and BPCHAR are the same, we include BPCHAR for internal usage.
 */
public static enum PrimitiveType {
	BOOL, BIT, VARBIT, BYTEA, INT2, INT4, INT8, FLOAT4, FLOAT8, NUMERIC,
	CHAR, BPCHAR, VARCHAR, TEXT, DATE, TIME, TIMETZ, TIMESTAMP, TIMESTAMPTZ, INTERVAL,
	POINT, LSEG, BOX, CIRCLE, PATH, POLYGON, MACADDR, INET, CIDR, XML
}
```
We can add it, but I think it should be a separate PR
Also it would be nice to port other "cash" functions from postgres, allowing convertation of numeric, int and float to money

changleicn:
A PR to support "money" type and porting other "cash" functions are great.

And another suggestion is to add a installcheck-good test to cover this fix.

0x0FFF:
Added installcheck to cover "money" data type (both to good, serial and parallel checks)

changleicn:
nice. LGTM.

liming01:
Cool! 

BTW, There are 2 points here:
(1) It seems we need to add meta info for functions you added(e.g. int8_mul_cash() ) into below files:
apache-hawq/src/backend/utils/fmgrtab.c
apache-hawq/src/include/catalog/pg_operator.h
apache-hawq/src/include/catalog/pg_proc.h 

(2) I saw you changed argument type to int64 for function cash_div_int4(), but didn't change for fucntion 
cash_div_int2(). I guess you may want to fix the problem caused by (1) point above. So please revert this change. It may cause the result precision.

liming01:
Hi Alexey,

If you need my help to move on, please let me know. Thanks.

liming01:
I have fixed the issue I pointed out above, and merge this change into master branch. So I close this pull request.

0x0FFF:
Sorry, I was in a trip so didn't have time to work on this. Thank you for fixing it

changleicn:
@0x0FFF the code is in, can you close this pull request? Thanks!

PR #327, opened 2016-02-03
HAWQ-390. Fix memory leak in dispatcher.

ztao1987:


huor:
It is better to set the pointers to NULLs after pfree.
LGTM. +1

ztao1987:
Sure, set it to NULL in the code

PR #330, opened 2016-02-03
HAWQ-252. fix a coredump when yarn heartbeat thread doesn't exit quickly

linwen:
Please review.


jiny2:

liming01:
Looks good to me. +1.

depends/libyarn/src/libyarnclient/LibYarnClient.cpp
```diff
@@ -106,7 +106,7 @@ list<ResourceRequest> LibYarnClient::getAskRequests() {
 
 void* heartbeatFunc(void* args) {
```
huor: 1. Do we need to retry here as it already retries in "client->dummyAllocate();"?

2. If we need to retry, the code can be refined as below. The main points are:

1) use macro instead of magic number for maximum retry number
2) provide retry time in logging information
3) use /* ... */ for multiple line comment

{noformat}
	#define HEARTBEAT_MAX_RETRY_NUM 2

	int failcounter = 0;
	LibYarnClient *client = (LibYarnClient*)args;

	while (client->keepRun) {
		try {
			client->dummyAllocate();
			failcounter = 0;
		}
		catch(const YarnException &e) {
			failcounter++;

			LOG(WARNING, "LibYarnClient::heartbeatFunc, dummy allocation "
						 "is not correctly executed in try %d with exception raised. %s",
						 failcounter;
						 e.msg());

			if ( failcounter >= HEARTBEAT_MAX_RETRY_NUM ) {
				/**
				 * In case retry too many times with errors/exceptions, this
				 * thread will return. LibYarn has to re-register application
				 * and start the heartbeat thread again.
				 */
				LOG(WARNING, "LibYarnClient::heartbeatFunc, there are too many "
						     "failures raised after %d maximum retries. This heart-beat thread exits now.", HEARTBEAT_MAX_RETRY_NUM);
				client->keepRun = false;
				break;
			}
		}
		usleep((client->heartbeatInterval) * 1000);
	}
{noformat}

linwen: Thanks, I will refactor the whole file in future.

PR #333, opened 2016-02-04
HAWQ-396. Fixed crashed at nested error report calling when close tem…

liming01:
…p file at AbortTransaction()


wangzw:
looks good to me. +1

huor:
LGTM. +1

PR #335, opened 2016-02-05
HAWQ-402. Fix bug that core dump due to nullptr in GroupClause

ztao1987:


src/backend/optimizer/util/tlist.c
```diff
@@ -317,6 +317,10 @@ get_sortgrouplist_exprs(List *sortClauses, List *targetList)
 		SortClause *sortcl = (SortClause *) lfirst(l);
 		Node	   *sortexpr;
 
+		// if GroupClause in grouping sets is null,
```
huor: Use /* ... */ for line or multiple line comment instead of //

huor:

src/backend/optimizer/util/tlist.c
ztao1987: Fixed.

PR #337, opened 2016-02-09
HAWQ-404. Add sort during INSERT of append only row oriented partition tables

hsyuan:
In this PR, 1 GUC was added. 

`optimizer_parts_to_force_sort_on_insert ` implies minimum number of partitions required for sorting tuples during insertion in an append only row-oriented partitioned table.

For example, 
If `optimizer_parts_to_force_sort_on_insert ` was set to 160, then

GIVEN a HAWQ user
AND I have 161 partitioned table
WHEN I INSERT INTO tableA (SELECT * FROM tableB);
THEN the tables should first be sorted so that HDFS does not get overwhelmed
AND then the tables should be opened one by one for the insertion

GIVEN a HAWQ user
AND I have 159 partitioned table
WHEN I INSERT INTO tableA (SELECT * FROM tableB);
THEN the tables do not need to be sorted
AND insertion can happen all at once with each table open

### DIFF
--- a/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
+++ b/src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
@@ -597,6 +597,7 @@ CTranslatorRelcacheToDXL::Pmdrel
 	DrgPmdid *pdrgpmdidIndexes = NULL;
 	DrgPmdid *pdrgpmdidTriggers = NULL;
 	DrgPul *pdrgpulPartKeys = NULL;
+	ULONG ulLeafPartitions = 0;
 	BOOL fConvertHashToRandom = false;
 	DrgPdrgPul *pdrgpdrgpulKeys = NULL;
 	DrgPmdid *pdrgpmdidCheckConstraints = NULL;
@@ -652,6 +653,13 @@ CTranslatorRelcacheToDXL::Pmdrel
 				erelstorage = IMDRelation::ErelstorageAppendOnlyParquet;
 			}
 		}
+
+		// get number of leaf partitions
+		if (gpdb::FRelPartIsRoot(oid))
+		{
+		   ulLeafPartitions = gpdb::UlLeafPartitions(oid);
+		}
+
 		// get key sets
 		BOOL fAddDefaultKeys = FHasSystemColumns(rel->rd_rel->relkind);
 		pdrgpdrgpulKeys = PdrgpdrgpulKeys(pmp, oid, fAddDefaultKeys, fPartitioned, pulAttnoMapping);
@@ -722,6 +730,7 @@ CTranslatorRelcacheToDXL::Pmdrel
 							pdrgpmdcol,
 							pdrpulDistrCols,
 							pdrgpulPartKeys,
+							ulLeafPartitions,
 							fConvertHashToRandom,
 							pdrgpdrgpulKeys,
 							pdrgpmdidIndexes,
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -762,6 +762,7 @@ bool 		optimizer_multilevel_partitioning;
 bool        optimizer_enable_derive_stats_all_groups;
 bool		optimizer_explain_show_status;
 bool		optimizer_prefer_scalar_dqa_multistage_agg;
+int			optimizer_parts_to_force_sort_on_insert;
 
 /* Security */
 bool		gp_reject_internal_tcp_conn = true;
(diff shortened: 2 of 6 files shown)


hsyuan:
@vraghavan78 @xinzweb 
please review this PR.

src/backend/utils/misc/guc.c
```diff
@@ -6155,6 +6156,16 @@ static struct config_int ConfigureNamesInt[] =
...
+		{"optimizer_insert_sort_partition_number", PGC_USERSET, DEVELOPER_OPTIONS,
+			gettext_noop("Minimum number of partitions required for sorting tuples during insertion in an append only row-oriented partitioned table"),
+			NULL,
+			GUC_NOT_IN_SAMPLE
+		},
+		&optimizer_insert_sort_partition_number,
+		INT_MAX, 0, INT_MAX, NULL, NULL
+	},
+
+	{
```
hsyuan: Can you guys come up with a better GUC name for this?
@vraghavan78 @xinzweb @oarap

vraghavan78: optimizer_parts_to_force_sort_on_insert;

src/backend/utils/cache/lsyscache.c
```diff
@@ -1799,6 +1799,25 @@ get_rel_name_partition(Oid relid)
...
+ * get_rel_number_partitions
+ *
+ *		Returns the number of leaf partitions for a given root relation.
+ *		Returns 0 if the relation is not a partition table or non-root partition table.
+ */
+int32
+get_rel_number_partitions(Oid relid)
+{
+	if (PART_STATUS_NONE == rel_part_status(relid))
+		return 0;
```
vraghavan78: comment this please.

src/backend/utils/cache/lsyscache.c
```diff
@@ -1799,6 +1799,25 @@ get_rel_name_partition(Oid relid)
...
+
+	if (!rel_is_partitioned(relid))
+		return 0;
+
+	int count = countLeafPartTables(relid);
```
vraghavan78: how are intermediate partitions handled?

src/include/gpopt/gpdbwrappers.h
```diff
@@ -303,6 +303,9 @@ namespace gpdb {
 	// find the oid of the root partition given partition oid belongs to
 	Oid OidRootPartition(Oid oid);
 	
+	// find the number of partitions for a given root relation
+	int INumberPartitions(Oid oid);
+
```
vraghavan78: how is it different to 
{code}
gpos::ULONG UlLeafPartitions(Oid oidRelation);
{code}

Can we get rid one of them? Also can we make the return of the gpos::ULONG?

hsyuan: Looked at this function again, `UlLeafPartitions` returns the number of leaf partitions, it only accept oid of root partitioned table. 
Both function uses `countLeafPartTables` to count the number of leaf partitions. 
The only difference is that `INumberPartitions ` can accept non root partition oid, just return 0 when it is neither a root partition table nor a partition table.
In fact, we can get rid of INumberPartitions.

src/backend/utils/cache/lsyscache.c
hsyuan: for interior partitions, rel_is_partitioned(relid) will return false. 0 will be returned.

src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
```diff
@@ -652,6 +653,10 @@ CTranslatorRelcacheToDXL::Pmdrel
 				erelstorage = IMDRelation::ErelstorageAppendOnlyParquet;
 			}
 		}
+
+		// get number of leaf partitions
+		ulLeafPartitions = gpdb::INumberPartitions(oid);
+
```
hsyuan:
```cpp
ulLeafPartitions = 0;
if (gpdb::FRelPartIsRoot(oid))
{
   ulLeafPartitions = gpdb::UlLeafPartitions(oid);
}
```

I recommend we change to this code snippet, thus we can get rid of INumberPartitions.

src/backend/utils/misc/guc.c
```diff
@@ -6155,6 +6156,16 @@ static struct config_int ConfigureNamesInt[] =
 	},
 
 	{
+		{"optimizer_parts_to_force_sort_on_insert", PGC_USERSET, DEVELOPER_OPTIONS,
+			gettext_noop("Minimum number of partitions required for sorting tuples during insertion in an append only row-oriented partitioned table"),
```
vraghavan78: "require to force sorting tuples"

src/backend/gpopt/translate/CTranslatorRelcacheToDXL.cpp
```diff
@@ -652,6 +653,13 @@ CTranslatorRelcacheToDXL::Pmdrel
 				erelstorage = IMDRelation::ErelstorageAppendOnlyParquet;
 			}
 		}
+
+		// get number of leaf partitions
+		if (gpdb::FRelPartIsRoot(oid))
+		{
+		   ulLeafPartitions = gpdb::UlLeafPartitions(oid);
+		}
```
vraghavan78: cleaner!

vraghavan78:
LGTM!

hsyuan:
Merged to master: https://github.com/apache/incubator-hawq/commit/321f7e2b737baf6fe674bd0ebc0ea05f9b166897

PR #338, opened 2016-02-10
HAWQ-405. Fix for handling of pg_temp schema

shivzone:


src/backend/catalog/namespace.c
```diff
@@ -1495,18 +1495,34 @@ LookupInternalNamespaceId(const char *nspname)
...
             return caql_getoid(
 			NULL,
 			cql("SELECT oid FROM pg_namespace "
 				" WHERE nspname = :1",
 				CStringGetDatum((char *) nspname)));
-        }else{
-            elog(ERROR, "Upgrade cann't process the namespace: %s in dbid: %i", nspname, dboid);
         }
+        else
+        	elog(ERROR, "Upgrade cann't process the namespace: %s in dbid: %i", nspname, dboid);
```
hornn: typo cannot

src/backend/catalog/namespace.c
```diff
@@ -1495,18 +1495,34 @@ LookupInternalNamespaceId(const char *nspname)
...
+			return myTempNamespace;
+		/*
+		 * Since this is used only for looking up existing objects, there
+		 * is no point in trying to initialize the temp namespace here;
+		 * and doing so might create problems for some callers.
+		 * Just fall through and give the "does not exist" error.
+		 */
+	}
+
+    if(gp_upgrade_mode)
```
hornn: indentation looks off

GodenYao: I am thinking we may be able to remove the entire if(gp_upgrade_mode), as there's no upgrade path from previous Pivotal commercial 1.3.x HAWQ to 2.0 (only way is to do data reloading).
So this logic probably is no longer needed.

hornn: It should not be part of this commit though.

hornn:
I think it would be good to add the two cases that failed to installcheck-good. I checked, and there is already a temp test suite (src/test/regress/expected/temp.out, src/test/regress/sql/temp.sql) that can be a good place for them.

sansanichfb:
+1

shivzone:
Added to installcheck-good

src/test/regress/expected/temp.out
```diff
@@ -170,3 +170,19 @@ select * from whereami;
 -- you can invoke a temp function explicitly, though
 -- select pg_temp.whoami();
 drop table public.whereami;
+create temp table temptest (row integer, count integer);
```
hornn: I think it would be better to have the exact scenario that failed:
```
CREATE TABLE pg_temp.temptest ...
SELECT avg(pg_temp.temptest.count) ... 
```
Do you think we should cover all three cases (db.table.col, table.col, col)?

hornn:
+1 :)

PR #339, opened 2016-02-12
No need to use ksh and add an extra runtime dep

c0s:
Why a single file in the whole project will be using ksh without any ksh specific functionality it might need?

### DIFF
--- a/tools/bin/diskqual.ksh
+++ b/tools/bin/diskqual.ksh
@@ -1,4 +1,4 @@
-#!/bin/ksh
+#!/bin/bash
 #
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file


rvs:
LGTM. Unless somebody complains I'll be merging this today.

hsyuan:
@rvs  Is this merged into master?

c0s:
No, 2 months later it is still sitting in the queue.

radarwave:
I tested the changes on RHEL6.5, seems many command in the scripts not available. 

And as my check, no others using this script. And this feature of this script is not important. It just do a simple test of disk speed. 

So I suggest we remove this file.

[gpadmin@test1 bin]$ ./diskqual.ksh
./diskqual.ksh: line 2: format: command not found
./diskqual.ksh: line 2: nawk: command not found

edespino:
@c0s - can you please update this PR to essentially remove the file altogether? It is not needed in the HAWQ context.

PR #341, opened 2016-02-15
HAWQ-407. memory leak when insert into parquet table

wangzw:


wangzw:
@ictmalili Would you please review this PR. Thanks

PR #343, opened 2016-02-15
HAWQ-49. Remove madlib related codes and tests.

liming01:
Madlib related code will be moved to madlib project, we need to remove all related codes and tests.


linwen:
looks good.

huor:
LGTM. +1

PR #345, opened 2016-02-16
HAWQ-412. Allocate query resource for multiple EXECUTIONs in explicit and implicit prepared statement

huor:
The resolution is to:
1. during planning phase, store the parameters that were used to allocate the query resource along with query resource itself in planned statement
2. in multiple EXECUTIONs, use the stored parameters to allocate query resource in the runtime

The query resource parameter includes:
1. Query resource life (i.e., QRL_ONCE, QRL_INHERIT)
2. Slice number
3. Query cost (i.e., iobytes)
4. Maximum target segment number
5. Minimum target segment number
6. Hostname volume information
7. Hostname volume information size

### DIFF
--- a/src/backend/cdb/cdbdatalocality.c
+++ b/src/backend/cdb/cdbdatalocality.c
@@ -4004,12 +4004,20 @@ calculate_planner_segment_num(Query *query, QueryResourceLife resourceLife,
 	/*
 	 * Initialize QueryResourceParameters in QD
 	 *
-	 * We use TopMemoryContext here so that the QueryResourceParameter
-	 * can be available in the session. Thus, it can be used in multiple
-	 * "EXECUTION"s of the prepared statement (i.e., "PREPARE", "BIND",
-	 * "EXECUTION"
+	 * We use CacheMemoryContext/TopMemoryContext here so that the
+	 * QueryResourceParameter can be available in the session. Thus,
+	 * it can be used in multiple "EXECUTION"s of the prepared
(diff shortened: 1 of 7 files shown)


zhangh43:
+1

src/backend/cdb/cdbdatalocality.c
```diff
@@ -3967,6 +4001,18 @@ calculate_planner_segment_num(Query *query, QueryResourceLife resourceLife,
...
 
+	/*
+	 * Initialize QueryResourceParameters in QD
+	 *
+	 * We use TopMemoryContext here so that the QueryResourceParameter
+	 * can be available in the session. Thus, it can be used in multiple
+	 * "EXECUTION"s of the prepared statement (i.e., "PREPARE", "BIND",
+	 * "EXECUTION"
+	 */
+	MemoryContext oldcontext = MemoryContextSwitchTo(TopMemoryContext);
```
wangzw: TopMemoryContext has only 8kB. You will get trouble to store information in it.

wangzw:
+1 if memory context issue is fixed.

huor:
Good idea. For a quick fix, we can use CacheMemoryContext which is 8M at maximum.

We are investigating on plan context for different queries including:
1) SPI query
2) Named prepared statement
3) Unnamed prepared statement
After that, we can use more appropriate memory context for prepared plan accordingly. This will be a separate improvement.

PR #346, opened 2016-02-16
HAWQ-413. Create_table_distribution sporadically fails.

zhangh43:
"create_table_distribution test in installcheck-good fails sporadically.
I was running it on my Mac with one segment - sometimes it fails and sometimes not."

The way I fix it is to change default_segment_num to 8. The recommend value of default_segment_num is 8* host_number


wengyanqing:
LGTM

PR #348, opened 2016-02-17
HAWQ-407. Fix memory leak in parquet metadata processing for reading table with data

ictmalili:
Move the parquetMeta struct construction from two callers to one place, and avoid memory leak.


ztao1987:
Looks good to me. +1

PR #349, opened 2016-02-17
HAWQ-414. Share memory become inconsistent if the number of segment f…

wangzw:
…iles created more than max_appendonly_segfiles


ztao1987:
+1

wangzw:
merged

PR #351, opened 2016-02-17
HAWQ-416. core dump when cancel delegation token if fatal error happe…

wangzw:
…ned when setup connection to NAMENODE


ztao1987:
+1

wangzw:
merged

PR #353, opened 2016-02-18
HAWQ-420. Memory leaks in DataLocality during ANALYZE lasting 1 ~ 2 d…

zhangh43:
…ays on 100 nodes cluster.

Just add TRY CATCH in calculate_planner_segment_num()


wengyanqing:
LGTM

PR #356, opened 2016-02-18
HAWQ-423. Updated gradle and download plugin version

shivzone:


GodenYao:
+1

pxf/build.gradle
```diff
@@ -434,7 +435,6 @@ task tomcatRpm(type: Rpm) {
         packageName 'apache-tomcat'
         summary = 'Apache Tomcat RPM'
         vendor = project.vendor
-        release = buildNumber()
```
sansanichfb: Why do we need to remove build number from rpm name?

shivzone: The HAWQ pulse build number is not relevant for the apache tomcat rpm.
Besides we are not releasing internal custom builds of apache tomcat

sansanichfb:
+1

PR #357, opened 2016-02-19
HAWQ-275. Add connection check when dispatcher wants to reuse the QE forked by former queries, and when dispather dispatches the detailed query command to QE

ictmalili:
Add connection check when dispatcher wants to reuse the QE forked by former queries, and when dispather dispatches the detailed query command to QE

### DIFF
--- a/src/backend/cdb/dispatcher.c
+++ b/src/backend/cdb/dispatcher.c
@@ -1803,3 +1803,40 @@ dispatcher_print_statistics(StringInfo buf, DispatchData *data)
 			INSTR_TIME_GET_MILLISEC(data->time_total_free) / data->num_of_dispatched);
 }
 
+
+
+/*
+ * Check the connection from the dispatcher to verify that it is still there.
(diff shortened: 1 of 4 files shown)


zhangh43:
+1 and it's better to reuse dispatcherAYT, because executormgr_validate_conn  and dispatcherAYT have a lot of repeated code.

zhangh43:
One more thing, is there a need to consider errno == EWOULDBLOCK in your executormgr_validate_conn()  if connection is nonblocking

ictmalili:
Yes, function executormgr_validate_conn and dispatcherAYT have repeated code, but I'm not sure whether we can merge the two functions into one since dispatcherAYT is in ic_udp.c and interconnect seems a independent part. Shall we merge that? @changleicn

ictmalili:
Yes, I'll add that errno check errno == EWOULDBLOCK in the code. And seems that EAGAIN and EWOULDBLOCK often return same values.

zhangh43:
+1

PR #361, opened 2016-02-20
HAWQ-426. Fix memory leak for query and prepared statement due to introduction of query resource parameters

huor:
The query resource parameters is kept in planned statement. So, the resolution is to make query resource parameters using the same memory context as planned statement. Thus, they can be freed at the same time.

From implementation standpoint, we do a deep copy of query resource parameters in copy function of plannedstmt so that they use the same memory context.


wengyanqing:
+1

ztao1987:
+1

PR #363, opened 2016-02-21
HAWQ-425. Add default value for hawq_rm_memory_limit_perseg and hawq_…

linwen:
Add default value for hawq_rm_memory_limit_perseg and hawq_rm_nvcore_limit_perseg, so that, segment can start without these two properties configured in yarn mode. If in NONE mode, these two properties should be configured in hawq-site.xml to overwrite the default value.


jiny2:
+1 LGTM.

huor:
LGTM. +1

PR #364, opened 2016-02-22
HAWQ-428. ORDER/GROUP BY not found in target list error found in olap…

wengyanqing:
ignore error if fail in get_sortgroupclause_tle function  when check targetentry in rollup. Because rollup clause only need groupkeys.

### DIFF
--- a/src/backend/optimizer/plan/plangroupext.c
+++ b/src/backend/optimizer/plan/plangroupext.c
@@ -1867,7 +1867,12 @@ convert_gs_to_rollups(AttrNumber *grpColIdx,
 		 * resno.
 		 */
 		sc.tleSortGroupRef = no+1;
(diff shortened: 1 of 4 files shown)


src/backend/optimizer/plan/plangroupext.c
```diff
@@ -1867,7 +1867,12 @@ convert_gs_to_rollups(AttrNumber *grpColIdx,
 		 * resno.
 		 */
 		sc.tleSortGroupRef = no+1;
-		tle = get_sortgroupclause_tle(&sc, tlist);
+		tle = get_sortgroupclause_tle_internal(&sc, tlist);
```
zhangh43: neet to format here

zhangh43:
+1

PR #365, opened 2016-02-22
HAWQ-428. ORDER/GROUP BY not found in target list error found in olap…

wengyanqing:
ignore error check in get_sortgroupclause_tle


ztao1987:
+1

PR #367, opened 2016-02-22
HAWQ-430. HAWQ resource manager prunes those registered segments having different memory to core ratio

jiny2:
1) When there are different ratio machines registered, HAWQ resource manager will choose a ratio having least memory wast as cluster level memory to core ratio;
2) All registered segments having different ratios will be pruned to ensure each segment in HAWQ cluster has the same resource memory to core ratio.


linwen:
looks good.

yaoj2:
+1

PR #370, opened 2016-02-23
Hawq438 EagerlyReleased hash table involved in hash join in explain statement introduce core on hawq dbg build.

zhangh43:


ictmalili:
It looks good to me. +1

PR #371, opened 2016-02-23
HAWQ-440. Fix bug in pg_regress of commit HAWQ-432.

zhangh43:
initfile  and expected_statuses_file are constant string and should not be freed


ztao1987:
+1

PR #375, opened 2016-02-23
Karthifoyzur hawq 368

karthijrk:
Fixing the runaway query termination. HAWQ issue link - https://issues.apache.org/jira/browse/HAWQ-368

@gcaragea @armenatzoglou @foyzur please take a look when you get chance

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -4652,7 +4652,22 @@ PostgresMain(int argc, char *argv[], const char *username)
 		}
 
 		IdleTracker_DeactivateProcess();
-		firstchar = ReadCommand(&input_message);
+		/*
+		 * During read command, if we have any exception we want to reactivate the process
+		 * before falling back to postgres main exception handler.
+		 */
+		PG_TRY();
+		{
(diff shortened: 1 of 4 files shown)


src/backend/tcop/postgres.c
```diff
@@ -4652,7 +4652,22 @@ PostgresMain(int argc, char *argv[], const char *username)
...
+		{
+			firstchar = ReadCommand(&input_message);
+		}
+		PG_CATCH();
+		{
+			IdleTracker_ActivateProcess();
+			elog(LOG, "Caught exception while reading command");
+			PG_RE_THROW();
+		}
+        PG_END_TRY();
```
hornn: fix indentation

src/test/regress/pg_regress_main.c
```diff
@@ -108,12 +108,7 @@ static void
 psql_init(void)
 {
 	/* set default regression database name */
-	if (!dblist)
```
hornn: why is the if check not needed anymore? can't psql_init be called a second time when dblist is not null?

PR #376, opened 2016-02-23
HAWQ-444. The counter for increasing and decreasing memory quota is not correctly subtracted

jiny2:
This code change is to remove the code for the two counters.

### DIFF
--- a/src/test/regress/pg_regress.c
+++ b/src/test/regress/pg_regress.c
@@ -91,8 +91,8 @@ static int	port = -1;
 static char *user = NULL;
 static char *srcdir = NULL;
 static _stringlist *extraroles = NULL;
-char *initfile = NULL;
-char *expected_statuses_file = NULL;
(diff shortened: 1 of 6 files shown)


src/test/regress/pg_regress.c
```diff
@@ -91,8 +91,8 @@ static int	port = -1;
 static char *user = NULL;
 static char *srcdir = NULL;
 static _stringlist *extraroles = NULL;
-char *initfile = NULL;
-char *expected_statuses_file = NULL;
+static char *initfile = "./init_file";
```
sansanichfb: Just curios, what is the reason of this change?

jiny2: Should be a error from git, another irrelevant commit is involved. This is why I close it at once.

sansanichfb: Got it, thanks.

PR #379, opened 2016-02-24
HAWQ-444. The counter for increasing and decreasing memory quota is not correctly subtracted

jiny2:
This fix is to remove counter of increasing and decreasing memory quota.


linwen:
+1

yaoj2:
＋1

PR #381, opened 2016-02-24
HAWQ-447. YARN mode resource manager does not count available segment back in to available segment counter

jiny2:
The fix is to add available segment counter when segment is FTS available again but GRM available already.


huor:
LGTM. +1

yaoj2:
+1

PR #382, opened 2016-02-24
HAWQ-448. hawq_global_rm_type is case sensitive

jiny2:
This makes HAWQ RM able to recognize string of hawq_global_rm_type in a case insensitive way.


yaoj2:
LGTM

huor:
LGTM. +1

PR #385, opened 2016-02-25
HAWQ-452. Remove the workfile views from hawq_toolkit

gcaragea:
In the hawq_toolkit schema we have a few views with live information about disk space used for workfiles (or spillfiles).
Generating this views means running a UDF on every segment. This doesn't apply to the HAWQ 2 architecture:
* The notion of "every segment" is not well defined in HAWQ 2, since we have virtual segments that are created dynamically based on resource requirements and availablitiy
* There is no mechanism to execute a UDF on all the existing virtual or physical segments

Removing the views from hawq_toolkit.


wengyanqing:
I had merged this PR, @gcaragea would you please close it ?

gcaragea:
Merged to master: 87d13b673c19d0975cd2408c732c8c4836640e66, closing. 

@wengyanqing I think it would be better if you can preserve authorship when merging PRs. That way the original author can take the credit / blame for a fix :)

PR #386, opened 2016-02-25
HAWQ-454. 'hawq stop' should return true if master/standby already stopped

radarwave:
Currently if we do 'hawq stop master/standby' while master/standby already in stop stage, we will recognize the stop action is failed.
Now we will recognize they are success.

Passed sanity and multinodeparallel tests.


linwen:
+1

yaoj2:
＋1

PR #388, opened 2016-02-25
HAWQ-453. Do not allocate query resource in prepare stage for prepared statement

huor:
The change includes:
1. In SPI prepare, we generate plan, get and save query resource parameters
2. In SPI execute, we allocate query resource according to saved query resource parameters, and allocate datalocality


ztao1987:
+1

wengyanqing:
+1

zhangh43:
+1

PR #389, opened 2016-02-25
HAWQ-458. Wrongly judge validate connection context progress for failing to acquire resource

jiny2:


jiny2:
git error cause another commit involved

PR #392, opened 2016-02-26
HAWQ-404. Add sort during INSERT of append only AO partition tables

hsyuan:
This PR adds a new GUC for GPORCA: `optimizer_parts_to_force_sort_on_insert `, which implies the minimum number of partitions required to force sorting tuples during insertion in an append only row-oriented partitioned table.

For AO with small number of partitions that is less than the GUC value of `optimizer_parts_to_force_sort_on_insert `, sort is not needed for insert. For a relatively large number of partitions that is larger or equal to `optimizer_parts_to_force_sort_on_insert `, sort is needed.  

Sort during inserting on append only row oriented partitioned tables is disabled by default, with the default GUC value as INT_MAX.

This requires GPORCA version number >= 1.623

### DIFF
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -6150,7 +6150,7 @@ static struct config_int ConfigureNamesInt[] =
 		{"optimizer_parts_to_force_sort_on_insert", PGC_USERSET, DEVELOPER_OPTIONS,
 			gettext_noop("Minimum number of partitions required to force sorting tuples during insertion in an append only row-oriented partitioned table"),
 			NULL,
-			GUC_NOT_IN_SAMPLE
+			GUC_NOT_IN_SAMPLE | GUC_GPDB_ADDOPT
 		},
 		&optimizer_parts_to_force_sort_on_insert,
 		INT_MAX, 0, INT_MAX, NULL, NULL
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -762,6 +762,7 @@ bool 		optimizer_multilevel_partitioning;
 bool        optimizer_enable_derive_stats_all_groups;
 bool		optimizer_explain_show_status;
 bool		optimizer_prefer_scalar_dqa_multistage_agg;
+int		optimizer_parts_to_force_sort_on_insert;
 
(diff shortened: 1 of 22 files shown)


huor:
LGTM. +1

src/include/nodes/plannodes.h
```diff
@@ -1149,6 +1149,7 @@ typedef struct DML
 	AttrNumber	actionColIdx;	/* index of action column into the target list */
 	AttrNumber	ctidColIdx;		/* index of ctid column into the target list */
 	AttrNumber	tupleoidColIdx;	/* index of tuple oid column into the target list */
+	bool		needsSort;		/* needs the data to be sorted */
```
vraghavan78: inputSorted

vraghavan78:
Looks good!

src/backend/utils/misc/guc.c
```diff
@@ -762,6 +762,7 @@ bool 		optimizer_multilevel_partitioning;
 bool        optimizer_enable_derive_stats_all_groups;
 bool		optimizer_explain_show_status;
 bool		optimizer_prefer_scalar_dqa_multistage_agg;
+int			optimizer_parts_to_force_sort_on_insert;
```
gcaragea: indentation is off the hook

gcaragea:
Just to confirm: this new flag in the DML node is set in all cases when the input is sorted on partition id. By "all cases" I mean: 
- destination is AO table, and number of parts is > optimizer_parts_to_force_sort_on_insert
- destination is Parquet table and gp_parquet_insert_sort is true
- any other case when ORCA inserts a sort on partition id before the DML node

Correct?

hsyuan:
Currently, only the first 2 cases will set the flag to true. no other case at this moment.

@vraghavan78

vraghavan78:
Only the first 2 case it is true. All other cases in false.

src/include/nodes/plannodes.h
hsyuan: Changed.

xinzweb:
@changleicn , please go ahead and merge it. Thanks. (Please don't squash the commits)

dyozie:
LGTM

PR #393, opened 2016-02-26
HAWQ-198. Build-in functions will be executed in resource_negotiator stage which causes it will be executed twice

huor:


liming01:
Looks good to me. +1.

src/backend/optimizer/plan/planner.c
```diff
@@ -309,6 +339,8 @@ planner(Query *parse, int cursorOptions,
       {
         resource_negotiator(parse, cursorOptions, boundParams, resourceLife, &ppResult);
 
+		decrease_planning_depth();
```
zhangh43: better to move decrease_planning_depth to the end of try. (just before end_memory_accounting)

zhangh43:
LGTM
better to move decrease_planning_depth to the end of try

PR #394, opened 2016-02-28
HAWQ-467. Add checks to HAWQ master start/stop

radarwave:
1. We should check if HAWQ master is doing recovery.
2. When doing hawq master stop, check if there any existing connections. Master stop abort if use smart stop, give warning if use fast stop.
3. Hawq start/stop use 600 seconds as default timeout.


liming01:
Looks fine for me. +1.

yaoj2:
LGTM+1

PR #396, opened 2016-02-29
HAWQ-473. Insert a row into gp_configuration_history when segment sta…

linwen:
Implement adding a entry into gp_configuration_history when segment' status is changed, so that users can know the reason of status change by querying this catalog table.
A cleanup function is needed when user decides to clear this table.
Please review. Thanks


jiny2:
I think it is better to improve this fix to update GRM availability of segments, currently, if a segment is FTS available, it is set in gp_segment_configuration as 'u' regardless the status of GRM availability, i.e. yarn availability. Can this fix include the following functions:
1) in not NONE mode, update gp_segment_configuration according to GRM availability as well, 
2) save the availability history in gp_segment_configuration caused by the changes of GRM availability.

linwen:
reasonable. Thank you!

PR #400, opened 2016-03-01
HAWQ-477. Copy table to file do not real execute for lineitem.

zhangh43:
GetFileSplitsOfSegment should be called after appendonly_beginscan


wengyanqing:
+1

PR #402, opened 2016-03-02
HAWQ-472. Add checks to HAWQ standby start up

radarwave:
We should add HAWQ standby start up check, to make sure standby master is started successfully.
Now if standby master is not initialized, then start standby will report success but not failed.
Use fast mode to stop master while remove standby master.


yaoj2:
+1

PR #403, opened 2016-03-03
HAWQ-481. Optimise analyse virtual segment number for random tables.

zhangh43:
when there is no prefer seg number, we use hash table bucket number and if no hash table, we use the total size of all the random table(1 vseg for 128M data)


ictmalili:
In function calculate_virtual_segment_number , what if the former table has larger bucket num, but the latter table has smaller one? maxHashBucketNumber should be the larger value between  maxHashBucketNumber and  targetPolicy->bucketnum.
maxHashBucketNumber = targetPolicy->bucketnum;

ictmalili:
Looks good to me now. +1

PR #404, opened 2016-03-03
HAWQ-484. Fix bug in gpcopy and pgdump bug.

zhangh43:
targetPolicy may be null in initialization
copy case should consider parquet code branch


wengyanqing:
+1

PR #406, opened 2016-03-03
HAWQ-483. Fix bug for Error in debug1 state.

realdawn:
"serializedIdentity" doesn't include '\0', which will cause core dump in SetupProcessIdentity() with using elog(DEBUG1).
So palloc a new string with '\0'.

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -6204,4 +6204,3 @@ SyncAgentMain(int argc, char *argv[], const char *username)
 	return 1;					/* keep compiler quiet */
 }
 
-
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -4760,6 +4760,7 @@ PostgresMain(int argc, char *argv[], const char *username)
 					const char *serializedIdentity = NULL;
 					const char *serializedResource = NULL;
 					
+					char *completeSerializedIdentity = NULL;
 					int query_string_len = 0;
 					int serializedSnapshotlen = 0;
 					int serializedQuerytreelen = 0;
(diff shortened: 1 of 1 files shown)


ictmalili:
Looks good to me. +1

ictmalili:
I think it's better to put the code inside function SetupProcessIdentity, instead of putting it in the PostgresMain, since the function already takes up so many lines.

src/backend/tcop/postgres.c
```diff
@@ -4880,7 +4881,18 @@ PostgresMain(int argc, char *argv[], const char *username)
...
+					{
+						/*
+						 * serializedIdentity doesn't include '\0', which will cause core dump in SetupProcessIdentity() with using elog(DEBUG1).
+						 * So palloc a new string with '\0'.
+						 */
+						completeSerializedIdentity = (char *) palloc((serializedIdentityLen + 1) * sizeof(char) );
+						memcpy(completeSerializedIdentity, serializedIdentity, serializedIdentityLen);
+						completeSerializedIdentity[serializedIdentityLen] = '\0';
+						serializedIdentity = completeSerializedIdentity;
 						SetupProcessIdentity(serializedIdentity);
```
huor: Here pfree only handle the situation when SetupProcessIdentity successfully compltes. We also need to do the pfree if it errors in SetupProcessIdentity.

realdawn: As the MemoryContext which is MessageContext will reset every time, so it will not cause memory leak even it  errors.

realdawn:
If I put the code into function SetupProcessIdentity, I must put serializedIdentityLen into the function which will change the function interface. So I just make the minimal change.

PR #408, opened 2016-03-03
HAWQ-487. Ensure virtual segment of analyze not greate than GetUtilPa…

zhangh43:
rtitionNum
analyze on big table should use less vsegs to reach better performance.


yaoj2:
+1

PR #409, opened 2016-03-03
HAWQ-488. Fix bug which cannot drop schema madlib and re-deploy MADlib in schema madlib

huor:
The fix are in two folds:

1. Cannot drop schema madlib and re-deploy MADlib in schema madlib

2. pg_dump skip dumping content of schema madlib/retail_demo, which should be dumped by default actually. To skip dumping some schema, you can specify it in command line option of pg_dump.


liming01:
Looks fine to me. +1.

PR #412, opened 2016-03-04
HAWQ-462. Adding dfs_address from pgfilespace along with fragment dat…

shivzone:
HAWQ-317 attempts to use the dfs_address from pg_filespace_entry. The segments (gpbridge) do not have catalog access. Adding the dfs_address along with the fragment data passed from the master to the segments. This fixes the pxf handling of secure isilon and hcatalog with or without HA

### DIFF
--- a/src/backend/access/external/hd_work_mgr.c
+++ b/src/backend/access/external/hd_work_mgr.c
@@ -769,11 +769,15 @@ make_allocation_output_string(List *segment_fragments)
 	initStringInfo(&segwork);
 	appendStringInfoString(&segwork, SEGWORK_PREFIX);
 	
-	/* Add dfs_address from pg_filespace to the segment data. Fixes HAWQ-462 *//* dfs_address from pg_filespace entry */
-	char* dfs_address = NULL;
-	get_hdfs_location_from_filespace(&dfs_address);
-	appendStringInfoString(&segwork, dfs_address);
-	appendStringInfoChar(&segwork, SEGWORK_DFS_DELIM);
+	if (enable_secure_filesystem)
+	{
+		/* Add dfs_address from pg_filespace to the segment data. Fixes HAWQ-462 *//* dfs_address from pg_filespace entry */
+		char* dfs_address = NULL;
+		get_hdfs_location_from_filespace(&dfs_address);
+		appendStringInfoString(&segwork, dfs_address);
+		appendStringInfoChar(&segwork, SEGWORK_DFS_DELIM);
+		pfree(dfs_address);
+	}
 
 	foreach(frag_cell, segment_fragments)
 	{
@@ -802,7 +806,6 @@ make_allocation_output_string(List *segment_fragments)
 		appendStringInfoString(&segwork, fragment_str.data);
 		pfree(fragment_str.data);
 	}
-	pfree(dfs_address);
 
 	return segwork.data;
 }
--- a/src/backend/access/external/hd_work_mgr.c
+++ b/src/backend/access/external/hd_work_mgr.c
@@ -768,6 +768,9 @@ make_allocation_output_string(List *segment_fragments)
 	initStringInfo(&segwork);
 	appendStringInfoString(&segwork, SEGWORK_PREFIX);
 	
+	char* dfs_address = NULL;
+	get_hdfs_location_from_filespace(&dfs_address);
+
 	foreach(frag_cell, segment_fragments)
 	{
 		AllocatedDataFragment *frag = (AllocatedDataFragment*)lfirst(frag_cell);
@@ -783,6 +786,10 @@ make_allocation_output_string(List *segment_fragments)
 		appendStringInfoChar(&fragment_str, SEGWORK_IN_PAIR_DELIM);
 		if (frag->fragment_md)
 			appendStringInfo(&fragment_str, "%s", frag->fragment_md);
+		/* Adding dfs_address from pg_filespace entry required for HAWQ-462 */
+		appendStringInfoChar(&fragment_str, SEGWORK_IN_PAIR_DELIM);
+		if (dfs_address)
+			appendStringInfo(&fragment_str, "%s", dfs_address);
 		if (frag->user_data)
 		{
 			appendStringInfoChar(&fragment_str, SEGWORK_IN_PAIR_DELIM);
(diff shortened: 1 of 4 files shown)


src/backend/access/external/hd_work_mgr.c
```diff
@@ -783,6 +786,10 @@ make_allocation_output_string(List *segment_fragments)
 		appendStringInfoChar(&fragment_str, SEGWORK_IN_PAIR_DELIM);
 		if (frag->fragment_md)
 			appendStringInfo(&fragment_str, "%s", frag->fragment_md);
+		/* Adding dfs_address from pg_filespace entry required for HAWQ-462 */
+		appendStringInfoChar(&fragment_str, SEGWORK_IN_PAIR_DELIM);
+		if (dfs_address)
```
hornn: why do we need to pass this value for each fragment? wouldn't it be better to pass it just once, perhaps append it in the end of the segments string (after the loop)?

src/backend/access/external/pxfuriparser.c
```diff
@@ -709,7 +709,15 @@ GPHDUri_parse_fragment(char* fragment, List* fragments)
 	*value_end = '\0';
```
hornn: in line 666, please update the expected syntax.

src/backend/access/external/pxfuriparser.c
```diff
@@ -719,7 +727,7 @@ GPHDUri_parse_fragment(char* fragment, List* fragments)
 		has_user_data = true;
 		*value_end = '\0';
 	}
-	fragment_data->fragment_md = pstrdup(value_start);
+	fragment_data->dfs_address = pstrdup(value_start);
```
hornn: please make sure to pfree it in GPHDUri_free_fragments.
I also noticed that we don't free fragment_md and user_data.

hornn:
General question - does it also solve the problem for writable? I thought for writable tables the code path didn't pass through hd_work_mgr.

GodenYao:
can you also check if you differentiate secure and non-secure queries - for non-secure query we don't need dfs_url info

src/backend/access/external/hd_work_mgr.c
GodenYao: Second that. it'll be an overhead for each fragment to have the duplicated info.

shivzone: Yes, this is the first cut solution, the optimal solution is to pass it as part of the segment level data instead of fragment. Wasn't quite able to figure out how to add additional metadata with the segment data. Wanted to first certify against secure cluster before investigating addition into segment data.

src/backend/access/external/pxfuriparser.c
shivzone: Done

src/backend/access/external/pxfuriparser.c
shivzone: Added

shivzone:
Yet to explore the writable table path.

shivzone:
The dfs_address will not be required in the case of non-secure cluster. Will handle this once we update this to be passed via segment data as opposed to fragment data

src/backend/access/external/hd_work_mgr.c
shivzone: I've updated the pull request to move the dfs address to seg level metadata instead of fragment level metadata

src/include/access/pxfuriparser.h
```diff
@@ -76,6 +76,9 @@ typedef struct GPHDUri
 	 * NNHAConf  will also occupy <host> and <port> members
 	 */
 	NNHAConf        *ha_nodes;
+
+	/* dfs address from pg_filespace (optional) */
+	char			*dfs_address;
```
GodenYao: indention seems off.

src/backend/access/external/hd_work_mgr.c
```diff
@@ -768,6 +769,12 @@ make_allocation_output_string(List *segment_fragments)
 	initStringInfo(&segwork);
 	appendStringInfoString(&segwork, SEGWORK_PREFIX);
 	
+	/* Add dfs_address from pg_filespace to the segment data. Fixes HAWQ-462 *//* dfs_address from pg_filespace entry */
+	char* dfs_address = NULL;
+	get_hdfs_location_from_filespace(&dfs_address);
```
GodenYao: probably add check to see if it's secure or not. enable_secure_filesystem == true

shivzone: Updated to check for secure

src/backend/access/external/pxfuriparser.c
```diff
@@ -634,9 +638,20 @@ GPHDUri_parse_segwork(GPHDUri *uri, const char *uri_str)
...
 	if (segwork == NULL)
 		return;
-
 	segwork += strlen(segwork_substring);
 
+	if (enable_secure_filesystem)
+	{
+		/* parse dfs address */
+		size_end = strchr(segwork, segwork_dfs_separator);
+		if(size_end != NULL)
```
hornn: missing space after if

src/backend/access/external/pxfuriparser.c
```diff
@@ -744,6 +759,9 @@ GPHDUri_free_fragments(GPHDUri *uri)
 		pfree(data->authority);
 		pfree(data->index);
 		pfree(data->source_name);
+		pfree(data->fragment_md);
+		if(data->user_data)
```
hornn: missing space after if

src/backend/access/external/pxfuriparser.c
hornn: if it's a secured environment, if there is no dfs address data maybe we should error out. What do you think?

src/backend/access/external/pxfuriparser.c
```diff
@@ -124,6 +125,9 @@ freeGPHDUri(GPHDUri *uri)
 	if (uri->ha_nodes)
 		GPHD_HA_release_nodes(uri->ha_nodes);
 
+	if (uri->dfs_address)
```
hornn: :)

hornn:
Looks good. Do you plan to open a separate story for writable?

src/backend/access/external/pxfuriparser.c
shivzone: Should this be checked in the master instead before we add this info to the segwork data ?

hornn: Probably. And then here just add an assert that it's expected to not be null.

shivzone:
This pull request is closed as it is outdated.
Refer to the below 2 pull requests which solves it more generically for both readable and writable tables.
https://github.com/apache/incubator-hawq/pull/503
https://github.com/apache/incubator-hawq/pull/604

PR #415, opened 2016-03-07
HAWQ-495. check null cond for explain analyse

wengyanqing:
1. check null cond for explain analyse
2. remove metadatacache dead code


zhangh43:
+1

PR #419, opened 2016-03-09
HAWQ-499: Fix IPv6 on OSX > 10.11

jfrank-pivotal:
On OSX < 10.11, IPv6 addresses are currently fetched with a command that doesn't exist on OSX.

The impact is that `hawq init cluster -a` **fails**:
`/usr/local/hawq/bin/lib/hawqinit.sh: line 196: ip: command not found`

This PR introduces a quick fix for that issue that is exactly the same as the corresponding fix to the IPv4 path.

### DIFF
--- a/tools/bin/lib/hawqinit.sh
+++ b/tools/bin/lib/hawqinit.sh
@@ -190,7 +190,7 @@ LOAD_GP_TOOLKIT () {
 }
 
 get_master_ipv6_addresses() {
-    if [ "${distro_based_on}" = "Mac" ] && [ "${distro_version:0:5}" = "10.11" ]; then
+    if [ "${distro_based_on}" = "Mac" ]; then
         MASTER_IPV6_LOCAL_ADDRESS_ALL=(`${IFCONFIG} | ${GREP} inet6 | ${AWK} '{print $2}' | cut -d'%' -f1`)
     else
         MASTER_IPV6_LOCAL_ADDRESS_ALL=(`ip -6 address show |${GREP} inet6|${AWK} '{print $2}' |cut -d'/' -f1`)


vraghavan78:
LGTM!

jfrank-pivotal:
@radarwave Can you take a look at this PR?  Thanks!

radarwave:
'ip' command can be installed on the machine which have OSX<10.11, then we don't need to fix this.

Not quite sure now, but as my remember, ifconfig have some issues to get the ipv6 ip addresses on some old Mac OS. So we turned to use command 'ip'.

If you can test on the previous version OSX, then this PR is good to go. 

Thanks.

hsyuan:
LGTM.

jfrank-pivotal:
We don't run any of those OS versions here, but we can try to test it with Vagrant.  Alternatively, we can add the `ip` executable to the list of toolchain requirements (but I'd prefer the former).  @radarwave, what OSX version should we target for testing?

xinzweb:
@changleicn  and @radarwave please test this PR on your side, and if all good, please push them in. Thanks.

radarwave:
Tested on 10.11, 10.10, 10.9, it works. Will push this in.

As check, the ifconfig output not the same issue is handled in other place. At least I can see redhat7 have different ifconfig output than the previous redhat versions. See below 'inet' vs 'inet addr':

rhel6:
eth0      Link encap:Ethernet  HWaddr 0A:5A:F1:42:BE:29
          inet addr:10.32.35.35  Bcast:10.32.35.255  Mask:255.255.254.0
          inet6 addr: fe80::85a:f1ff:fe42:be29/64 Scope:Link

rhel7:
ens32: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.168.2.100  netmask 255.255.255.0  broadcast 192.168.2.255
        inet6 fe80::250:56ff:fe01:223b  prefixlen 64  scopeid 0x20<link>

radarwave:
@jfrank-pivotal , please close this pull request since it's already been merged. Thanks.

PR #421, opened 2016-03-09
HAWQ-504. Crash when WALSendServer and standby connection timeout

liming01:
Here just fixed the timeout problem. The auto re-sync standby functionality will be implemented as a feature in the future.


wengyanqing:
+1

PR #427, opened 2016-03-10
HAWQ-507. Remove useless code for cdbpersistentrelfile.c

realdawn:
As the code coverage of cdbpersistentrelfile.c is only 59% due to dead code, so we should remove the dead code in that file.


liming01:
+1.

yaoj2:
＋1

PR #428, opened 2016-03-10
HAWQ-509. Change pg_stat_activity definition, only returns one record…

ictmalili:
Change pg_stat_activity definition, only returns one record for one query.


zhangh43:
+1

PR #429, opened 2016-03-10
HAWQ-510. Change cdbProc->listenerAddr to "localhost" on entryDB when…

ictmalili:
HAWQ-510. Change cdbProc->listenerAddr to "localhost" on entryDB when it tries to parse connection from QD, instead of direct using "[local]"


wengyanqing:
+1

huor:
LGTM. +1

PR #433, opened 2016-03-11
Fixing error message for exttab1 source file when optimizer is turned…

vraghavan78:
@changleicn @huor Please take a look.

Venky

### DIFF
--- a/src/test/regress/output/exttab1.source
+++ b/src/test/regress/output/exttab1.source
@@ -669,7 +669,7 @@ HINT:  use the gpfdist protocol or COPY FROM instead
 create writable external table wet_neg1(a text, b text) location('gpfdist://@hostname@:7070/wet.out', 'gpfdist://@hostname@:7070/wet.out') format 'text';
 ERROR:  location uri "gpfdist://@hostname@:7070/wet.out" appears more than once
 create writable external web table wet_pos5(a text, b text) execute 'some command' on segment 0 format 'text';
-ERROR:  ON clause may not be used with a writable external table
+ERROR:  the ON segment syntax for writable        external tables is deprecated
 --
 -- SELECT from WET (negative)
 --


src/test/regress/output/exttab1.source
```diff
@@ -669,7 +669,7 @@ HINT:  use the gpfdist protocol or COPY FROM instead
 create writable external table wet_neg1(a text, b text) location('gpfdist://@hostname@:7070/wet.out', 'gpfdist://@hostname@:7070/wet.out') format 'text';
 ERROR:  location uri "gpfdist://@hostname@:7070/wet.out" appears more than once
 create writable external web table wet_pos5(a text, b text) execute 'some command' on segment 0 format 'text';
-ERROR:  ON clause may not be used with a writable external table
+ERROR:  the ON segment syntax for writable        external tables is deprecated
```
hornn: should there be so many spaces in the middle of the error?

vraghavan78: I just copied the expected error message from the source file for the optimizer. There was a change in the error message but the source file when planner is on was not changed.

hornn: just looks like a typo: `writable        external`

changleicn:
looks better to have a jira for this and remove the extra spaces.

ryantang:
Can we make the test green now, and then adjust the error message (and the corresponding test) in a later story? 

This is the last red test in our effort of a passing installcheck_good run for HAWQ without optimizer. Prolonging the red means that we'll continue to hide potential meaningful test failures.

tom-meyer:
ASF's jira server is down, as of Friday evening (Pacific time). So we don't have an associated jira issue.

tom-meyer:
This should be closed. We've made a new pull request that references a jira: https://github.com/apache/incubator-hawq/pull/449

PR #436, opened 2016-03-11
HAWQ-510. Modify proc->listenerAddr to 127.0.0.1 in entryDB for conne…

ictmalili:
Modify proc->listenerAddr to 127.0.0.1 in entryDB for connection to QD


ztao1987:
+1

PR #437, opened 2016-03-12
Fix expected function.out when orca is not enabled

tom-meyer:
[Venky's fix ](https://github.com/apache/incubator-hawq/pull/433) is similar to this.


tom-meyer:
ASF's jira server is down, as of Friday evening (Pacific time). So we don't have an associated jira issue.

tom-meyer:
We've made a new pull request that references a jira issue: https://github.com/apache/incubator-hawq/pull/448

PR #444, opened 2016-03-14
HAWQ-524. Modify processing for cdbDispatchResult->resultBuf

ictmalili:
Modify processing for cdbDispatchResult->resultBuf when creation and free.
Should judge whether NULL after creation and when do destroy.


zhangh43:
+1

PR #445, opened 2016-03-14
HAWQ-528. Reset gp_connections_per_thread for dispatcher guc range

ictmalili:
Reset gp_connections_per_thread for dispatcher guc range from 1 to 512, 0 is invalid.


zhangh43:
+1

PR #448, opened 2016-03-14
HAWQ-536. Fix expected function.out when orca is not enabled

tom-meyer:


tom-meyer:
Although this PR has not been accepted, the commit https://github.com/apache/incubator-hawq/commit/d259efffbce3d9b1eb4e68646632d6419cf2c745 has made the same change, so I'll close it.

PR #451, opened 2016-03-15
HAWQ-541. When YARN containers disappear from container report, the segment available resource maybe negative

jiny2:


yaoj2:
+1

PR #455, opened 2016-03-15
HAWQ-542. Passing empty string to attribute vseg_resource_quota makes default value set instead of reporting error

jiny2:


jiny2:
This is a wrong pull request, I close it.

PR #460, opened 2016-03-16
HAWQ-548. Fix wrong unit test case for ip address parsing

ictmalili:
Fix wrong unit test case for ip address parsing of interconnect


ztao1987:
+1

PR #461, opened 2016-03-16
HAWQ-550. Automatically clean up records in gp_configuration_history longer than a GUC

linwen:
HAWQ-550. Automatically clean up records in gp_configuration_history longer than a GUC value.
default is 365 days.
Please review.


jiny2:
+1 The code looks good to me.

huor:
+1

PR #462, opened 2016-03-16
HAWQ-549. Ignore empty 'pg_log' folder while check standby directories

radarwave:
As Ambari's requirement, we should ignore empty 'pg_log' folder while doing standby directory check.
If standby data directory only have an empty pg_log, we assume it's clean enough to initialize standby master.


yaoj2:
LGTM

liming01:
+1.

PR #465, opened 2016-03-18
optimize udf resource, copy analyze and so on

zhangh43:
udf request resource in top level to accelerate madlib
fix copy using to many vsegs 
fix analyze rootpartition inaprropriate vseg number


wengyanqing:
LGTM

PR #469, opened 2016-03-18
HAWQ-558. The query running on random table does not push resource manager to request resource from YARN as maimum possible resource consumption

jiny2:


yaoj2:
+1

radarwave:
+1

PR #472, opened 2016-03-18
HAWQ-559. Fix QD hang when QE is immediately killed after created

ictmalili:
Add errcode for executor after QD thread detects QE is no longer exists in function executormgr_is_dispatchable.


ztao1987:
+1

PR #473, opened 2016-03-18
HAWQ-560. Set 'output.replace-datanode-on-failure' depend on datanode…

radarwave:
…s number

Set 'output.replace-datanode-on-failure' depend on datanode number.
By default, we should set this value while hawq cluster init. To set this value automatically, we assume segment number is the same as datanode number.
When datanodes < 4, change the default value from true to false.

### DIFF
--- a/tools/bin/hawq_ctl
+++ b/tools/bin/hawq_ctl
@@ -182,6 +182,18 @@ class HawqInit:
             logger.warn("Set standby host name failed")
         return result
 
+    def set_number_by_datanodes_num(self):
+        reuslt = 0
+        if self.hosts_count_number < 4:
+            # Set output.replace-datanode-on-failure to false if datanode number less than 4.
+            logger.info("Set output.replace-datanode-on-failure to false as %s datanode detected" % self.hosts_count_number)
+            logger.info("If you actually have more than 4 datanodes, please set it to 'true'")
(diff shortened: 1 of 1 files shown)


ztao1987:
+1

tools/bin/hawq_ctl
```diff
@@ -182,6 +182,18 @@ class HawqInit:
             logger.warn("Set standby host name failed")
         return result
 
+    def set_number_by_datanodes_num(self):
+        reuslt = 0
+        if self.hosts_count_number < 4:
+            # Set output.replace-datanode-on-failure to false if datanode number less than 4.
+            logger.info("Set output.replace-datanode-on-failure to false as %s datanode detected" % self.hosts_count_number)
+            logger.info("If you actually have more than 4 datanodes, please set it to 'true'")
+            cmd = "hawq config -c 'output.replace-datanode-on-failure'  -v 'false' --skipvalidation -q > /dev/null"
```
yaoj2: the parameter is in hdfs client

radarwave:
Thanks, will modify and submit for review again.

tools/bin/hawq_ctl
```diff
@@ -182,6 +182,18 @@ class HawqInit:
             logger.warn("Set standby host name failed")
         return result
 
+    def set_number_by_datanodes_num(self):
+        reuslt = 0
```
denalex: typo here, should be result

PR #478, opened 2016-03-22
HAWQ 569. Fixing an assertion failure and a potential double release of workfile_queryspace queryEntry

armenatzoglou:
This commit ports Foyzur's fix on assertion failure because of multiple ExecutorEnd calls (story 114507991).

### DIFF
--- a/src/backend/utils/workfile_manager/workfile_queryspace.c
+++ b/src/backend/utils/workfile_manager/workfile_queryspace.c
@@ -302,7 +302,6 @@ WorkfileQueryspace_ReleaseEntry(void)
 	}
 
 	querySpaceNestingLevel--;
-	Assert(querySpaceNestingLevel >= 0);
 
 	if (querySpaceNestingLevel > 0)
 	{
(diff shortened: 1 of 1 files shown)


liming01:
Hi @armenatzoglou,

I had merged it into branch master for the release date coming. 

But I have a question: it seems that your fix doesn't fix the problem: 
(1) you just move the line  'queryEntry = NULL;'  one line( if clause ) ahead.
(2) moreover, the interrupt may occurs before the line 'queryEntry = NULL;' after called SyncHTRelease(). 

The solution I suggest: 
(1)If all memory and resource can be released even if WorkfileQueryspace_ReleaseEntry() not called. e.g. all memory palloced will be released for specific memory context released( transaction finished/statement execution finished....), then we can move the line 'queryEntry = NULL;'  to the line before calling SyncHTRelease().
(2) If cannot be auto-cleaned, then we should "HOLD_INTERRUPTS();" until we set  'queryEntry = NULL;' 

Thanks.

liming01:
Hi @armenatzoglou ,

Please ignore the comments above. 
I look into the code of SyncHTRelease(), there is no calling 'CHECK_FOR_INTERRUPTS()' in this function or the functions it called if there is no error reported. So your fix can work I think.

Please close this PR. Thanks.

armenatzoglou:
Thanks

PR #485, opened 2016-03-22
HAWQ-575. QE core dumped when report "Resume interrupt holdoff count is bad (0)…

liming01:
… (xact.c:2907)"

The root cause is: 
AT AbortTransaction(),  we will disable interrupt processing by calling  "HOLD_INTERRUPTS();" and "RESUME_INTERRUPTS();", these 2 functions will change "InterruptHoldoffCount" to indicate that we are in such a section. When in such a section, we can't report error, because it will set  InterruptHoldoffCount to 0. 

On the other side, from the user demand, when transaction abort, we can't report error if temp file cannot be deleted, we need to just report warning.


huor:
+1

linwen:
+1

PR #486, opened 2016-03-22
HAWQ-573. Fix Resource leak when cancel a query during calling a RPC …

realdawn:
…to AllocateResource.

When QD is running in the function processAllCommFileDescs (called by acquireResourceFromRM, which is a RPC to allocate resource from QD to RM), it received a "query cancel pending'" interrupt, and it handles the interrupt and error out. So we should mark and clean up resource as resource could have been allocated in RM side. Otherwise it will cause resource leak.

### DIFF
--- a/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
+++ b/src/backend/resourcemanager/communication/rmcomm_QD2RM.c
@@ -31,6 +31,7 @@
 #include "resourcemanager/resourcemanager.h"
 #include "resourcemanager/conntrack.h"
 #include "resourcemanager/communication/rmcomm_MessageProtocol.h"
+#include "tcop/pquery.h"
 
 #include "funcapi.h"
 #include "fmgr.h"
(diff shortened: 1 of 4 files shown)


src/backend/resourcemanager/communication/rmcomm_QD2RM.c
```diff
@@ -461,7 +462,22 @@ int cleanupQD2RMComm(void)
     {
         if ( QD2RM_ResourceSets[i] != NULL )
         {
-            if ( QD2RM_ResourceSets[i]->QD_ResourceList != NULL )
+            if (ResourceIndex == i)
```
jiny2: I think this if block can be merged into the else block.

jiny2: What does ResourceIndex stand for? why it needs special processing logic comparing with the other resource contexts? Is it ok to use the same logic in the else block below?

realdawn:
I will change the code and pull a new request.

PR #489, opened 2016-03-22
HAWQ-537. Accept filenames with spaces

cjcjameson:
Without this fix, the `make distdir` target fails because files such as `./src/pl/pljava/.externalToolBuilders/Header\ generation.launch` have spaces in their names

### DIFF
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -129,7 +129,7 @@ postgresql-test-$(VERSION).tar: distdir
 
 distdir:
 	-rm -rf $(distdir)* $(dummy)
-	for x in `cd $(top_srcdir) && find . -name CVS -prune -o -print`; do \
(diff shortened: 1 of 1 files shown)


cjcjameson:
@radarwave I think this is still a valid PR; can you comment and close?

radarwave:
@cjcjameson Most looks good to me.
Please use this template file 'GNUMakefile.in' to generate 'GNUMakefile' and test the run result. They should be check-in together.  Thanks.

radarwave:
@cjcjameson Apologize to the review mistake, I must confused with the old way HAWQ used to be. In the past, HAWQ requires check-in Makefile beside Makefile.in when we put 'configure' inside Makefile.

Now we do not need to check-in GNUMakefile.

@xunzhang @paul-guo- Please help to review too, one more +1 I can merge this commit it. 

Thanks.

paul-guo-:
+1

xunzhang:
LGTM

PR #490, opened 2016-03-23
HAWQ-543. Remove references to bootstrap_tokens.h

cjcjameson:
`bootstrap_tokens.h` isn't a file that exists anymore; was causing `make distprep` to fail


hsyuan:
+1

PR #491, opened 2016-03-23
HAWQ-573. Fix Resource leak when cancel a query during calling a RPC …

realdawn:
…to allocate resource.

When QD is running in the function processAllCommFileDescs (called by acquireResourceFromRM, which is a RPC to allocate resource from QD to RM), it received a "query cancel pending'" interrupt, and it handles the interrupt and error out. So we should add the resource into the GlobalQueryResources before it actually allocate. Otherwise it will cause resource leak.


jiny2:
+1 LGTM

zhangh43:
+1

PR #497, opened 2016-03-23
Writable security fix

kavinderd:


kavinderd:
@hornn please take a look

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -313,6 +313,22 @@ FinalizeQueryContextInfo(QueryContextInfo *cxt)
 			WriteData(cxt, buffer.data, buffer.len);
 			WriteData(cxt, credential, size);
         }
+
+    }
+
+    //Add Namespace for PXF
```
hornn: please change comment to c standard `/* */`

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -313,6 +313,22 @@ FinalizeQueryContextInfo(QueryContextInfo *cxt)
...
+    if(Gp_role != GP_ROLE_EXECUTE){
+    	const char *namespace = cxt->sharedPath;
+		int size = strlen(namespace);
```
hornn: indentation look off...

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -616,6 +632,37 @@ RebuildFilesystemCredentials(QueryContextInfo *cxt, HTAB ** currentFilesystemCre
 	pfree(binary);
 }
 
+
+/*
+ * deserialize the namespace data
+ */
+
```
hornn: remove new line

src/backend/gpopt/ivy.xml
```diff
@@ -38,7 +38,7 @@ under the License.
     </configurations>
 
     <dependencies>
-      <dependency org="emc"             name="optimizer"       rev="1.623"          conf="osx106_x86->osx106_x86_32;osx106_x86_32->osx106_x86_32;rhel5_x86_64->rhel5_x86_64;suse10_x86_64->suse10_x86_64" />
+      <dependency org="emc"             name="optimizer"       rev="1.625"          conf="osx106_x86->osx106_x86_32;osx106_x86_32->osx106_x86_32;rhel5_x86_64->rhel5_x86_64;suse10_x86_64->suse10_x86_64" />
```
hornn: relevant to this commit?

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -273,6 +277,26 @@ void gpbridge_export_start(PG_FUNCTION_ARGS)
 
 }
 
+void parse_namespace(char *uri, char** extracted_namespace){
+	int i = 0;
+	int len = strlen(uri);
+
+	while(uri[i] != '*' && i < len){
```
hornn: `*` can be part of the PXF location (URL), because we support regex matching for file names (e.g. `file*.txt`).
If this is going to be a permanent part of the URL (not just for secured env), perhaps we can scan from the end? 
Another option is to replace the delimiter.

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -273,6 +277,26 @@ void gpbridge_export_start(PG_FUNCTION_ARGS)
...
+		i++;
+	}
+
+	uri[i++] = '\0';
+	if(i < len){
+		*extracted_namespace = palloc(len - i + 1);
+		int extracted_namespace_len = len - i;
+		int j = 0;
+		while(i < len && j < extracted_namespace_len){
```
hornn: I think this loop can be replaced by `pnstrdup`

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -508,17 +532,16 @@ size_t fill_buffer(gphadoop_context* context, char* start, size_t size)
...
 
 	if (!enable_secure_filesystem)
 		return;
 
 	token = palloc0(sizeof(PxfHdfsTokenData));
 
-	get_hdfs_location_from_filespace(&dfs_address);
-
-    elog(DEBUG2, "locating token for %s", dfs_address);
+//	get_hdfs_location_from_filespace(&dfs_address);
```
hornn: remove :)

src/include/access/fileam.h
```diff
@@ -44,12 +44,13 @@
  * ExternalInsertDescData is used for storing state related
  * to inserting data into a writable external table.
  */
-typedef struct ExternalInsertDescData
+typedef struct ExternalInsertDescDat
```
hornn: typo?

src/include/access/fileam.h
```diff
@@ -44,12 +44,13 @@
...
 {
 	Relation		ext_rel;
 	FILE*			ext_file;
 	char*			ext_uri;		/* "command:<cmd>" or "tablespace:<path>" */
 	bool			ext_noop;		/* no op. this segdb needs to do nothing (e.g. mirror seg) */
+	char*           ext_namespace;  /* Namespace data */
```
hornn: indentation looks off

src/bin/gpupgrade/crc32c.c
```diff
@@ -0,0 +1 @@
+../../../src/backend/utils/hash/crc32c.c
```
hornn: why are these 3 files needed?

src/backend/access/external/fileam.c
```diff
@@ -681,6 +687,8 @@ external_insert(ExternalInsertDesc extInsertDesc, HeapTuple instup)
 
 		fcinfo.arg[0] = HeapTupleGetDatum(instup);
 		fcinfo.argnull[0] = false;
+		fcinfo.arg[1] = "test";
```
hornn: what does it mean?

src/backend/access/external/hd_work_mgr.c
```diff
@@ -768,6 +769,16 @@ make_allocation_output_string(List *segment_fragments)
 	initStringInfo(&segwork);
 	appendStringInfoString(&segwork, SEGWORK_PREFIX);
 	
+	if (enable_secure_filesystem)
```
hornn: if we append the dfs_address through the external framework, why do we need to append it here as well? wouldn't it work for both readable and writable tables?

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -616,6 +632,37 @@ RebuildFilesystemCredentials(QueryContextInfo *cxt, HTAB ** currentFilesystemCre
...
+	len = (int) ntohl(*(uint32 *) buffer);
+	binary = palloc(len);
+	if(ReadData(cxt, binary, len, TRUE))
+	{
+		StringInfoData buffer;
+		initStringInfoOfString(&buffer, binary, len);
+
+		cxt->sharedPath = palloc(buffer.len + 1);
+		while(buffer.cursor < buffer.len) {
```
hornn: missing space after while.
Also I think the palloc + loop can be replaced by a copy function `pnstrdup`

src/bin/gpfusion/gpbridgeapi.c
hornn: Actually, shouldn't we parse the URI as part of the parse function we already have? that way we can know that we parsed everything else before we expect to read the delimiter and the dfs_address - similar to the way it's implemented right now, when we parse dfs_address that was appended by hd_work_mgr.

hornn:
Not sure I understood everything, so correct me if I am wrong. 
So right now we have two flows
1. For readable tables we append dfs_address in hd_work_mgr and parse it in pxfuriparser.
2. For writable tables we append dfs_address as part of the URI that is part of the external table context. Then we parse it in gpbridgeapi, but only in the export case (writable).
My question is - shouldn't we consolidate the two solutions and adopt the 2nd flow for both readable and writable?

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -102,7 +102,7 @@ int QueryContextDispatchingSizeMemoryLimit = 100 * 1024; /* KB */
 
 enum QueryContextDispatchingItemType
 {
-    MasterXid, TablespaceLocation, TupleType, EmptyTable, FileSystemCredential
+    MasterXid, TablespaceLocation, TupleType, EmptyTable, FileSystemCredential, Namespace
```
hornn: can you please explain this part? does it mean the namespace becomes part of the dispatched data? if so, could we potentially get it without changing the external table URL?

src/bin/gpfusion/gpbridgeapi.c
kavinderd: Good point on the delimiter, I'll replace it.

I chose not to parse the URI in `parseGPHDUri` because even with this fix we will support two different flows for the readable and writable case. So this is a temporary solution whose effects I wanted to contain only to writable cases.

src/bin/gpfusion/gpbridgeapi.c
kavinderd: Ok, good to know

src/include/access/fileam.h
kavinderd: oops

src/bin/gpupgrade/crc32c.c
kavinderd: They're not. They accidentally got commited. Will remove

src/backend/access/external/fileam.c
kavinderd: Remnants of when I was prototyping the solution. Will remove.

src/backend/access/external/hd_work_mgr.c
kavinderd: Currently the implementation to add the dfs_address through the external framework is not triggered in the readable case. Ideally, yes, the same implementation would apply to both. So we would like to get this fix in, tested, and then work on consolidating them.

src/backend/cdb/cdbquerycontextdispatching.c
kavinderd: So these enums identify portions of the `buffer` in `QueryContextInfo`. When the segment receives the data it checks for the presence of the enums and then rebuilds the data accordingly. 

The issue with dispatching the namespace from master to segment to pxf was two fold:
1. Master needs to serialize the data, segment needs to deserialize it
2. Segement needs to hold the `dfs_address` in a variable until the moment it actually invokes the pxf udf

The reason I added the namespace to the URL was so that I wouldn't need to add a new field to the `PG_FUNCTION_ARGS`.

I agree that adding namespace to URL is a bit of a hack, and this whole implementation is really the shortest path to get the namespace from master to segment so it definitely needs improvement.

src/backend/gpopt/ivy.xml
kavinderd: No, local change needed to get HAWQ to build. Will remove

src/backend/access/external/fileam.c
```diff
@@ -513,7 +515,7 @@ external_getnext(FileScanDesc scan, ScanDirection direction)
  * this function to initialize our various structures and state..
  */
 ExternalInsertDesc
-external_insert_init(Relation rel, int errAosegno)
+external_insert_init(Relation rel, int errAosegno, char *namespace)
```
GodenYao: maybe add comments to explain why we need namespace here.

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -508,17 +531,14 @@ size_t fill_buffer(gphadoop_context* context, char* start, size_t size)
 void add_delegation_token(PxfInputData *inputData)
```
GodenYao: probably should clean this parameter as it's not used in the fuction

GodenYao: can we just pass inputData->gphduri->dfs_address ? instead of the entire object?

kavinderd:
Closing since https://github.com/apache/incubator-hawq/pull/503 is a more complete solution

PR #499, opened 2016-03-24
HAWQ-584. Choosing cluster to core ratio to a multiple of 1GB

jiny2:


jiny2:
Something wrong, I close this pull

PR #502, opened 2016-03-24
HAWQ-584. Choose cluster memory to core ratio a multiple of 1GB

jiny2:
introduced guc hawq_rm_clusterratio_core_to_memorygb_factor with default value 5. This is used to balance resource waste of core and memory, the default value means every 1 core wasted is equal to 5 gb memory wasted. 

Resource manager tries to find the ratio as min as possible to minimize total evaluated resource waste.


linwen:
+1

yaoj2:
LGTM

PR #503, opened 2016-03-25
HAWQ-462. Dispatch dfs_address from master to segment in pxf secure f…

kavinderd:
…ilesystem case

Set global var dfs_address on segment after deserialization

### DIFF
--- a/src/backend/cdb/cdbquerycontextdispatching.c
+++ b/src/backend/cdb/cdbquerycontextdispatching.c
@@ -786,7 +786,7 @@ RebuildNamespace(QueryContextInfo *cxt)
 	{
 		StringInfoData buffer;
 		initStringInfoOfString(&buffer, binary, len);
-		dfs_address = pnstrdup(buffer.data, buffer.len);
+		dfs_address = strdup(buffer.data);
 	} else {
 		elog(ERROR, "Couldn't rebuild Namespace");
 	}
--- a/src/backend/cdb/cdbquerycontextdispatching.c
+++ b/src/backend/cdb/cdbquerycontextdispatching.c
@@ -3016,7 +3016,7 @@ GetResultRelSegFileInfos(Oid relid, List *segnomaps, List *existing_seginfomaps)
 void
 prepareDfsAddressForDispatch(QueryContextInfo* cxt)
 {
-	if (enable_secure_filesystem)
+	if (!enable_secure_filesystem)
 		return;
 	const char *namespace = cxt->sharedPath;
 	int size = strlen(namespace);
--- a/src/backend/cdb/cdbquerycontextdispatching.c
+++ b/src/backend/cdb/cdbquerycontextdispatching.c
@@ -772,7 +772,6 @@ RebuildTupleForRelation(QueryContextInfo *cxt)
 /*
  * Deserialize the Namespace data
  */
-
 static void
 RebuildNamespace(QueryContextInfo *cxt)
 {
@@ -3015,9 +3014,10 @@ GetResultRelSegFileInfos(Oid relid, List *segnomaps, List *existing_seginfomaps)
 }
 
 void
-prepareDfsAddressForDispatch(QueryContextInfo* cxt){
-//	if (enable_secure_filesystem)
-//		return;
+prepareDfsAddressForDispatch(QueryContextInfo* cxt)
+{
+	if (enable_secure_filesystem)
+		return;
 	const char *namespace = cxt->sharedPath;
 	int size = strlen(namespace);
 	StringInfoData buffer;
--- a/src/backend/cdb/cdbquerycontextdispatching.c
+++ b/src/backend/cdb/cdbquerycontextdispatching.c
@@ -102,7 +102,7 @@ int QueryContextDispatchingSizeMemoryLimit = 100 * 1024; /* KB */
 
 enum QueryContextDispatchingItemType
 {
-    MasterXid, TablespaceLocation, TupleType, EmptyTable, FileSystemCredential
+    MasterXid, TablespaceLocation, TupleType, EmptyTable, FileSystemCredential, Namespace
 };
 typedef enum QueryContextDispatchingItemType QueryContextDispatchingItemType;
 
@@ -770,6 +770,31 @@ RebuildTupleForRelation(QueryContextInfo *cxt)
 }
 
 /*
+ * Deserialize the Namespace data
+ */
+
+static void
+RebuildNamespace(QueryContextInfo *cxt)
+{
+
+	int len;
+	char buffer[4], *binary;
+	ReadData(cxt, buffer, sizeof(buffer), TRUE);
+
+	len = (int) ntohl(*(uint32 *) buffer);
+	binary = palloc(len);
+	if(ReadData(cxt, binary, len, TRUE))
+	{
+		StringInfoData buffer;
+		initStringInfoOfString(&buffer, binary, len);
+		dfs_address = pnstrdup(buffer.data, buffer.len);
+	} else {
+		elog(ERROR, "Couldn't rebuild Namespace");
+	}
+	pfree(binary);
+}
+
+/*
  * rebuild execute context
  */
 void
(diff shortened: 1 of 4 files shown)


kavinderd:
@hornn @shivzone  Alternate implementation that handles both readable and writable case by setting a global var on segment after deserialization

I declared the var `dfs_address` in `fd.h` since that is where `enable_filesystem_credential` is declared there as well

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -508,15 +508,12 @@ size_t fill_buffer(gphadoop_context* context, char* start, size_t size)
 void add_delegation_token(PxfInputData *inputData)
 {
 	PxfHdfsTokenData *token = NULL;
-	char* dfs_address = NULL;
 
 	if (!enable_secure_filesystem)
 		return;
```
hornn: perhaps add assert that dfs_address is not null

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -2984,3 +3013,20 @@ GetResultRelSegFileInfos(Oid relid, List *segnomaps, List *existing_seginfomaps)
 
 	return existing_seginfomaps;
 }
+
+void
+prepareDfsAddressForDispatch(QueryContextInfo* cxt){
+//	if (enable_secure_filesystem)
```
hornn: remove comments?

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -770,6 +770,31 @@ RebuildTupleForRelation(QueryContextInfo *cxt)
 }
 
 /*
+ * Deserialize the Namespace data
+ */
+
```
hornn: remove new line

hornn:
I really like this solution - it looks really clean. 
What happens in the non secured case - do we always pass this var regardless?

src/backend/cdb/cdbquerycontextdispatching.c
kavinderd: oh yeah, oops

kavinderd:
In the non-secure case the `dfs_address` is not serialized and dispatched from the master so on the segment the global var will just be `NULL`

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -2984,3 +3012,21 @@ GetResultRelSegFileInfos(Oid relid, List *segnomaps, List *existing_seginfomaps)
 
 	return existing_seginfomaps;
 }
+
+void
+prepareDfsAddressForDispatch(QueryContextInfo* cxt)
+{
+	if (enable_secure_filesystem)
```
shivzone: This should be ! enable_secure_filesystem

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -2984,3 +3012,21 @@ GetResultRelSegFileInfos(Oid relid, List *segnomaps, List *existing_seginfomaps)
 
 	return existing_seginfomaps;
 }
+
+void
+prepareDfsAddressForDispatch(QueryContextInfo* cxt)
```
shivzone: add comment to describe the purpose of this function

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -801,6 +825,9 @@ RebuildQueryContext(QueryContextInfo *cxt, HTAB **currentFilesystemCredentials,
         	RebuildFilesystemCredentials(cxt, currentFilesystemCredentials,
         	        currentFilesystemCredentialsMemoryContext);
         	break;
+        case Namespace:
+            RebuildNamespace(cxt);
```
shivzone: Will this case be invoked in the non secure case where you don't set the namespace

shivzone:
+1

src/backend/cdb/cdbquerycontextdispatching.c
kavinderd: No, it won't. The `Namespace` enum is only added to the buffer in the case of a secure filesystem

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -1747,6 +1774,7 @@ prepareDispatchedCatalogExternalTable(QueryContextInfo *cxt,
 	{
 		Insist(array_size == 1);
 		AddFileSystemCredentialForPxfTable(location);
```
GodenYao: is this still needed? AddFileSystemCredentialForPxfTable as you dispatch it through query context now.

GodenYao:
+1

kavinderd:
Merged into master by @shivzone [here](https://github.com/apache/incubator-hawq/commit/59ebfa7072621117827ae3d9464c971a61919672)

ictmalili:
@shivzone @kavinderd , actually this commit blocks build on MAC. Since the function is referenced before the declaration.
{code}
cdbquerycontextdispatching.c:3026:1: error: conflicting types for 'prepareDfsAddressForDispatch'
prepareDfsAddressForDispatch(QueryContextInfo* cxt)
^
cdbquerycontextdispatching.c:1777:3: note: previous implicit declaration is here
                prepareDfsAddressForDispatch(cxt);
                ^
2 warnings and 1 error generated.
make[3]: *** [cdbquerycontextdispatching.o] Error 1
make[2]: *** [cdb-recursive] Error 2
make[1]: *** [all] Error 2
make: *** [all] Error 2
{code}

kavinderd:
@ictmalili will fix asap

ictmalili:
@kavinderd  I have fixed this. Could you see the pull request? https://github.com/apache/incubator-hawq/pull/554

kavinderd:
@ictmalili Oh ok cool. Thanks for taking care of it

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -770,6 +770,30 @@ RebuildTupleForRelation(QueryContextInfo *cxt)
...
+	char buffer[4], *binary;
+	ReadData(cxt, buffer, sizeof(buffer), TRUE);
+
+	len = (int) ntohl(*(uint32 *) buffer);
+	binary = palloc(len);
+	if(ReadData(cxt, binary, len, TRUE))
+	{
+		StringInfoData buffer;
+		initStringInfoOfString(&buffer, binary, len);
+		dfs_address = strdup(buffer.data);
```
ztao1987: Why not just strdup(binary) to dfs_address?

kavinderd: Good point, no need for `buffer`. I'll update this

src/bin/gpfusion/gpbridgeapi.c
```diff
@@ -536,3 +536,8 @@ void free_token_resources(PxfInputData *inputData)
 
 	pfree(inputData->token);
 }
+
+void free_dfs_address()
+{
+	free(dfs_address);
+}
```
ztao1987: Will free_dfs_address only be called in secure mode? Otherwise it will coredump

kavinderd: Yes only in the secure mode. However it is only dispatched from master in the secure mode as well so it is never allocated in any other case

ztao1987: I checked free_dfs_address is called by gpbridge_export_start and gpbridge_import_start, do you mean this two funtions are only called in the secure mode? If not, you are freeing NULL pointer.

kavinderd: Oh right, good catch. I will add a check in `free_dfs_address` for secure filesystem. Sound good?

ztao1987: Yes, just check NULL condition is enough.

kavinderd: Sure, thanks

ictmalili:
Adding @ztao1987 's comments, LGTM.

PR #509, opened 2016-03-25
HAWQ-591. Explain analyze doesn't enter function cdbexplain_recvExecS…

ictmalili:
…tats after QD finding QE error


ztao1987:
+1

PR #513, opened 2016-03-26
HAWQ-597. Add Travis CI(OSX)

xunzhang:
Add Travis CI support under OSX, help user verify local commit before sending pull request.

More info can be found: https://travis-ci.org/xunzhang/incubator-hawq/builds/118761850

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ compiler:
   - clang
 
 before_install:
-  - brew update
+  - "brew update 2>&1 > /tmp/brew-update.txt || (cat /tmp/brew-update.txt && false)"
 
 install:
   - brew install
@@ -36,6 +36,8 @@ install:
   - sudo pip install
     "http://darcs.idyll.org/~t/projects/figleaf-${FIGLEAF_VERSION}.tar.gz"
   - brew uninstall postgresql
+
+before_script:
   - wget -O "/tmp/libhdfs3" "https://github.com/Pivotal-DataFabric/libhdfs3/archive/apache-rpc-9.zip"
   - unzip -d "/tmp" "/tmp/libhdfs3"
   - cd "/tmp/libhdfs3-apache-rpc-9"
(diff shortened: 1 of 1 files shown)


xunzhang:
@cjcjameson plz review

cjcjameson:
@xunzhang thank you for your pull request and contribution to HAWQ! I would also want to loop in another contributor, Michael Schubert.

schubert:
I saw this warning in the build:

`This log is too long to be displayed. Please reduce the verbosity of your build or download the raw log.`

Due to what looks like Travis's limitation of 10,000 lines displayed for the build log, I would recommend you pipe non-essential information to a file and only print out in an error condition so you minimize the log noise and usually only see essential build output without having to look at the raw log.

you can do this for example with brew update: 

```bash
brew update 2>&1 > /tmp/brew-update.txt || (cat /tmp/brew-update.txt && false)
```

Redirect stderr to stdout and if you get a non-zero exit, print out the logfile of what happened and stop the build. 

I think you would want to do this for everything in the `before_install` and in the `install` with the exception of libhdfs3 due to it being a tightly coupled and relevant dependency. Maybe move libhdfs3 down into the before_script to make that separation clear?

Hopefully with that all of the critical build steps relevant to HAWQ itself will be displayed the vast majority of the time to make it more developer friendly.

xunzhang:
@schubert Thank you for the advice, I modified the `brew update` and move `libhdfs3` into `before_script` into another commit. Since all `install` info take only less than 200 lines and travis will fold these info automatically, I don't modify `install` for the code simplicity. Check out https://travis-ci.org/xunzhang/incubator-hawq/builds/119114627.

schubert-apt:
Yup that looks like a better solution and you've got another 1300 log lines to spare. LGTM.

schubert:
Yup that looks like a better solution and you've got another 1300 log lines to spare. LGTM.

cjcjameson:
+1

xunzhang:
Already merged into master by @radarwave, so I will close.

xunzhang:
Already merged into master by @radarwave, so I will close.

PR #515, opened 2016-03-27
HAWQ-596. change gucs hawq_rm_tolerate_nseg_limit and hawq_rm_rejectrequest_nseg_limit to a percentage value

jiny2:
double type gucs, the value range is [0.0,1.0], the default values of both hawq_rm_tolerate_nseg_limit and hawq_rm_rejectrequest_nseg_limit are changed to 0.25.


yaoj2:
+1

linwen:
+1

PR #517, opened 2016-03-28
HAWQ-560. Set 'output.replace-datanode-on-failure' depend on datanodes number to hdfs-client.xml

radarwave:


huor:
It would be better to utilize xml related utilities instead of re-implement it here again, i.e., append and update property in xml files. +1

yaoj2:
LGTM

PR #521, opened 2016-03-28
HAWQ-587. Set GUC 'default_hash_table_bucket_number' value dynamically

radarwave:
When doing 'hawq init master/cluster', we will check command option '--bucket_number' first, if it was not given by user, then we use the value defined in 'hawq-site.xml'.

If no option given and not defined in 'hawq-site.xml', we will calculate a number out for 'default_hash_table_bucket_number' base on segment number and the up limit value 'hawq_rm_nvseg_perquery_limit'.

### DIFF
--- a/tools/bin/hawq_ctl
+++ b/tools/bin/hawq_ctl
@@ -65,9 +65,9 @@ class HawqInit:
         self.hawq_lc_monetary = opts.hawq_lc_monetary
         self.hawq_lc_numeric = opts.hawq_lc_numeric
         self.hawq_lc_time = opts.hawq_lc_time
-        self.vseg_num_per_node = opts.virtual_seg_num
         self.max_connections = opts.max_connections
         self.shared_buffers = opts.shared_buffers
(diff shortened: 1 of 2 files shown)


radarwave:
@denalex, please help to review this, since it can solve conflict with Ambari.

If Ambari already set 'default_hash_table_bucket_number' in hawq-site.xml, then hawq init will just use that value. 

Ambari also can give option '--default_hash_table_bucket_number' to 'hawq init master/cluster' command to specify the value. 

Thanks.

radarwave:
Some related changes:
1. Do not set GUC 'hawq_rm_nvseg_perquery_perseg_limit' any more since the default value is good enough.
2. Remove option ''--vsegment-number" since we do not need it any more.
3. Add new option "--default_hash_table_bucket_number" to let user define GUC 'default_hash_table_bucket_number'.

denalex:
The logic looks good, great that you reused the calculation code, so we can guarantee the same value is calculated by Ambari and HAWQ.

yaoj2:
LGTM

PR #522, opened 2016-03-28
HAWQ-577. Updated PXF metadata api to stream response

shivzone:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponseFormatter.java
```diff
@@ -46,76 +45,50 @@
...
+
+        if (null == metadataList || metadataList.isEmpty()) {
+            LOG.debug("No metadata");
+            return;
+        }
 
         for(Metadata metadata: metadataList) {
-            result.append(" Metadata for item \"");
+            StringBuilder result = new StringBuilder();
+            result.append("Metadata for item \"");
```
hornn: the quotes are not closed for the case of null metadata.
perhaps move `\"` to be part of line 79.

shivzone: good catch

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/MetadataResource.java
```diff
@@ -41,6 +41,7 @@
 import org.apache.hawq.pxf.api.MetadataFetcher;
 import org.apache.hawq.pxf.api.utilities.InputData;
 import org.apache.hawq.pxf.service.MetadataFetcherFactory;
+import org.apache.hawq.pxf.service.MetadataResponse;
```
hornn: is this a new file?

shivzone: yes, it is. I missed adding it and i've just updated

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponse.java
```diff
@@ -0,0 +1,94 @@
...
+        String prefix = "";
+        for (Metadata metadata : metadataList) {
+            if(metadata == null) {
+                throw new IllegalArgumentException("metadata object is null - cannot serialize");
+            }
+            if ((metadata.getFields() == null) || metadata.getFields().isEmpty()) {
+                throw new IllegalArgumentException("metadata for " + metadata.getItem() + " contains no fields - cannot serialize");
+            }
+            StringBuilder result = new StringBuilder();
+            /* metaData and userData are automatically converted to Base64 */
```
hornn: probably remove comment :)

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponse.java
```diff
@@ -0,0 +1,94 @@
...
+        DataOutputStream dos = new DataOutputStream(output);
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.setSerializationInclusion(Inclusion.NON_EMPTY); // ignore empty fields
+
+        if(metadataList == null || metadataList.isEmpty()) {
+            dos.write(METADATA_DEFAULT_RESPONSE.getBytes());
+            return;
+        }
+
+        dos.write("{\"PXFMetadata\":[".getBytes());
```
hornn: perhaps make PXFMetadata into a const, and use it here and as part of of METADATA_DEFAULT_RESPONSE.

hornn:
+1. looks good

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponse.java
shivzone: I didn't want to do that. The default response can possibly change tomorrow and wanted to decouple that from a valid response string.

hornn: You mean that the default might not contain the header "PXFMetadata"?
In that case I would do something like
```
final String METADATA_HEADER_RESPONSE = "PXFMetadata";
final String METADATA_DEFAULT_RESPONSE = "{\"" + METADATA_HEADER_RESPONSE + "\":[]}"
```

shivzone: What i meant was I wouldn't want to have any coorelation between metadata_default_response and the header string used in the actual response. I wouldn't mind introducing a separate constant for the string used in the acutal response but wouldn't use that as part of the default response

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponse.java
```diff
@@ -0,0 +1,94 @@
...
+        this.metadataList = metadataList;
+    }
+
+    /**
+     * Serializes the metadata list in JSON, To be used as the result string for HAWQ.
+     */
+    @Override
+    public void write(OutputStream output) throws IOException,
+            WebApplicationException {
+        DataOutputStream dos = new DataOutputStream(output);
```
sansanichfb: Should we close resource in the end?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponseFormatter.java
```diff
@@ -46,76 +45,49 @@
...
+            return;
+        }
 
         for(Metadata metadata: metadataList) {
-            result.append(" Metadata for item \"");
+            StringBuilder result = new StringBuilder();
 
             if (metadata == null) {
-                return "No metadata";
+                result.append("None");
```
sansanichfb: Literal "None" was used couple times, might me a constant.

sansanichfb:
LGTM

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/MetadataResponseFormatter.java
shivzone: This is just used for logging and not for the response. Defining constants just for log message seems an overkill

PR #524, opened 2016-03-29
 HAWQ-599. Fixed coverity issues.

sansanichfb:
We were carrying character buffer size of 64k over stack which was unnecessary because it was used in method pxfutils.process_request only, moved it to local variable.


shivzone:
lgtm +1

PR #526, opened 2016-03-29
HAWQ-568. Check validation the QD to QE connection by checking whethe…

ictmalili:
…r the recv character is 'E' which means there's error on the other side


ztao1987:
+1

PR #532, opened 2016-03-30
HAWQ-592. QD shouldn't try to reconnect to segment when meeting inval…

ictmalili:
…id QE for queries which only need to dispatch to all cached QEs


ztao1987:
+1

PR #534, opened 2016-03-30
HAWQ-608. Libyarn hotrefactor

xunzhang:
Solve https://issues.apache.org/jira/browse/HAWQ-608.


xunzhang:
@linwen @wangzw plz review. Local CI passed at https://travis-ci.org/xunzhang/incubator-hawq/builds/119545927.

wangzw:
The fix is ok to me. +1
ps. It is a good C++ practice to not include ```using namespace ...``` or ``` using ...``` in header file since header file will be included into other source code without any protection. It may cause conflicts. So I believe libyarn code needs more refactor.

xunzhang:
Of course, this pr only contains a hot refactor for build. I will refactor it later on since there are so many mistakes and wrong ways to use STL in `libyarn`.

xunzhang:
@linwen , thanks for merging, I will close this pull request.

PR #535, opened 2016-03-31
remove gp_enable_column_oriented_table and change default seg_max_connections default value

yaoj2:
1. Change seg_max_connections from 1280 to 3000
2. since gp_enable_column_oriented_table is not used, remove the guc value


linwen:
+1

radarwave:
Since the local build test passed.
+1

PR #538, opened 2016-03-31
HAWQ-573. Refix the bug as last fix has imported more warning messages.

realdawn:
As we changed the allocate resource logic, which means we add resource into GlobalQueryResources before we actually allocate the resource, it has a possibility that we allocate failed and want to return it. Then it return failed and report warning messages, which should not be reported. So this fix is to judge and mark if it should report the message.


zhangh43:
LGTM

jiny2:
+1

PR #541, opened 2016-03-31
HAWQ-613. Cannot cancel query when PL UDF has infinity loops

liming01:
Notes:
(1) Here just change some UDF calling may introduce hung problem, I don't go through all code lines PL UDF maybe called.
(2)  in client_read_ended(), i just set QueryCancelPending to saveQueryCancelPending, instead of set it to false, it means it only process the cancel request during client read period.
(3) InterruptWhenCallingPLUDF need to be reset during error report and process exists.

Please review, thanks.

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -265,6 +265,9 @@ static bool renice_current_process(int nice_level);
 static int getSlaveHostNumber(FILE *fp);
 static bool CheckSlaveFile();
 
+/*saved interrupt global variable for client_read_xxx functions*/
+static bool saveImmediateInterruptOK = false;
+static bool saveQueryCancelPending = false;
 
 /*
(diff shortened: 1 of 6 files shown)


src/backend/tcop/postgres.c
```diff
@@ -608,9 +611,11 @@ prepare_for_client_read(void)
 		EnableClientWaitTimeoutInterrupt();
 
 		/* Allow "die" interrupt to be processed while waiting */
+		saveImmediateInterruptOK = ImmediateInterruptOK;
```
changleicn: following convention: saveImmediateInterruptOK => SaveImmediateInterruptOK

changleicn:
LGTM except for the variable convention change.

huor:
+1 It is better to use SavedXXX instead of SaveXXX

PR #544, opened 2016-04-01
HAWQ-616 Mininum resource needed by copy random table from a file sho…

zhangh43:
…uld be one.

Now, we use hawq_rm_nvseg_for_copy_from_perquery as the min and max vseg number for copy random table from a file, For max vseg num is OK, but min value should be one in case there is not enough resource


ictmalili:
+1

PR #546, opened 2016-04-01
HAWQ-617 Add option to skip syncing configuration files

bhuvnesh2703:
Currently Activate standby operations uses 'hawq config' to update hawq-site.xml. However, in case any of the HAWQ host is down, the operation fails while syncing up the configuration. In case of big clusters, there can be few segment hosts which are down and that will prohibit activating standby.
Thus, introduced a new flag in hawq config to ignore bad hosts and continue syncing files on hosts on which test ssh is successful

### DIFF
--- a/tools/bin/gppylib/util/ssh_utils.py
+++ b/tools/bin/gppylib/util/ssh_utils.py
@@ -160,6 +160,29 @@ class HostList():
         
         return self.list
 
+    def removeBadHosts(self):
+        ''' Update list of host to include only the host on which SSH was successful'''
+
+        pool = WorkerPool()
+
+        for h in self.list:
+            cmd = Echo('ssh test', '', ctxt=REMOTE, remoteHost=h)
+            pool.addCommand(cmd)
+
+        pool.join()
+        pool.haltWork()
+
+        bad_hosts = []
+        working_hosts = []
+        for cmd in pool.getCompletedItems():
+            if not cmd.get_results().wasSuccessful():
+                bad_hosts.append(cmd.remoteHost)
+            else:
+                working_hosts.append(cmd.remoteHost)
+
+        self.list = working_hosts[:]
+        return bad_hosts
+
 # Session is a command session, derived from a base class cmd.Cmd
 class Session(cmd.Cmd):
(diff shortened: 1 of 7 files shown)


.gitignore
```diff
@@ -1,54 +1 @@
-# Object files
```
radarwave: Why changing to .gitignore file?

radarwave:
How about pass this option to hawq stop/start cluster?

.gitignore
bhuvnesh2703: will delete it, not needed.

bhuvnesh2703:
Thank you @radarwave for the feedback
Added the option to start all segments, but not to stop, because stop already has a check which verifies if there is an existing postgres process before attempt to stop.

Let me know if there are any comments which needs to be fixed.

.gitignore
radarwave: I see you removed all the lines in .gitignore, I think we should not modify and change it here.

radarwave:
@bhuvnesh2703 Now 'hawq stop' works just like '--ignore-bad-hosts' always enabled. But 'hawq stop' should only check if hawq process is running on the node, it should fail if the node is not available. We will fix that in another jira.

It's ok to apply the 'option' to 'hawq stop' next time when we fixing above hawq stop check issues. But if it convenience, please also apply it to 'hawq stop' so we can keep the option works for both start and stop. Thanks.

radarwave:
BTW, it's better to squash the commits since there is one commit have merged things and un-format name. 

See the commit 'c5c7d8fc0757822d21e18958e03d1e536fc27a56':
"Merge branch 'master' into HAWQ-617"

bhuvnesh2703:
@radarwave Thanks for the review, will fix the stop option after this pull request, may be next week. Since this is a blocker for Ambari, i will prefer merge this and continue with Ambari testing, and do the stop option thing next week. Is that okay for you ? If so, please +1 and i will then commit. I will squash the commits before merging.

radarwave:
Sure, we will track the stop opition in https://issues.apache.org/jira/browse/HAWQ-626

+1

PR #549, opened 2016-04-01
HAWQ-621. Fixing tuple descriptor leak in split_rows during an ALTER TABLE SPLIT PARTITION. 

foyzur:
During split_rows, we may create new tuple table slot inside reconstructMatchingTupleSlot() if the descriptor of the new target relation does not match the source relation. We then cache this slot inside the target relation's ri_resultSlot. This process pins the tuple descriptor of the new target relation. However, at the end of split_rows, when we are finished with this "cached" tuple table slot, we don't free them. This results in a WARNING that we leaked a tuple descriptor.

Solution:
-----------

Following the call path of split_rows, the method split_rows() is only called from 

```
split_rows(Relation, Relation, Relation) : void
	ATPExecPartSplit(Relation *, AlterPartitionCmd *) : void
		ATExecCmd(List * *, AlteredTableInfo *, Relation *, AlterTableCmd *) : void
			ATPExecPartAlter(List * *, AlteredTableInfo *, Relation, AlterPartitionCmd *) : void
			ATRewriteCatalogs(List * *) : void
```

By the time we call split_rows, we know precisely the two relation where we will be inserting into. We create ResultRelInfo for both of these relations (the two target relations) at the beginning of split_rows(), called rria, and rrib. Both of these ResultRelInfo only have lifespan of the containing method split_rows(). We use the ri_resultSlot of these two rria and rrib to cache the TupleTableSlot inside reconstructMatchingTupleSlot() where we create additional TupleTableSlot using the tuple descriptors of the two target relations, therefore pinning these two target relation's TupleDesc. As the rria and rrib only live during split_rows() and only used for all the tuples in ONE CALL of split_rows(), we can easily drop the additional TupleTableSlot ri_resultSlot of both rria and rrib, and that will also unpin underlying TupleDesc.

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -15601,6 +15601,23 @@ split_rows(Relation intoa, Relation intob, Relation temprel, List *splits, int s
 	ExecDropSingleTupleTableSlot(rria->ri_partSlot);
 	ExecDropSingleTupleTableSlot(rrib->ri_partSlot);
 
+	/*
+	 * We may have created "cached" version of our target result tuple table slot
+	 * inside reconstructMatchingTupleSlot. Drop any such slots.
+	 */
+	if (NULL != rria->ri_resultSlot)
+	{
+		Assert(NULL != rria->ri_resultSlot->tts_tupleDescriptor);
+		ExecDropSingleTupleTableSlot(rria->ri_resultSlot);
+		rria->ri_resultSlot = NULL;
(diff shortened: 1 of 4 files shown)


liming01:
Looks fine. +1.

liming01:
@foyzur , merged into master branch. please close this PR. Thank.

changleicn:
@foyzur could you close this PR. thank!

PR #551, opened 2016-04-04
HAWQ-615. Handle incomptible tables with getMetadata PXF API

shivzone:
Incompatible tables will be ignored when getMetadata is invoked with a pattern that matches > 1 table

### DIFF
--- a/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
+++ b/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
@@ -52,7 +52,7 @@ public class HiveMetadataFetcher extends MetadataFetcher {
 
     /**
      * Fetches metadata of hive tables corresponding to the given pattern
-     * For patterns matching more than one table, the tables are skipped.
+     * For patterns matching more than one table, the unsupported tables are skipped.
      * If the pattern correspond to exactly one table, throws an exception if
      * the table type is not supported or contains unsupported field types.
      * Supported HCatalog types: TINYINT,
--- a/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
+++ b/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
@@ -50,9 +50,21 @@ public class HiveMetadataFetcher extends MetadataFetcher {
         client = HiveUtilities.initHiveClient();
     }
 
+    /**
+     * Fetches metadata of hive tables corresponding to the given pattern
+     * For patterns matching more than one table, the tables are skipped.
+     * If the pattern correspond to exactly one table, throws an exception if
+     * the table type is not supported or contains unsupported field types.
+     * Supported HCatalog types: TINYINT,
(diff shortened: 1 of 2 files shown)


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
```diff
@@ -50,9 +50,21 @@ public HiveMetadataFetcher(InputData md) {
         client = HiveUtilities.initHiveClient();
     }
 
+    /**
+     * Fetches metadata of hive tables corresponding to the given pattern
+     * For patterns matching more than one table, the tables are skipped.
```
sansanichfb: Maybe For patterns matching more than one table, the unsupported tables are skipped.?

pxf/pxf-hive/src/test/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcherTest.java
```diff
@@ -154,6 +155,115 @@ public void getTableMetadata() throws Exception {
...
+
+        // Prepare for tables
+        List<FieldSchema> fields = new ArrayList<FieldSchema>();
+        fields.add(new FieldSchema("field1", "string", null));
+        fields.add(new FieldSchema("field2", "int", null));
+        StorageDescriptor sd = new StorageDescriptor();
+        sd.setCols(fields);
+
+        // Mock hive tables returned from hive client
+        for(int index=1;index<=2;index++) {
```
sansanichfb: Add comment why starting from 1 not from 0?

sansanichfb:
LGTM other than small cosmetic comments.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
shivzone: will update

pxf/pxf-hive/src/test/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcherTest.java
```diff
@@ -154,6 +155,115 @@ public void getTableMetadata() throws Exception {
...
+            String tableName = tablenamebase + index;
+            tableNames.add(tableName);;
```
hornn: two ;

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveMetadataFetcher.java
```diff
@@ -62,11 +74,24 @@ public HiveMetadataFetcher(InputData md) {
 
         List<Metadata> metadataList = new ArrayList<Metadata>();
 
+        if(tblsDesc.size() > 1) {
```
hornn: What if there are 2 tables but both are not supported? in that case the function will return 0 tables. Is that an acceptable scenario? if so, perhaps add a test to cover it?

Did you consider changing the condition to have at least one 'good' table - check in the end of the loop if there metadataList is not empty. If it is, throw an exception.

shivzone:
@hornn the reason we went with returning an empty response was we wanted to be consistent with the way we treat the metadata api based on the given pattern. 
If the pattern has wildcards in it, we will never throw an exception in any case based on table incompatibilities. If the pattern corresponds to a specific table, we will throw the appropriate exception.

PR #557, opened 2016-04-05
HAWQ-605. Some segment capacity changes are not logged out and when segment goes to up status, the capacity is not adjusted

jiny2:


yaoj2:
+1

linwen:
+1

PR #558, opened 2016-04-06
HAWQ-629. Insert into table select generate_series free resource too …

zhangh43:
…early.
the query"insert into table select generate_series" may run a long time, but resource is freed at the begining of the query.
This behaviour may lead to RM considers the resouece is unlimited


wengyanqing:
+1

PR #561, opened 2016-04-06
HAWQ-632. HAWQ remove offline standby fails

radarwave:
If standby host is down, init standby -r commands fails to cleanup the directories.
We should ignore cleanup the standby node if it's down.


yaoj2:
LGTM

linwen:
+1

PR #563, opened 2016-04-07
HAWQ-629. Insert into table select generate_series free resource too …

zhangh43:
…early.

If exception happends, we also need to free the resource in catch.


wengyanqing:
+1

PR #564, opened 2016-04-07
HAWQ-635. QE process does not exit in libhdfs

liming01:
1) The problem is caused by the wrong refs in RpcChannelImpl class, so it run a dead loop when process exiting. I suspect it is caused by the exception thrown by RpcClientImpl::getChannel(), which already addRef(), but doesn't call close() when exception occurs.

2) This problem cannot be fixed by enable SIGNAL handle before calling libhdfs functions, because the clean up process is already called when dead loop occurs.

### DIFF
--- a/depends/libhdfs3/src/rpc/RpcClient.cpp
+++ b/depends/libhdfs3/src/rpc/RpcClient.cpp
@@ -155,8 +155,6 @@ RpcChannel & RpcClientImpl::getChannel(const RpcAuth & auth,
             allChannels[key] = rc;
         }
 
-        rc->addRef();
-
         if (!cleaning) {
             cleaning = true;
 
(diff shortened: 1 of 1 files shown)


liming01:
@wangzw , could you help me to review it?  Thanks.

wangzw:
Sure, will do.

wangzw:
If the issue is caused by the exception thrown by RpcClientImpl::getChannel(). The only chance is that the system run out of memory or fail to create thread.  In either case, you will see error message "RpcClient failed to create a channel to" in HAWQ log. 

Please refer the code in RpcClientImpl::getChannel()


    try {
       ...

        rc->addRef();

        if (!cleaning) {
            cleaning = true;

            if (cleaner.joinable()) {
                cleaner.join();
            }

            CREATE_THREAD(cleaner, bind(&RpcClientImpl::clean, this));
        }
    } catch ...



Isn't it much easier to fix this issue by moving ```rc->addRef();``` to the end of the ```try``` block?

liming01:
@wangzw , yes, I agree with you. Thanks.

xunzhang:
@wangzw this fix seems right.
But in my sight, I think we should limit the usage of `try ... catch` for the consideration of performance. Specially, catch small lines of codes is better. Another advice is that we should try to use `try catch` clause in library code in a flat scope as possible as we could.

PR #565, opened 2016-04-07
HAWQ-637. Fix RM process error if property is missing or invalid in h…

linwen:
Please review.


jiny2:
LGTM +1

huor:
@jiny2:
1. For critical parameters that might make RM unable to start if incorrect value is specified, we need to error out while loading them. This makes user be aware of critical configuration issues as early as possible (at hawq start stage, etc).
2. While for parameters that its correctness is not so obvious, we can use the fix in this change to postpone it to runtime.

PR #566, opened 2016-04-07
HAWQ-636. hawq check include /etc/security/limits.d

radarwave:
Tested config files in '/etc/security/limits.d', it works well with different config file and different user if defined user domain.


huor:
+1

yaoj2:
+1

PR #568, opened 2016-04-08
HAWQ-634. Let resource manager cancel waiting and return allocated resource when unregister connection rpc is called

jiny2:


linwen:
LGTM

yaoj2:
LGTM

PR #569, opened 2016-04-08
HAWQ-642. Make sure the number of records generated in AppendOnlyWrit…

ictmalili:
…er equal to the number of QEs in function SetSegnoForWrite when we do insert/copy


zhangh43:
+1

PR #585, opened 2016-04-10
HAWQ-655. Allocate resource for PerformSharedStorageOpTasks bug.

zhangh43:
Allocate resource for PerformSharedStorageOpTasks pass the wrong order of parameters: should be(.., maxvseg,minvseg,...) while now is (.., minvseg,maxvseg,...)


wengyanqing:
+1

PR #595, opened 2016-04-11
HAWQ-628. Return -1 instead of error.

sansanichfb:
As for now Postgres' verison of psql fails on \l+ command because of error during call of pg_catalog.pg_database_size(6120) stored procedure. This fix changes error to -1 value so \l+ works on Postgres' psql as well.

### DIFF
--- a/src/backend/utils/adt/dbsize.c
+++ b/src/backend/utils/adt/dbsize.c
@@ -179,10 +179,9 @@ calculate_database_size(Oid dbOid)
 
 	Assert(Gp_role != GP_ROLE_EXECUTE);
 
+    /* Do not error out because it would break some existing client calls*/
 	if (dbOid == HcatalogDbOid)
-		ereport(ERROR,
-			(ERRCODE_UNDEFINED_DATABASE,
-			errmsg("database hcatalog (OID 6120) is reserved")));
+		return -1;
 
 	/* User must have connect privilege for target database */
 	aclresult = pg_database_aclcheck(dbOid, GetUserId(), ACL_CONNECT);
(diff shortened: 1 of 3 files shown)


kavinderd:
LGTM

src/backend/utils/adt/dbsize.c
```diff
@@ -180,9 +180,7 @@ calculate_database_size(Oid dbOid)
 	Assert(Gp_role != GP_ROLE_EXECUTE);
 
 	if (dbOid == HcatalogDbOid)
-		ereport(ERROR,
-			(ERRCODE_UNDEFINED_DATABASE,
-			errmsg("database hcatalog (OID 6120) is reserved")));
+		return -1;
```
GodenYao: can you add comments here to explain why we return -1 instead of erroring out

src/test/regress/output/hcatalog_lookup.source
```diff
@@ -277,9 +277,17 @@ alter table test_schema.p exchange partition p1 with table hcatalog.test_schema.
 ERROR:  reference to hcatalog table "hcatalog.test_schema.r" is not allowed in this context
 -- negative test: cannot run pg_database_size on hcatalog
```
GodenYao: please update comments here.

src/backend/utils/adt/dbsize.c
shivzone: Should we return -1 ? For other non existent tables we return ERROR:  database <dbname> does not exist.
Shouldn't we return an actual error instead of -1 ?

GodenYao: This is to solve incompatibility with psql clients (not hawq client). In hawq client, we hide hcatalog from being displayed in \l or \l+. But with other psql clients, if we error out in dbsize, \l and \l+ will also error out and show nothing.

A compromise solution, which we implemented here is to return hcatalog dbsize as -1. Based on my investigation and interview with some users they think -1 is better than 0, as it's not a valid size info, so users who may rely on this value can adjust their logic to exclude hcatalog as necessary.

shivzone:
+1

src/backend/utils/adt/dbsize.c
sansanichfb: Added comment to justify.

PR #596, opened 2016-04-12
HAWQ-662. Fix memory leak when mppExecutorCleanup throws error.

zhangh43:
if mppExecutorCleanup throws error, resource may not be freed


wengyanqing:
+1

PR #602, opened 2016-04-13
HAWQ-669. Fix interconnect guc gp_interconnect_transmit_timeout type …

wengyanqing:
Interconnect guc gp_interconnect_transmit_timeout type is int, when compute it by us, integer is overflowed.


zhangh43:
+1

xunzhang:
There might be some problems with this modification: you add a `int64` type of guc value into a `config_int` option list. @wengyanqing

PR #604, opened 2016-04-13
HAWQ-644. Account for '\0' when dispatching namespace length

kavinderd:


src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -3031,7 +3031,7 @@ prepareDfsAddressForDispatch(QueryContextInfo* cxt)
 	if (!enable_secure_filesystem)
 		return;
 	const char *namespace = cxt->sharedPath;
-	int size = strlen(namespace);
+	int size = strlen(namespace) + 1;
```
shivzone: please add a comment mentioning about null terminator character

shivzone:
+1

sansanichfb:
Any regression tests needed to cover edge case?

kavinderd:
@sansanichfb We should add some, but need to run in a secure environment.

ztao1987:
+1

ictmalili:
LGTM. +1

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -779,21 +779,24 @@ static void
...
 	ReadData(cxt, buffer, sizeof(buffer), TRUE);
 
 	len = (int) ntohl(*(uint32 *) buffer);
 	binary = palloc(len);
 	if(ReadData(cxt, binary, len, TRUE))
 	{
-		StringInfoData buffer;
-		initStringInfoOfString(&buffer, binary, len);
-		dfs_address = strdup(buffer.data);
+		dfs_address = pstrdup(binary);
```
shivzone: identation needs to be 4 spaces

src/backend/cdb/cdbquerycontextdispatching.c
```diff
@@ -779,21 +779,24 @@ static void
...
 	} else {
```
shivzone: convention in C is to have 
else 
{
     Multi line stmts
}

or 
else
    Single line sttm

shivzone:
+1 on the switch to use messagecontext instead of global variable to store dfs_address

kavinderd:
@ictmalili @ztao1987 Please review use of MessageContext for `dfs_address` in `RebuildNamespace`

src/backend/cdb/cdbquerycontextdispatching.c
ztao1987: pstrdup is a kind of malloc, not covered by memorycontext. So there is memory leak in your code. Why not just use dfs_address directly instead of binary? Also even memorycontext used, the best practice is to pfree the memory explictly. Memorycontext is the last barrier which protects the memory.

kavinderd: In the definition of `pstrdup` it states 

```
/*
 * MemoryContextStrdup
 *		Like strdup(), but allocate from the specified context
 */
```

I chose to use the intermediary `binary` so that if there is an error `dfs_address` is not set with any data, but your point for not using `binary` at all is valid.

We are aware that we should `pfree` even with MemoryContext, but we need the `dfs_address` for the lifetime of each query so this is the simplest & safest means of doing so with the time given

ztao1987: ok, I thought you used strdup, missed you have changed to pstrdup, this is covered by memorycontext

changleicn:
@kavinderd the code is in, can you close this PR? thanks!

kavinderd:
Merged

PR #606, opened 2016-04-14
HAWQ-672. Add python module pygresql back into hawq workspace

paul-guo-:
In this patch, all of the new added files come from the workspace after reverting HAWQ-271. Other changes mostly follow the original patch for HAWQ-271.


tools/bin/generate-greenplum-path.sh
```diff
@@ -90,16 +90,9 @@ EOF
...
 
 #setup PYTHONPATH
-# OSX does NOT need pygresql/ path
-if [ "${PLAT}" = "Darwin" ] ; then
-cat <<EOF
-PYTHONPATH=\$GPHOME/lib/python:\$PYTHONPATH
-EOF
-else
 cat <<EOF
 PYTHONPATH=\$GPHOME/lib/python:\$GPHOME/lib/python/pygresql:\$PYTHONPATH
```
radarwave: Please check if we need '$GPHOME/lib/python/pygresql' in $PYTHONPATH ?

tools/bin/ext/__init__.py
```diff
@@ -0,0 +1,306 @@
...
+
+from tokens import *
+from events import *
+from nodes import *
+
+from loader import *
+from dumper import *
+
+try:
+    from cyaml import *
```
radarwave: We would better remove the codes not related to pygresql in this '__init__.py' file.

radarwave:
Looks good to me. +1

yaoj2:
＋1

PR #609, opened 2016-04-14
HAWQ-673. Unify out put of explain analyze.

zhangh43:
output of explain analyze is different between one QE and multiple QE, we should unify the output format.


wengyanqing:
+1

PR #614, opened 2016-04-14
HAWQ-678. Resource manager should close connection with QD when QD is cancelled and try to return resource to clean up all registered resource context

jiny2:
When one FD is waiting for resource from resource manager, QD is canclled, QD sends return resource from another FD, in this case old one should be disconnected at once to ensure the response is sent to QD through new FD.


yaoj2:
＋1

zhangh43:
+1

PR #618, opened 2016-04-15
HAWQ-682 Include ignore-bad-hosts option in hawq init master command …

bhuvnesh2703:
…to avoid syncup failures during setting default bucket number


radarwave:
Looks good to me. +1

PR #620, opened 2016-04-18
HAWQ-685. Pass wrong pointer to transform connection track instance status when canceling a waiting query resource request

jiny2:


ictmalili:
+1

liming01:
+1

PR #621, opened 2016-04-18
HAWQ-681. Removed hcatalog_enable GUC.

sansanichfb:


kavinderd:
LGTM

src/test/regress/input/hcatalog_lookup.source
```diff
@@ -2,16 +2,9 @@
 -- test hcatalog lookup
 -- --------------------------------------
 
--- Negative test with GUC disabled
-SET hcatalog_enable = false;
-SELECT * from hcatalog.db.t;
-
 SELECT * FROM pxf_get_item_fields('Hive', '*');
```
shivzone: for each of the higher level use case. please add a one line comment on the type of test

shivzone:
+1

src/test/regress/input/hcatalog_lookup.source
sansanichfb: @shivzone good to have, but not related to this JIRA

PR #622, opened 2016-04-19
HAWQ-668. hawq check should be able to check yarn settings

radarwave:


linwen:
+1

tools/bin/gpcheck
```diff
@@ -567,7 +683,7 @@ def testHAWQ(host):
             return
 
         # check HAWQ master's memory size
-        expected_vmemory_size = 1024
+        expected_vmemory_size = 8192
```
denalex: do we want to check exact value, or minimal value ? in other words, should it be guc_vmemsize_master < expected_vmemory_size rather than guc_vmemsize_master != expected_vmemory_size in the statement below and other similar comparisons ?

src/backend/utils/misc/etc/gpcheck.cnf
```diff
@@ -54,5 +54,60 @@ dfs.namenode.handler.count = 60
...
+dfs.block.access.token.enable = FALSE
+
+[hdfs.ha]
+hdfs.ha = none
+dfs.block.access.token.enable = FALSE
+
+[hdfs.kerberos]
+hdfs.kerberos = none
+dfs.block.access.token.enable = TRUE
+dfs.datanode.data.dir.perm = 750
```
ztao1987: For non-kerberos, it should be 700

tools/bin/gpcheck
radarwave: If user set the value too big, like set it as '81920' by mistake, 'hawq check' won't give a warning since it's bigger than the expect values.

So I think we can keep current way. User will decide wether to fix a hawq check warning base on the document and their environment. 

If any new requirement, we can change it in another jira. Thanks.

src/backend/utils/misc/etc/gpcheck.cnf
radarwave: Since 700 is default values and user might have something like '755', we do not check for non-kerberos mode.

yaoj2:
＋1

PR #623, opened 2016-04-19
HAWQ-687. The segment resource increase pending counter is not adjusted correctly when failing to increase resource quota

jiny2:


liming01:
+1.

linwen:
+1

PR #624, opened 2016-04-19
HAWQ-668. Fix hawq check run without HADOOP_HOME

radarwave:
1. Fix hawq check runs without HADOOP_HOME

2. Update hawq check helps with new options.


huor:
+1

PR #625, opened 2016-04-19
HAWQ-683. Updates required for Javadocs api generation

shivzone:
Change for javadoc prior to hosting on hawq apache site.
Package info sites will be updated to add more details in the future.

### DIFF
--- a/pxf/README.md
+++ b/pxf/README.md
@@ -20,7 +20,7 @@ Package Contents
 ================
 
 PXF is distributed as a set of RPMs -
-git dif
+
     pxf/
         └── rpm
             ├── pxf-service-$version.noarch.rpm
--- a/pxf/README.md
+++ b/pxf/README.md
@@ -20,7 +20,7 @@ Package Contents
 ================
 
 PXF is distributed as a set of RPMs -
-
+git dif
     pxf/
         └── rpm
             ├── pxf-service-$version.noarch.rpm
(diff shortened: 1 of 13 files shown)


pxf/README.md
```diff
@@ -20,7 +20,7 @@ Package Contents
 ================
 
 PXF is distributed as a set of RPMs -
-
+git dif
```
hornn: dif?

shivzone: my bad. removed it

GodenYao:
+1

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/AvroResolver.java
```diff
@@ -152,7 +152,7 @@ boolean isAvroFile() {
      * The record can arrive from one out of two different sources: a sequence
      * file or an AVRO file. If it comes from an AVRO file, then it was already
      * obtained as a {@link GenericRecord} when when it was fetched from the
```
kavinderd: Not related to your change but typo: "when when" should just be "when"

kavinderd:
Looks good

sansanichfb:
Might be useful to have @version tag in each file, but do not add it manually to each class, but just substituting it on some build phase, reading current version from graddle properties, the similar way we do for PXF api version.

PR #635, opened 2016-05-03
HAWQ-713 Make lc_numeric guc to have GUC_GPDB_ADDOPT

karthijrk:
Without GUC_GPDB_ADOPT, the values for guc is not going to be dispatched
to the QE processes.

This fix is ported from [GPDB commit ](https://github.com/greenplum-db/gpdb/commit/d5f3d6fa94b78115424b1e3703cf9913e54507dd).

I am not able to add ICG test since we don't have guc `gp_vmem_idle_resource_timeout`. I attached the test in Jira and that file can be added once the guc is available


karthijrk:
I am closing this PR without merging it. The test I had in GPDB cause failure in some machine due to missing locale. I will fix the test and open the PR again for HAWQ.

PR #636, opened 2016-05-04
HAWQ-716. Support code coverage build and report in hawq Makefile

huor:


paul-guo-:
1) You might want to avoid using the absolute path of lcov, etc. (users could install them anywhere).
2)  Does the code check whether lcov packages have been installed at the configure stage?

huor:
@paul-guo-, please find my comments below:
1. it now uses lcov, gcov, genhtml directory instead of using the absolute path, assuming they are system search path
2. lcov, gcov, genhtml are checked in configure already

paul-guo-:
+1

src/bin/gpfdist/Makefile
```diff
@@ -22,7 +22,8 @@ include $(top_builddir)/src/Makefile.global
 
 LIBS := -levent -lyaml -lz -lbz2 -lssl -lcrypto
 ifeq ($(BUILD_TYPE), gcov)
-LIBS := -lgcov $(LIBS)
+LIBS   := -lgcov $(LIBS)
```
xunzhang: indent

xunzhang:
LGTM

src/bin/gpfdist/Makefile
huor: The indent of LIBS is correct since it is consistent with CFLAGS in next line.

PR #637, opened 2016-05-05
HAWQ-717: Use password as fed by the user

bhuvnesh2703:
Use password as fed by the user

### DIFF
--- a/tools/bin/hawq
+++ b/tools/bin/hawq
@@ -18,6 +18,7 @@
 
 try:
     import os
+    import json
     import sys
     import subprocess
     from hawqpylib.HAWQ_HELP import *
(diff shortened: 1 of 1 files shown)


radarwave:
Looks good to me. +1

tools/bin/hawq
```diff
@@ -82,7 +83,13 @@ def main():
     if len(sys.argv) > 2:
         hawq_command = sys.argv[1]
         second_arg = sys.argv[2]
-        sub_args = " ".join(sys.argv[2:])
+        sub_args_list = sys.argv[2:]
+        # Password can have special characters like semicolon (;), quotes(", ') etc, convert input password to a string
+        if hawq_command == 'ssh-exkeys' and '-p' in sub_args_list:
+            password_index = sub_args_list.index('-p') + 1
+            if len(sub_args_list) > password_index:
+              sub_arg_list[password_index] = json.dumps(sub_args_list[password_index])
```
denalex: a bit cautious about using json here. Tried this from python shell:
>>> json.dumps('foo zzz')
'"foo zzz"'
So it seems the call preserved single quotes but also added double quotes inside the value, which would essentially change the password, no ?

bhuvnesh2703: This is because we are printing that on the python console.
So essentially, to json.dumps a string is passed which foo zzz
json.dumps adds double quotes to it = "foo zzz"
since on the python console the result is printed, to show that its a string python console adds single quotes to it.

if you do the same using the script you will just get "foo zzz"
```
cat t1.py
[gpadmin@c6401 ~]$ cat t1.py
import json
print json.dumps('foo zzz')

[gpadmin@c6401 ~]$ python t1.py
"foo zzz"
[gpadmin@c6401 ~]$ python
Python 2.6.2 (r262:71600, May 12 2009, 15:34:31)
[GCC 4.1.1] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import json
>>> json.dumps('foo zzz')
'"foo zzz"'
>>> quit()
```

PR #638, opened 2016-05-05
HAWQ-711. Integrate libhdfs3 and libyarn makefile into hawq

paul-guo-:
This is based on Liming's work, i.e. commit 1b0e6e2d728b5f4f9da09ecc19dc32c87a9fdc81


huor:
It would be good to support "make unittest"  for hawq and its dependencies as we have "make unittest" for libyarn, libhdfs3, pxf and "make unittest-check" for contrib/hawq-hadoop.

configure
```diff
@@ -11461,157 +11427,6 @@ rm -f core conftest.err conftest.$ac_objext \
 pgac_args="$pgac_args with_xml_prefix with_xml_exec_prefix enable_xmltest"
 CFLAGS="$CFLAGS $XML_CPPFLAGS"
 
-# libhdfs3
```
xunzhang: Why delete configure test for libhdfs3 here?

xunzhang:
I think it is just a workaround solution. HAWQ use autotools but in this pull request, `libyarn` and `libhdfs3` use makefile which is dangerous for portability.
I think a better solution is to write a shell/python script which is transparent to user that could build `libyarn` and `libhdfs3` before building `hawq`.

configure
paul-guo-: We do not build libhdfs3 separately now so we do not need to check the dependency.

paul-guo-:
To Huor, I checked the source code. It seems that it is easy to add some Makefile targets for the issues you mentioned. To better track them, I'd leave them to another issue.

To xunzhang,  yes, we should do them in configure or make to make the build of libhdfs3 and libyarn transparent to users. Currently doing it in "make pre_config" (dependency of "make all") seems to be better to handle. Maybe I misunderstood your words.

Thanks all.

huor:
+1

xunzhang:
+1

xunzhang:
@paul-guo- plz sync `configure.in` asap. Any problem, ping me.

paul-guo-:
Merged.

PR #659, opened 2016-05-17
HAWQ-738. Allocate query resource twice in function call through jdbc

huor:
Root cause analysis:

1. It uses exec_parse_message and exec_bind_message and allocate query resource twice in parse and bind phases accordingly when using jdbc to call the function: SELECT public.fn_debug(). Thus query resource is allocated twice in function call through jdbc.

2. It uses exec_simple_query and allocate query resource there only once when using psql/libpq to call the function: SELECT public.fn_debug().

The resolution is to do not allocate query resource in parse, and only allocate query resource in bind.


jiny2:
+1 LGTM.

zhangh43:
+1

PR #660, opened 2016-05-17
A temporary pull for test framework review

jiny2:
goh_prepare in checkinstall-good


src/test/feature/query/ans/prepare-parameters.ans
```diff
@@ -0,0 +1,85 @@
...
+    7802 |    7508 |   0 |    2 |   2 |      2 |       2 |      802 |        1802 |      2802 |     7802 |   4 |    5 | COAAAA   | UCLAAA   | AAAAxx
+    8002 |    9980 |   0 |    2 |   2 |      2 |       2 |        2 |           2 |      3002 |     8002 |   4 |    5 | UVAAAA   | WTOAAA   | AAAAxx
+    8302 |    7800 |   0 |    2 |   2 |      2 |       2 |      302 |         302 |      3302 |     8302 |   4 |    5 | IHAAAA   | AOLAAA   | AAAAxx
+    8402 |    5708 |   0 |    2 |   2 |      2 |       2 |      402 |         402 |      3402 |     8402 |   4 |    5 | ELAAAA   | OLIAAA   | AAAAxx
+    8602 |    5440 |   0 |    2 |   2 |      2 |       2 |      602 |         602 |      3602 |     8602 |   4 |    5 | WSAAAA   | GBIAAA   | AAAAxx
+    9502 |    1812 |   0 |    2 |   2 |      2 |       2 |      502 |        1502 |      4502 |     9502 |   4 |    5 | MBAAAA   | SRCAAA   | AAAAxx
+    9602 |    9972 |   0 |    2 |   2 |      2 |       2 |      602 |        1602 |      4602 |     9602 |   4 |    5 | IFAAAA   | OTOAAA   | AAAAxx
+(29 rows)
+
+psql:/tmp/TestQueryPrepare_TestPrepareParameters.sql:4: ERROR:  wrong number of parameters for prepared statement "q3"
```
huor: During test run, we need to copy TestQueryPrepare_TestPrepareParameters.sql from $HAWQ_SRC/src/test/feature/query/ to /tmp, that means we suppose /tmp always exists and we have access privileges on it.

src/test/feature/query/test-prepare.cpp
```diff
@@ -0,0 +1,75 @@
...
+class TestQueryPrepare : public ::testing::Test {
+ public:
+  TestQueryPrepare() {}
+  ~TestQueryPrepare() {}
+};
+
+
+TEST_F(TestQueryPrepare, TestPrepareUniqueness) {
+  SQLUtility util;
+  util.execSQLFile("query/sql/prepare-uniqueness.sql",
```
huor: If we run feature-test not in $HAWQ_SRC/src/test/feature, then "query/sql/prepare-uniqueness.sql" might be not found.

src/test/feature/query/test-prepare.cpp
```diff
@@ -0,0 +1,75 @@
...
+		  	   "	odd			int4,"
+		  	   "	even		int4,"
+		  	   "	stringu1	name,"
+		  	   "	stringu2	name,"
+		  	   "	string4		name) with oids");
+  util.execute("create table test2 ("
+		  	   "	name		text,"
+		  	   "	thepath		path)");
+
+  std::string pwd = Command::getCommandOutput("pwd");
```
huor: If we run feature-test not in $HAWQ_SRC/src/test/feature, the "pwd" might not be expected. For example:
1) cd $HAWQ_SRC/src/test/feature; ./feature-test
2) $HAWQ_SRC/src/test/feature/feature-test

src/test/feature/query/ans/prepare-parameters.ans
jiny2: I think this is a sqlutility implementation, what's the concern?

src/test/feature/query/test-prepare.cpp
jiny2: Seems, it has to be run in feature directory ...

ztao1987: There is one bug in SQLUtility::execSQLFile. Fixed now.

src/test/feature/query/ans/prepare-parameters.ans
ztao1987: I think /tmp is supposed to be exist and test have full access to it. If anything wrong happened, exception will be thrown.

src/test/feature/query/test-prepare.cpp
ztao1987: We used __FILE__ which is decided in compilation.

ztao1987: There is one method getTestRootPath to get feature-test abs path. Currently it's private. We can make it public for case usage.

src/test/feature/query/test-prepare.cpp
jiny2: Cool!

src/test/feature/query/test-prepare.cpp
jiny2: I think a function call is better than __FILE__.

ztao1987: Yes, will make it public

PR #662, opened 2016-05-18
HAWQ-740. Fix bug in SQLUtility.

ztao1987:


huor:
+1 Overall it is good, except that a more meaningful name instead of execute1 would be better

ztao1987:
used void execute(const std::string &sql, bool check = true) instead

PR #663, opened 2016-05-18
HAWQ-741. Set system login user as default when PGUSER is not set.

zhangh43:
in SQLUtility we use getenv("PGUSER") as the user to connect to DB. But if PGUSER is not set, we need to use system login user instead of gpadmin.


wengyanqing:
+1

PR #664, opened 2016-05-18
HAWQ-739. Add doxygen template to generate test common lib document.

zhangh43:
after install doxygen, user can generate document of test common lib in doc folder of feature test when running make doc.


wengyanqing:
+1

PR #672, opened 2016-05-23
HAWQ-751. Add plr, pgcrypto, gporca into Apache HAWQ

paul-guo-:


configure
```diff
@@ -1545,7 +1543,8 @@ Optional Packages:
   --with-perl             build Perl modules (PL/Perl)
   --with-python           build Python modules (PL/Python)
   --with-java             build Java modules (PL/Java)
-  --with-r             build R modules (PL/R)
+  --with-r                build R modules (PL/R)
```
huor: It would be better to use:
```
--with-plperl
--with-plpython
--with-pljava
--with-plr
--with-pgcrypto
```

Also the comment for "with-pgcrypto" should be "build with pgcrypto"

paul-guo-: --with-perl, ... follows postgresql though I agree with you but we can keep it for tradition.  pgcypto typo, yes, thanks.

huor:
+1

paul-guo-:
Merged into upstream.

PR #679, opened 2016-05-27
HAWQ-760. Add hawq register to support registering parquet file(s) into hawq

Oliver-Luo:
Users sometimes want to register data files generated by other system like hive into hawq. These commits add support for the registration for parquet-format file(s). Users could integrate their external parquet file(s) into hawq conveniently.


tools/doc/hawqregister_help
```diff
@@ -0,0 +1,97 @@
...
+EXAMPLES
+*****************************************************
+
+Run hawq register to register a parquet file with path '/temp/hive.paq' 
+in hdfs generated by hive into table 'p' in hawq, which is in the
+database named 'postgres'.
+Assume the location of the database is 'hdfs://localhost:8020/hawq_default',
+tablespace id is '16385', database id is '16387', table filenode id is '77160'.
+
+$ hawq register posrgres p /temp/hive.paq
```
changleicn: typo here: posrgres => postgres

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,97 @@
+COMMAND NAME: hawq register
+
+Register parquet files generated by other system into the corrsponding table in hawq
```
changleicn: hawq => HAWQ

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,97 @@
...
+
+*****************************************************
+DESCRIPTION
+*****************************************************
+
+Hawq register si a utility to register file(s) in hdfs into
+the table in hawq. It will move the file in the path(if path
+refers to a file) or files under the path(if path refers to a
+directory) into a new place corresponding to the table <tablename>,
+and then update the meatadata in the table.
```
changleicn: Hawq register si a utility to register file(s) in hdfs into the table in hawq. It will move the file in the path(if path refers to a file) or files under the path(if path refers to a directory) into a new place corresponding to the table <tablename>,
and then update the meatadata in the table.

=>

"hawq register" is a utility to register file(s) on HDFS into the table in HAWQ. It moves the file in the path (if path refers to a file) or files under the path (if path refers to a directory) into the table directory corresponding to the table <tablename>, and then updates the table meta data to include the files.

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,97 @@
...
+
+To use hawq register, HAWQ must have been started.
```
changleicn: hawq register => "hawq register"

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,97 @@
...
+
+Currently hawq register supports parquet file(s) and table only.
+User have to make sure that the structure of the parquet file(s)
+and the table are the same.
+
```
changleicn: Currently "hawq register" supports parquet tables only. Users have to make sure that the meta data of the parquet file(s) and the table are consistent.

src/backend/cdb/cdbparquetfooterserializer.c
```diff
@@ -577,12 +620,14 @@ readRowGroupInfo(
 	if (!isset_columns)
 		ereport(ERROR,
 				(errcode(ERRCODE_GP_INTERNAL_ERROR), errmsg("file metadata row group column chunk not set")));
+	#if 0
 	if (!isset_total_byte_size)
 		ereport(ERROR,
 				(errcode(ERRCODE_GP_INTERNAL_ERROR), errmsg("file metadata row group total byte size not set")));
```
ictmalili: Can this if 0 be removed now? I think this should be removed.

Oliver-Luo: Yes, these are some codes left due to the debugging process. Thanks for mentioning that. I will check that again.

tools/doc/hawqregister_help
Oliver-Luo: Thanks for the correction. I will check this again.

tools/doc/hawqregister_help
```diff
@@ -1,33 +1,35 @@
...
+To use "hawq register", HAWQ must have been started.
 
-Currently hawq register supports parquet file(s) and table only.
-User have to make sure that the structure of the parquet file(s)
-and the table are the same.
+Currently "hawq register" supports parquet tables only.
+User have to make sure that the meta data of the parquet file(s)
+and the table are consistent.
+The table to be registered into should not be hash-typed, which
+is created by using "distributed by" statement when creating that table.
```
changleicn: hash-typed => hash distributed

src/test/feature/testhawqregister/test-hawq-register.cpp
```diff
@@ -0,0 +1,80 @@
...
+  ~TestHawqRegister() {}
+};
+
+TEST_F(TestHawqRegister, TestSingleFile) {
+	SQLUtility util;
+	string rootPath(util.getTestRootPath()); 
+	string relativePath("/testhawqregister/test_hawq.paq");
+	string filePath = rootPath + relativePath;
+
+	EXPECT_EQ(Command::getCommandStatus("hadoop fs -put " + filePath + " /hawq_register_hawq.paq"), 0);
```
zhangh43: EXPECT_EQ(expected_value, your_value), so zero should be the first parameter.
some other EXPECT_EQ has the same problem.

zhangh43:
The main hawq register steps are as follows:
get_files_in_hdfs(filepath)
check_parquet_format(options, files)
firstsegno = get_metadata_from_database(options, databasename, seg_name)
move_files_in_hdfs(options, databasename, tablename, files, firstsegno, True)
insert_metadata_into_database(options, databasename, tablename, seg_name, firstsegno, sizes)

I have a question what will happen when one of some steps fails. E.g. hdfs files have been moved, but insert metadata failed, should we restore the moved data?

tools/bin/hawqregister
```diff
@@ -0,0 +1,272 @@
...
+            query += "insert into pg_aoseg.%s values(%d, %d, %d, %d);" % (seg_name, segno, eof, -1, -1)
+            segno += 1
+
+        dburl = dbconn.DbURL(hostname=options.host, port=options.port, username=options.user, dbname=databasename)
+        conn = dbconn.connect(dburl, True)
+        rows = dbconn.execSQL(conn, query)
+	conn.commit()
+        conn.close()
+
+    except DatabaseError, ex:
```
Oliver-Luo: > The main hawq register steps are as follows:
> get_files_in_hdfs(filepath)
> check_parquet_format(options, files)
> firstsegno = get_metadata_from_database(options, databasename, seg_name)
> move_files_in_hdfs(options, databasename, tablename, files, firstsegno, True)
> insert_metadata_into_database(options, databasename, tablename, seg_name, firstsegno, sizes)

If one of the reads fails, we just exit cause no file has been changed. If move_files_in_hdfs fails, the metadata will not be inserted. If insert_metadata_into_database fails, we will move the files back to their original location.

tools/bin/hawqregister
```diff
@@ -0,0 +1,272 @@
...
+        conn = dbconn.connect(dburl, False)
+        rows = dbconn.execSQL(conn, query)
+	conn.commit()
+        conn.close()
+
+    except DatabaseError, ex:
+        logger.error("Failed to connect to database, this script can only be run when the database is up")
+        logger.error("host = %s, port = %d, user = %s, dbname = %s, query = %s" % (options.host, options.port, options.user, databasename, query))
+        sys.exit(1)
+
```
Oliver-Luo: > The main hawq register steps are as follows:
> get_files_in_hdfs(filepath)
> check_parquet_format(options, files)
> firstsegno = get_metadata_from_database(options, databasename, seg_name)
> move_files_in_hdfs(options, databasename, tablename, files, firstsegno, True)
> insert_metadata_into_database(options, databasename, tablename, seg_name, firstsegno, sizes)

If one of the reads fails, we just exit cause no file has been changed. If move_files_in_hdfs fails, the metadata will not be inserted. If insert_metadata_into_database fails, we will move the files back to their original location.

Oliver-Luo:
> The main hawq register steps are as follows:
> get_files_in_hdfs(filepath)
> check_parquet_format(options, files)
> firstsegno = get_metadata_from_database(options, databasename, seg_name)
> move_files_in_hdfs(options, databasename, tablename, files, firstsegno, True)
> insert_metadata_into_database(options, databasename, tablename, seg_name, firstsegno, sizes)

If one of the reads fails, we just exit cause no file has been changed. If move_files_in_hdfs fails, the metadata will not be inserted. If insert_metadata_into_database fails, we will move the files back to their original location.

src/test/feature/testhawqregister/test-hawq-register.cpp
```diff
@@ -0,0 +1,80 @@
...
+	util.query("select * from hawqregister;", 0);
+
+	EXPECT_EQ(0, Command::getCommandStatus("hawq register postgres hawqregister /hawq_register_hawq.paq"));
+
+	util.query("select * from hawqregister;", 3);
+	util.execute("insert into hawqregister values(1);");
+	util.query("select * from hawqregister;", 4);
+}
+
+TEST_F(TestHawqRegister, TestSingleHiveFile) {
```
ictmalili: @Oliver-Luo Have you written tests about all the types? It would be better to check all the types we support.

ictmalili: And it's also better to add some cases for negative paths. For example, load data into hash table.

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,101 @@
...
+
+*****************************************************
+CONNECTION OPTIONS
+*****************************************************
+
+-h hostname
+  
+  Specifies the host name of the machine on which the HAWQ master
+  database server is running. If not specified, reads from the
+  environment variable $PGHOST or defaults to localhost.
```
ictmalili: What's the meaning for "or"? It may confuse user.  I think we can change "or" to "which".

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,101 @@
...
+
+-p port
+  
+  Specifies the TCP port on which the HAWQ master database server 
+  is listening for connections. If not specified, reads from the
+  environment variable $PGPORT or defaults to 5432.
```
ictmalili: Change "or" to "which"

tools/doc/hawqregister_help
```diff
@@ -0,0 +1,101 @@
...
+
+-U username
+
+  The database role name to connect as. If not specified, reads
+  from the environment variable $PGUSER or defaults to the current
```
ictmalili: change "or" to "which"

ictmalili:
LGTM. +1

changleicn:
LGTM

GodenYao:
How do you expect users to execute this in their SQL script?
-1

changleicn:
@GodenYao, good point. from sql side, it is just like loading data into native tables. the pure sql flow is like this: 

1) create table
2) create external web table that executes the "hawq register" command and load the data
3) more queries

GodenYao:
Thanks @changleicn , my questions:
1) where is the logic for external web tables to invoke this cmd tool in the code change? 
2) create external tables has special semantic meaning in this case, I don't see a good way to combine this logic with existing external table DDL

changleicn:
Please see the doc here: http://hdb.docs.pivotal.io/200/hawq/reference/sql/CREATE-EXTERNAL-TABLE.html

we also use external tables to execute "dbgen" command to load tpch data. this is somewhat common from customer side when loading data.

GodenYao:
I'm aware of grammar of external web table from GPDB.
Here's an example:
```
=# CREATE EXTERNAL WEB TABLE output (output text)
    EXECUTE 'PATH=/home/gpadmin/programs; export PATH; myprogram.sh' 
    FORMAT 'TEXT';
```

However, your implementation requires creating a native table first - which is not falling into create external web table DDL at all.

Then user needs to run cmdline tool to **register** a new path of the file which would swap the original paq file created by the original DDL.

changleicn:
Had a offline discussion with @GodenYao. This command line is used to register existing hdfs files to hawq native tables and it is used in use cases such as ingesting data from spark, storm et al. And we should not mix this with other external table usage.

shivzone:
@changleicn I don't see why the cases are different. Data ingestion generally means you are copying data from a source to a destination. We aren't quite doing in the case of HAWQ.
Ability to access existing unmanaged hdfs file(s) to hawq table is exactly what an external table is used for and we shouldn't confuse users with a data ingestion like approach.

changleicn:
@shivzone "mv" is almost zero cost in hdfs. Copy in hawq is serial and slow, does not work here,

shivzone:
Yes, if we treat it as "mv" tis a metadata only operation.
I still think it is beneficial to introduce a command that can be invoked from SQL similar to LOAD DATA or MOVE FROM.

ictmalili:
@shivzone Indeed if we further need to register file from other location, simple "mv" doesn't work. But I think in hdfs it works currently now. 

Btw,@shivzone, @changleicn  can we move the discussion to JIRA  https://issues.apache.org/jira/browse/HAWQ-760?

PR #683, opened 2016-05-31
HAWQ-768. Refactor c++ code for src/test/feature/lib.

xunzhang:
In this pull request, I refactored the code under `src/test/feature/lib` folder with following modification:
- Unify some C++ style: [The #define Guard](https://google.github.io/styleguide/cppguide.html#The__define_Guard), [Namespaces](https://google.github.io/styleguide/cppguide.html#Namespaces), [File Names](https://google.github.io/styleguide/cppguide.html#File_Names)(I think we should be the same style with postgresql which uses underscores, but for makefile targets, we could use dashes), [Function Declarations and Definitions](https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions) and so on.
- Refactor the implementation of string_util, xml_util and hawq_config, which contain some potential risks of original impl.
- Fix the clang++ compiling warning in `sql_util.cpp`:
```
lib/sql-util.cpp:112:10: warning: moving a local object in a return statement prevents copy elision
      [-Wpessimizing-move]
  return std::move(psql);
         ^
lib/sql-util.cpp:112:10: note: remove std::move call here
  return std::move(psql);
         ^~~~~~~~~~    ~
1 warning generated.
```

- Separate the modification into three commits for review convenience.

NOTICE: the interface of `SQLUtility` and `HawqConfig` is unchanged. But user must specify them with `using` clause, check [this](https://github.com/xunzhang/incubator-hawq/blob/d55b160d1363831a3a49659d15400cb83f0d469c/src/test/feature/parquet/test_parquet.cpp#L6) for example.


xunzhang:
cc @zhangh43 plz review the first commit, @wengyanqing @ztao1987 plz review the second commit, @ictmalili plz review the third commit.

xunzhang:
@wengyanqing I try to run `feature-test` together with `valgrind` but got some lost and leak. I am not sure  what's the problem(maybe gtest, mac, valgrind itself), have you ever tried that in running `feature-test`?

ictmalili:
LGTM. +1

wengyanqing:
LGTM

huor:
+1

xunzhang:
Merging this into master, thanks!

PR #690, opened 2016-06-03
HAWQ-778. Refine hawq register, add sanity checks.

Oliver-Luo:
Add check to the existence of command 'hadoop'. If not exist, remind user to export the path of  command 'hadoop' into environment variable $PATH.
Add check to the path entered by user. It must start with 'hdfs://', and the file(s) to be registered and the table in HAWQ must be in the same HDFS cluster.
Modify the output of hawq register to give more information.
Modify the tests to fit the changes above.


tools/bin/hawqregister
```diff
@@ -103,6 +113,61 @@ def check_hash_type(options, databasename, tablename):
...
+def check_files_and_table_in_same_hdfs_cluster(filepath, tabledir):
+    # check whether the files to be registered is in hdfs
+    filesystem = filepath.split('://')
+    if filesystem[0] != 'hdfs':
+        logger.error("Only support to register file(s) in hdfs")
+        sys.exit(1)
+    fileroot = filepath.split('/')
+    tableroot = tabledir.split('/')
+    print fileroot
+    print tableroot
```
ictmalili: I think it's better not to print meaningless codes out to console, since the command result is directly output to user, instead of written into log. Make sure every output it meaningful to user is necessary.
Currently when I run the hawq register, the output contains following:

['hdfs:', '', 'localhost:8020', 'tmpData']
['hdfs:', '', 'localhost:8020', 'hawq_default', '16385', '16387', '24616', '']

Please modify it to more meaningful sentences.

tools/bin/hawqregister
```diff
@@ -103,6 +113,61 @@ def check_hash_type(options, databasename, tablename):
...
+    # check the root url of them. eg: for 'hdfs://localhost:8020/temp/tempfile', we check 'hdfs://localohst:8020'
+    if fileroot[0] != tableroot[0] or fileroot[1] != tableroot[1] or fileroot[2] != tableroot[2]:
+        logger.error("Files to be registered and the table are not in the same hdfs cluster.")
```
ictmalili: It's better to print out the two hdfs clusters.

GodenYao:
What's the issue for registering files from another HDFS cluster?
If the node that's running the tool can access the other cluster, wouldn't it succeed? (as your register will pull the file into hawq data directory)?

ictmalili:
@GodenYao Currently the implementation is by calling "dfs mv" command, this command doesn't support moving data file between clusters.  If there's need for moving from another cluster, we can modify this policy.

ictmalili:
Have merged this pull request to hawq master.

changleicn:
@Oliver-Luo please close the pull request since the code is in.

PR #698, opened 2016-06-08
HAWQ-793. Temporarily remove the snappy info in metadata but keep the…

paul-guo-:
… snappy support for row oriented storage.


ictmalili:
LGTM. +1

PR #699, opened 2016-06-08
HAWQ-795. Import apache/orc into depends folder.

xunzhang:


xunzhang:
cc @ictmalili @liming01

xunzhang:
I will send another pull request soon which contains HAWQ-795, HAWQ-796, HAWQ-797 together, so this pull request makes no sense any more.

PR #701, opened 2016-06-10
HAWQ 795,796,797

xunzhang:
Extend apache/orc to support HDFS read. 
This pull request is contained first three sub-tasks of [HAWQ-786](https://issues.apache.org/jira/browse/HAWQ-786) which are listed below:
 - [HAWQ-795](https://issues.apache.org/jira/browse/HAWQ-795) Import apache/orc library for further development.
 - [HAWQ-796](https://issues.apache.org/jira/browse/HAWQ-796) Extend orc library to support hdfs located file read.
 - [HAWQ-797](https://issues.apache.org/jira/browse/HAWQ-797) Implement reading tools for testing HDFS located orc file usage.


xunzhang:
cc @liming01 @ictmalili @changleicn

depends/thirdparty/orc/tools/src/HdfsFileScan.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include "orc/ColumnPrinter.hh"
+
+#include "Exceptions.hh"
+
+#include "wrap/libhdfs3-wrapper.h"
+
+int main(int argc, char* argv[]) {
+  if (argc < 2) {
+    std::cout << "Usage: orc-scan <filename>\n";
```
liming01: In order to make different to local orc file,  we need to make a different command name or an extra param for all these tool commands,  to specify the orc is on hdfs. Or we need to add code to auto detect which type basing on the protocol prefix (hdfs:// or file:// ) of the URL path.

liming01:
@xunzhang,  it looks fine for me, except the third commit. I added a comment in the code. Thanks.

depends/thirdparty/orc/tools/src/HdfsFileScan.cc
xunzhang: Good advice! We will use prefix to auto-detect the protocol. For default, we regard it as `file://`.

xunzhang:
@liming01 I will modify the usage info inside code in the third commit.

liming01:
LGFM. +1.

depends/thirdparty/orc/tools/src/HdfsFileContents.cc
```diff
@@ -0,0 +1,101 @@
...
+  }
+  auto portStr = std::getenv("LIBHDFS3_NAMENODE_PORT");
+  tPort port = 8020;
+  if (portStr != NULL) {
+    port = static_cast<tPort>(atoi(portStr));
+  }
+  hdfsFS fs = hdfsConnect(nn.c_str(), port);
+  reader = orc::createReader(orc::readHdfsFile(fs, std::string(filename)), opts);
+
+  std::unique_ptr<orc::ColumnVectorBatch> batch = reader->createRowBatch(1000);
```
ictmalili: Can we extract "1000" out to a input parameter?  I think it will be very useful if the utility can help print some information depending on input.

depends/thirdparty/orc/tools/src/HdfsFileMemory.cc
```diff
@@ -0,0 +1,151 @@
...
+    return 1;
+  }
+
+  const std::string COLUMNS_PREFIX = "--columns=";
+  const std::string BATCH_PREFIX = "--batch=";
+  char* filename = ORC_NULLPTR;
+
+  // Default parameters
+  std::list<uint64_t> cols;
+  uint32_t batchSize = 1000;
```
ictmalili: Is this variable configurable?

ictmalili:
LGTM except for the two comments.

depends/thirdparty/orc/tools/src/HdfsFileContents.cc
xunzhang: It's ok, but actually it is not performance test. I will add it in option parser logic.

depends/thirdparty/orc/tools/src/HdfsFileMemory.cc
xunzhang: Ditto

depends/thirdparty/orc/tools/src/HdfsFileMemory.cc
```diff
@@ -0,0 +1,151 @@
...
+  char *param, *value;
+  for (int i = 1; i < argc; i++) {
+    if ( (param = std::strstr(argv[i], COLUMNS_PREFIX.c_str())) ) {
+      value = std::strtok(param+COLUMNS_PREFIX.length(), "," );
+      while (value) {
+        cols.push_back(static_cast<uint64_t>(std::atoi(value)));
+        value = std::strtok(ORC_NULLPTR, "," );
+      }
+    } else if ( (param=strstr(argv[i], BATCH_PREFIX.c_str())) ) {
+      batchSize =
```
xunzhang: @ictmalili It is already configurable now, assignment here using `BATCH_PREFIX = '--batch'`.

xunzhang:
@ictmalili I add option parse with `--batch` in `HdfsFileScan.cc` and `HdfsFileContents.cc`, remaining code is originally configurable. Plz review again.

xunzhang:
Merged into master.

rvs:
As it appears to be a complete fork of the ORC codebase (do correct me if I'm wrong). I'm very much -1 on this. Please see the email thread on dev@hawq because I think we may need to back it out of master and come up with a better approach.

xunzhang:
@rvs You are incorrect: this pull request contains three issues/commits. We add HDFS support for orc read in apache/orc code...
As communicated with @changleicn , now we just maintain the apache/orc codebase ourselves for development efficiency. And afterwards we can contribute our code back to apache/orc.
For further development of orc support related issues in HAWQ, we could wait PM's comments in apache JIRA before making a decision/pull request.

rvs:
Where was it communicated? I saw no email exchange on the dev@hawq mailing list.

Also, please follow with tour comments on the dev@hawq mailing list thread that I started. It would be much easier to have a more involved discussion there rather than on a JIRA.

I do have a lot of concerns about your approach and I'm still -1 until we clarify some of the issues involved.

xunzhang:
@rvs In my previous sight, HAWQ is an apache project which needs to keep our discussions open in JIRA and github. Sorry, I will discussion and comment more on mails in the future. Thanks for your negative points!

rvs:
This is a discussion that belongs on the mailing list. Please do follow up.

PR #704, opened 2016-06-13
HAWQ-803. Enable filter in code coverage.

zhangh43:
For code coverage is time consuming, we want to filter some sub directories which we care about.
Syntax:
make coverage-show filter=./src/backend/executor


huor:
+1

PR #711, opened 2016-06-14
HAWQ-815. KeyError while compiling PLPython function due to deletion of non-existent record from Python global dict

huor:


zhangh43:
+1

liming01:
+1.

PR #721, opened 2016-06-16
HAWQ-822. Add string replacement utility in feature test framework to support convert from source to sql and ans files

huor:


yaoj2:
+1

src/test/feature/lib/file_replace.cpp
```diff
@@ -0,0 +1,56 @@
...
+		start_pos += dst.length();
+	}
+
+	return str;
+}
+
+void FileReplace::replace(
+	const string& file_src,
+	const string& file_dst,
+	const map<string, string>& strs_src_dst)
```
xunzhang: `std::map` is implemented with red-black tree which not truly hashmap, using `std::unordered_map` instead?
`#include <unordered_map>`
`std::unordered_map<string, string> dict;`
`dict["hello"] = "world";`

src/test/feature/lib/file_replace.cpp
```diff
@@ -0,0 +1,56 @@
...
+{
+	ifstream fin(file_src);
+	ofstream fout(file_dst);
+	string line;
+
+	while ( getline(fin, line) )
+	{
+		typename map<string, string>::const_iterator mit;
```
xunzhang: Here you don't actually need declare it, you can use `auto`: `for(auto mit = strs_src_dst.begin(); mit != strs_src_dst.end(); mit++)`. Instead, here you can use syntax like:
`for(auto & kv : mit) { auto a = kv.first; auto b = kv.second; }`

src/test/feature/lib/file_replace.h
```diff
@@ -0,0 +1,26 @@
...
+#include <map>
+
+namespace hawq {
+namespace test {
+
+class FileReplace
+{
+public:
+	FileReplace() {}
+	~FileReplace() {}
```
xunzhang: Do you mean not to define the copy constructor here? If you don't define anyone of constructor, de-constructor, copy constructor, the compiler will auto create them. If you do define some of them, the other could not be auto created.

src/test/feature/lib/file_replace.cpp
huor: It is better to use unordered_map instead of map.
For adding element in unordered_map or map, I would prefer insert instead of operator [] since insert would do a check if duplicate keys are going to be added.

src/test/feature/lib/file_replace.h
huor: We don't need copy constructor here for FileReplace. We can explicitly do that in below way:
```

FileReplace() = default;
~FileReplace() = default;
FileReplace(const FileReplace&) = delete;
FileReplace& operator=(const FileReplace&) = delete;
```

src/test/feature/lib/file_replace.cpp
huor: Cool.

xunzhang:
+1

PR #723, opened 2016-06-16
HAWQ-824. Move orca submodules to depends/

paul-guo-:
This patch also optmized the makefiles for gpopt/gpora/gpos/gp-xerces.


wengyanqing:
LGTM

PR #731, opened 2016-06-20
HAWQ-830. Fix wrong result in CTE query due to CTE is treated as init plan by planner and evaluated multiple times

huor:


zhangh43:
GUC gp_cte_sharing cannot be false, and should be removed.
Because it's not consistent with CTE semantic(CTE should not be init plan more than once).

zhangh43:
We can keep this GUC gp_cte_sharing temporarily
+1

xunzhang:
+1

vraghavan78:
gp_cte_sharing when set to true, the planner sometimes gets into a deadlock.

huor:
@vraghavan78, yes, you are correct.

gp_cte_sharing was introduced to workaround deadlock in shared scan in some CTE queries. For example, https://issues.apache.org/jira/browse/HAWQ-852 is a hang issue when gp_cte_sharing is on.

However, if we disable gp_cte_sharing (especially by default), we get wrong result when running some other CTE queries. One of the common root cause is that: with gp_cte_sharing off, some of the CTE expressions are evaluated multiple times in one single query; if the CTE expression itself is somehow "volatile", it is obviously that wrong result (at least inconsistent result) will be generated.
For example, https://issues.apache.org/jira/browse/HAWQ-830, https://issues.apache.org/jira/browse/HAWQ-852 are two issues that generate wrong results with CTE query.

My recommendation is: we need to enable gp_cte_sharing by default (actually this GUC should be removed ideally), and then fix the hang issue in shared scan. Your thoughts?

edespino:
@huor - Any update on this PR?  There hasn't been any progress on it in months. This is the same for the corresponding Jira HAWQ-830

huor:
@edespino, it is better to merge this fix along with extra fix for wrong result and hang in CTE query. Given there is a lot of effort to fix the wrong result and hang in CTE, I think we can close this pr without merge it for now. We can do that when they are fixed.

edespino:
@huor - Yes let's close this PR for now.  Thanks, -=e

huor:
Close the pr for now. Need to merge the fix after wrong result and hang in CTE query is fixed.

PR #734, opened 2016-06-20
HAWQ-841. Running "make -j8 unittest-check" under src/backend will throw "File exists" error

paul-guo-:
Typical race condition issue.


yaoj2:
+1

xunzhang:
+1

paul-guo-:
Merged.

PR #735, opened 2016-06-21
HAWQ-837. Add python modules into HAWQ code

radarwave:
Added figleaf, pg8000, simplejson, pyyaml, lockfile, psi, pychecker, unittest2
Paramiko is not included because of License concern.
Pycrypto is not included because not in use anymore.

This will save user's time to install python modules by themselves.


yaoj2:
+1

huor:
+1

PR #742, opened 2016-06-23
HAWQ-863. Add python module pycrypto

radarwave:
Add this module to reduce user install works. So user don't need to do 'pip install pycrypto'.


yaoj2:
+1

xunzhang:
+1

PR #743, opened 2016-06-23
HAWQ-855. Make resource manager ask for cluster report quickly when YARN is detected down

jiny2:


jiny2:
Close and open another one, this is not a clean commit

PR #745, opened 2016-06-23
HAWQ-850. Planner supports refineCachedPlan interface.

zhangh43:
in PBE(prepare bind execution) cases, plan will be cached, and only be generated once in postgres.
But in HAWQ since resource is dynamic or elastic, the plan maybe changed during the different execution of PBE. Also, the file split allocation result which is stored in plan is dynamically too, so the split-vseg mapping may also be changed.

For example, We call a plpgsql function which do "insert into t select * from t".
in Prepare we cache the plan, with fixed split-vseg mapping(e.g. 10 blocks), but after insert, there would be 20 blocks so the old  split-vseg mapping is invalid. We need to update it. Also when we call this function again and again, the data size of t may request more resource to handle it than the first call, so we should recalculate the entire plan with the new vseg number.

This function is implemented in a interface called refineCachedPlan.
[Link to Jira](https://issues.apache.org/jira/browse/HAWQ-850)

### DIFF
--- a/src/backend/optimizer/plan/planner.c
+++ b/src/backend/optimizer/plan/planner.c
@@ -342,7 +342,7 @@ PlannedStmt *refineCachedPlan(PlannedStmt * plannedstmt,
     * then fall back to the planner.
     * TODO: caragg 11/08/2013: Enable ORCA when running in utility mode (MPP-21841)
     */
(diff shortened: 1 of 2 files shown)


ictmalili:
LGTM

wengyanqing:
LGTM

PR #746, opened 2016-06-23
HAWQ-859. Fix the naming of output columns of append node [#122086135]

vraghavan78:
In the DXL to PlStmt translator, we do not respect the renaming of the columns of the append node.

@oarap @d @hsyuan


hsyuan:
LGTM

PR #749, opened 2016-06-27
HAWQ-800: Fix less tuple is inserted due to data locality information is not refreshed and dispatched in prepared statement

huor:


xunzhang:
+1

liming01:
Looks good to me. +1.

PR #750, opened 2016-06-27
HAWQ-835. Fix cannot retrieve tuple from temp table created in function

huor:


xunzhang:
LGTM

liming01:
It is better add some comments. +1.

PR #751, opened 2016-06-27
HAWQ-869. Add regression test for less tuple is inserted issue in prepared statement

huor:


huor:
@yaoj2, @xunzhang, @ztao1987, please review.

xunzhang:
+1

PR #753, opened 2016-06-28
HAWQ-861. Wrong logerror position in insert into hashtable select * f…

zhangh43:
…rom gpfdist external table.

For statement insert into hashtable select * from gpfdist external table,
when hashtable bucket number is less than location number of gpfdist external table, error should be reported in cdbdatalocality module, with error message like this:
ERROR:  Could not allocate enough memory! bucket number of result hash table and external table should match each other (cdbdatalocality.c:4222)


wengyanqing:
LGTM

ztao1987:
+1

PR #758, opened 2016-06-29
HAWQ-876. Add the support for initFile option of gpdiff.pl in hawq go…

paul-guo-:
…ogletest framework.

	This patch also removes several annoying warnings.


wengyanqing:
LGTM

ztao1987:
+1

src/test/feature/lib/hawq_config.cpp
```diff
@@ -151,7 +151,7 @@ bool HawqConfig::isMultinodeMode() {
       psql.getQueryResult("select hostname from gp_segment_configuration");
   std::vector<std::vector<string> > table = result.getRows();
   std::unordered_set<string> hostnameMap;
-  for (int i = 0; i < table.size(); i++) {
+  for (unsigned int i = 0; i < table.size(); i++) {
```
xunzhang: use `size_t` instead

src/test/feature/lib/sql_util.cpp
```diff
@@ -108,8 +111,24 @@ void SQLUtility::execSQLFile(const string &sqlFile,
...
+    // double check to avoid empty fileBaseName
+    if (fp.fileBaseName.empty())
+      ASSERT_TRUE(false) << initFileAbsPath << " is invalid";
+  } else {
+    initFileAbsPath = "";
+  }
+
+  bool is_sql_ans_diff = conn->checkDiff(ansFileAbsPath, outFileAbsPath, true, initFileAbsPath);
+  EXPECT_FALSE(is_sql_ans_diff);
+  if (is_sql_ans_diff == false) {
```
xunzhang: `if (!is_sql_ans_diff) {`

xunzhang: why you need to check here? You have already check it above: `EXPECT_FALSE(is_sql_ans_diff);`, I think the if-else clause is redundant.

paul-guo-:
Merged before xunzhang commented.

src/test/feature/lib/sql_util.cpp
paul-guo-: That logic is for cleanup.

xunzhang: You can directly remove it without if-else.

paul-guo-: I think we can remove the else part, right?

 } else {
    EXPECT_FALSE(true);
  }

PR #761, opened 2016-06-30
HAWQ-812. only hdfs file/dir can be skipped at recovery PASS3

liming01:
The defect was introduced by the fix for https://issues.apache.org/jira/browse/HAWQ-232. mmxlog not only include hdfs file/dir xlog, but also include heap xlog for standby node. Only hdfs file/dir can be skipped at recovery PASS3.

Moreover as for the relfile's path changed from hawq1.x to hawq2.x,  there are some code forgot to change accordingly.


ztao1987:
+1

linwen:
LGTM +1

PR #762, opened 2016-06-30
HAWQ-867. Replace the git-submobule mechanism with git-clone

paul-guo-:


xunzhang:
LGTM
A suggestion: could we use `wget` to download a release source tarball instead of `git clone`? This way will not donwload `.git` folder of dependency libraries.

huor:
+1. It is good to be able to customize branch and commit for dependent project.

PR #763, opened 2016-07-01
HAWQ-885. Should fail earlier at creating for a table with quicklz co…

paul-guo-:
…mpression.

### DIFF
--- a/src/backend/access/common/reloptions.c
+++ b/src/backend/access/common/reloptions.c
@@ -515,22 +515,20 @@ default_reloptions(Datum reloptions, bool validate, char relkind,
 
 		compresstype = values[5];
 
+		if (strcmp(compresstype, "quicklz") == 0)
+		{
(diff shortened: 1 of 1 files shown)


wengyanqing:
LGTM

ztao1987:
+1

src/backend/access/common/reloptions.c
```diff
@@ -515,22 +515,20 @@ default_reloptions(Datum reloptions, bool validate, char relkind,
 
 		compresstype = values[5];
```
huor: The fix should remove quicklz from valid compression type list. It is better to add comment here we reserve quicklz in catalog for ease of upgrade.

paul-guo-: The code is not just for hacking here. Even we remove quicklz from system tables in the future, this code could be still kept to educate users for some time.

PR #766, opened 2016-07-04
HAWQ-887. Query should be continued when bucket number of result hash…

zhangh43:
… relation is bigger than the on number of command external table.


ztao1987:
+1

yaoj2:
+1

PR #768, opened 2016-07-04
HAWQ-888. Hack to avoid deleted files being shown in "git status" aft…

paul-guo-:
…er building.


xunzhang:
It works correct, but I do not recommend this implementation. I think further discussion needs to be done.

PR #771, opened 2016-07-05
HAWQ-867. Replace the git-submobule mechanism with git-clone

paul-guo-:
This is a further code change based on the previous patch for this JIRA.
This patch updates the pgcrypto commit number to the correct one
which we previously used.


radarwave:
+1

xunzhang:
+1

PR #773, opened 2016-07-05
HAWQ-888. Remove some dummy files which were used to keep related "em…

paul-guo-:
…pty" directories.

### DIFF
--- a/contrib/Makefile
+++ b/contrib/Makefile
@@ -25,7 +25,6 @@ pgcrypto_prepare: pgcrypto_prepare_timestamp
 	if [ ! -f $(abs_top_builddir)/$(subdir)/pgcrypto_prepare_timestamp ]; then \
 		[ "x$(pgcrypto_commit)" != "x" ] || exit 1; \
 		cd $(abs_top_srcdir)/depends/thirdparty; mkdir -p postgres; cd postgres; \
-		rm -f git_keep_empty_directory.readme; \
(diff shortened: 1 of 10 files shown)


radarwave:
I think this is better than the previous hack, usually users won't create the directories by themselves.

+1

xunzhang:
Could you add the clone info more specifically? For example, I tried your branch under my Mac and got error below, but I could not even know which repo is failed during cloning.

```
Cloning into '.'...
remote: Counting objects: 589045, done.
remote: Compressing objects: 100% (73/73), done.
error: RPC failed; curl 56 SSLRead() return error -36 251.00 KiB/s   
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
fatal: Could not parse object 'ca59dfa6f727fe3bf3a01904ec30e87f7fa5a67e'.
Now fetch from upstream and try again
From https://github.com/paul-guo-/incubator-hawq
 * branch            master     -> FETCH_HEAD
fatal: Could not parse object 'ca59dfa6f727fe3bf3a01904ec30e87f7fa5a67e'.
make[1]: *** [pgcrypto_prepare] Error 2
make: *** [all] Error 2
```

paul-guo-:
@xunzhang, make fails because "git clone" fails, probably because of network issue or something else. This is not a makefile bug. You can try run make again. By the way, the build log includes the git command line (thus includes repo info). Just check the log before "Cloning into".

xunzhang:
LGMT

PR #775, opened 2016-07-06
HAWQ-893. Only copy and index status_str_length

kavinderd:
Previously status_str was being indexed with a value that overflowed its
expected length and would zero-out the string 'test'


yaoj2:
LGTM

liming01:
+1.

kavinderd:
Merged

PR #778, opened 2016-07-06
HAWQ-900. Add dependency in PL/R rpm build spec file plr.spec

paul-guo-:


src/pl/plr.spec
```diff
@@ -3,10 +3,14 @@ Name:           plr-hawq
 Version:        08.03.00.14
 Release:		0
 Prefix:         /usr/local
-License:		GPL
+License:		Apache
```
huor: Do we use Apache License 2.0 or 1.1, or just use Apache in general?

radarwave:
If the License change is correct. Then +1.

src/pl/plr.spec
```diff
@@ -3,10 +3,14 @@ Name:           plr-hawq
...
+BuildRequires:  R-devel >= 3.1.0
```
huor: I would prefer: R-devel = 3.1.0 and R = 3.1.0. The reason is that we only certified plr in hawq with R 3.1.0, we don't know if there are any compatible issue R > 3.1.0. Any thought?

radarwave: No, we should not use '=' since we are installing R and R-devel via YUM. Users might have higher version YUM repo. It will be very hard to find the correct YUM repo.

huor: Suppose uses have R = 3.x (which is higher than 3.1.0) installed with yum. Since we don't try it out in hawq, what if below two potential issue:
1) plr can not be built and installed due to API change in R
2) plr can be build and installed successfully, but have runtime issue

radarwave: Now HAWQ have many dependencies, 'thrift', 'boost', 'curl', etc. Each have the concerns you mentioned. 

If this really a concern for plr, then I advise to test with the latest one on both centos6 and centos7.

huor: I guess it would be more proper to test "all" versions of R, not "lastest"

src/pl/plr.spec
paul-guo-: I guess we need to move back to GPL which is exactly the license of PL/R. There has been a upstream PL/R spec file in which license is said to be BSD, but I suspect that is wrong.

src/pl/plr.spec
paul-guo-: How about removing the version requirement, if it works fine with R on current rhel 6.x and rhel7?

huor:
+1

PR #780, opened 2016-07-08
HAWQ-872  Fix HAWQ check fails when kerberos is enabled

radarwave:
When kerberos is enabled, we run 'hawq check' without '--kerberos' option,
it will give 'global name 'hadoop' is not defined' error.


xunzhang:
LGTM

huor:
+1

PR #781, opened 2016-07-09
HAWQ-911. Optimize and refactor makefiles for feature test framework.

xunzhang:
This pull request contains below modifications:
1. Fix a bug in building UDF/lib.
2. Refactor feature test make system to avoid duplicate compiles.
3. Split generating `feature-test` binary into generating objects and linking. Optimize compiling time in parallel mode.


xunzhang:
cc @paul-guo- @huor

huor:
Good to have huge performance improvement in making feature test. +1

xunzhang:
@wengyanqing

wengyanqing:
LGTM.

paul-guo-:
The patch seems to have some improvement spaces.
lib/*.o was generated by feature/Makefile, however feature/lib/Makefile has a duplicate makefile rule to generate lib/*.o

I think you could move the rules in feature/lib/Makefile to feature/Makefile. This makes the code neat.

I'd suggest that the lib name could be more hawq or feature_test specific, i.e. libtest -> libfeature_test or something else?

Just my 2 cents.

xunzhang:
@paul-guo- Agree, I will modify it later on. Thanks for comments! 🍺

PR #787, opened 2016-07-12
HAWQ-917. Refactor feature tests for data type check with new googlet…

paul-guo-:
…est framework


wengyanqing:
LGTM

src/test/feature/type/ans/int8.ans
```diff
@@ -175,131 +179,131 @@ SELECT '' AS to_char_4, to_char( (q1 * -1), '9999999999999999S'), to_char( (q2 *
...
-            |                123.000
-            |                456.000
-            |   4567890123456789.000
-            |   4567890123456789.000
-            |  -4567890123456789.000
+            | $              456.000
+            | $ 4567890123456789.000
+            | $              123.000
+            | $ 4567890123456789.000
+            | $-4567890123456789.000
```
xunzhang: Why did you add `$` sign here?

src/test/feature/type/TestType.cpp
```diff
@@ -0,0 +1,77 @@
+#include "gtest/gtest.h"
+
+#include "lib/sql_util.h"
+
+using std::string;
```
xunzhang: add `#incude <string>`

paul-guo-: It could compile without this. sql_util.h has included <string>. Do we really need to include this?

xunzhang:
@paul-guo- , LGTM except above comments. Another comments is that according to the folders names defined by @yaoj2 , I suggest you to move your test cases into `catalog` folder instead of independent `type`.

src/test/feature/type/ans/int8.ans
paul-guo-: I looked into this. I is related to currency info of the locale. You are right this probably need a bit more work.

paul-guo-: I'm leaving the to another JIRA HAWQ-925
Set default locale, timezone & datastyle before running sql command/file.

That is for the whole feature test framework.

xunzhang:
+1

paul-guo-:
Merged.

PR #789, opened 2016-07-12
HAWQ-918. Fix memtuple forming bug when null-saved size is larger tha…

linwen:
Please review, Thanks!


xunzhang:
+1

armenatzoglou:
LGTM.

changleicn:
why do we still add the old installcheck tests?

linwen:
The reason is
test cases for aggregates haven't been moved to new feature test framework yet.
So I add this test case in old installcheck.
And all the test cases for aggregates will be moved together within a new jira.

huor:
+1

PR #793, opened 2016-07-14
HAWQ-925. Set default locale, timezone & datastyle befo.re running sq…

paul-guo-:
…l command/file


paul-guo-:
Closing it. We have checked in another implementation.

PR #795, opened 2016-07-14
HAWQ-860. Fix ORCA wrong plan when correlated subquery contains set-returning functions

hsyuan:
ORCA 1.633 returns wrong result for the following query:
```sql
select 0 is distinct from (select count(1) from (select unnest(array[1, 2, 3])) as foo);
?column?
----------
 t
 t
 t
(3 rows)
```
Correct result should be:
```sql
?column?
----------
 t
(1 row)
```
This bug was fixed by bumping ORCA version to 1.634.

For detail information, see ORCA pull request:
https://github.com/greenplum-db/gporca/pull/49

### DIFF
--- a/depends/thirdparty/gporca.commit
+++ b/depends/thirdparty/gporca.commit
@@ -1,1 +1,1 @@
-https://github.com/greenplum-db/gporca.git master 0d837569e61182f68bf1b92038c6616401790a16
+https://github.com/greenplum-db/gporca.git master 52030fd9cf56b32523f1712b094b1a2f8da8505c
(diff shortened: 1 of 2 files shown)


hsyuan:
@yaoj2 @wengyanqing @zhangh43 Please take a look.

vraghavan78:
LGTM

hsyuan:
Pushed to master:
https://github.com/apache/incubator-hawq/commit/e10f9a4e17f3012bb5afb98c53172dedefc8854e
https://github.com/apache/incubator-hawq/commit/4e08f10fc87301682721d6bffcb74a188ac7b595

PR #800, opened 2016-07-19
HAWQ-936. Add GUC for array expansion in ORCA optimizer

hsyuan:
Consider the query with the following pattern 'select * from foo where foo.a IN
(1,2,3,...)'. Currently, when the number of constants in the IN subquery is large,
the query optimization time is unacceptable. This is stopping customers from turning
Orca on by default since many of the queries are generated queries with such a
pattern. The root cause is due to the expansion of the IN subquery into an
expression in disjunctive normal form. The objective of this patch is to add a guc
parameter 'optimizer_array_expansion_threshold' to control the maximum
number of array elements in IN array list. By default, the GUC value is
set to 25, which means if the array size is larger than 25, the array
expansion is disabled.

### DIFF
--- a/depends/thirdparty/gporca.commit
+++ b/depends/thirdparty/gporca.commit
@@ -1,1 +1,1 @@
-https://github.com/greenplum-db/gporca.git master 0d837569e61182f68bf1b92038c6616401790a16
+https://github.com/greenplum-db/gporca.git master c5e40f283703b5fa4c2eb40f367ab7c1b1ab4d0d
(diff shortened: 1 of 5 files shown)


hsyuan:
This PR will give https://github.com/apache/incubator-hawq/pull/795 a free ride.

changleicn:
LGTM

vraghavan78:
LGTM

hsyuan:
pushed to master:
https://github.com/apache/incubator-hawq/commit/950065bb0db627992be6b96afb4671b3c5ab1c90

PR #804, opened 2016-07-20
HAWQ-940. Fix Kerberos ticket expired for libyarn operations

linwen:
Please review, thanks!


linwen:
After re-think this fix, I think it's better to check ticket expiration in resource broker process loop.
so close it.

linwen:
Another way of fix is to add the ticket check in resource broker process loop, for every time interval, login() is called. But this fix has to keep another variable to record the last updated time, which is duplicate with login().

jiny2:
LGTM +1

paul-guo-:
+1

PR #807, opened 2016-07-21
HAWQ-897. Add feature test for create table distribution with new test framework

jiny2:


linwen:
+1

yaoj2:
LGTM

src/test/feature/catalog/test_create_table.cpp
```diff
@@ -22,26 +22,26 @@ class TestCreateTable : public ::testing::Test {
 TEST_F(TestCreateTable, TestCreateTable1) {
   hawq::test::SQLUtility util;
   // prepare
-  util.execute("DROP TABLE IF EXISTS aggtest");
```
hsyuan: The easier way to do this is creating a new schema `X`, set the default search_path to `X`. At the end of the test, drop the schema. In this way, you don't need to take care of dropping tables when you add new test cases. This is how GPDB is doing.

paul-guo-: I think jinyi's intent is to test "create/drop table".

Actually the constructor and destructor of class hawq::test::SQLUtility are exactly doing what you've suggested.

See src/test/feature/lib/sql_util.cpp for details.

PR #808, opened 2016-07-21
HAWQ-944. Implement new pg_ltoa function as per postgres

kavinderd:


kavinderd:
@paul-guo- @xunzhang Please review, I think I covered all invocations of the two functions.

xunzhang:
@kavinderd I will review this next Monday, sorry for the latency.  If it is a hurry, you could at somebody else to review. Thanks!

kavinderd:
@xunzhang No rush, take your time

shivzone:
I would prefer if we have a general util function that determines precisely the number of bytes to "stringify" the given data type (int16 or int32) as a character array

src/backend/access/external/pxfutils.c
```diff
@@ -39,7 +39,7 @@ bool are_ips_equal(char *ip1, char *ip2)
 /* override port str with given new port int */
 void port_to_str(char **port, int new_port)
 {
-	char tmp[10];
+	char tmp[sizeof(new_port) + 2];
```
shivzone: This will not be sufficient. sizeof would return 4 and tmp would be allocated only 6 bytes which wouldn't be sufficient to store the 32 bit interger as a string

paul-guo-:
I suspect pg_ltoa() will be faster if it fills the characters one by one in descending order, instead of current solution: filling in ascending order and then swapping.

Given the code comes from pg, I'd give a +1 at first.

shivzone:
I don't see INT32_CHAR_SIZE defined anywhere

kavinderd:
Added `INT32_CHAR_SIZE` to `configure.in`

paul-guo-:
It looks like pg_ltoa() does not work for the corner case: integer 0?

edespino:
@kavinderd - Any progress on this issue?  It is currently in a conflict situation as reported in the PR checks "This branch has conflicts that must be resolved".  There doesn't appear to be any progress on the corresponding Jira as well (HAWQ-944) since last summer (July 2016).

kavinderd:
@huor @paul-guo- Can you look at this PR when you get a chance. I need one more +1

huor:
+1

paul-guo-:
+1

PR #811, opened 2016-07-25
HAWQ-1000: Set dummy workfile pointer to NULL after calling ExecWork…

liming01:
…File_Close()

The root cause of defect GPSQL-3288 maybe:
the parameter workfile for ExecWorkFile_Close() is freed in this function, but in the calling function outside, the pointer variable still exists, we need to set it to NULL pointer immediately. I checked all valid functions cascade calling this function, only find 2 occurrence of suspect.

### DIFF
--- a/src/backend/utils/sort/tuplesort.c
+++ b/src/backend/utils/sort/tuplesort.c
@@ -915,11 +915,13 @@ tuplesort_end(Tuplesortstate *state)
 	if (state->tapeset)
 	{
 		LogicalTapeSetClose(state->tapeset, NULL /* workset */);
+		state->tapeset = NULL;
 
 		if (state->pfile_rwfile_state)
         {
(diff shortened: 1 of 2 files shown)


zhangh43:
+1

GodenYao:
even this is an internal customer JIRA, we should still file Apache JIRA so the repro steps and RCA are available to public.

liming01:
Thanks @zhangh43, @GodenYao, 
Yes, I will open a new Apache JIRA for it.

xunzhang:
Additional risks:
- https://github.com/apache/incubator-hawq/blob/master/src/backend/executor/nodeHashjoin.c#L1576
- https://github.com/apache/incubator-hawq/blob/master/src/backend/utils/sort/logtape.c#L429

xunzhang:
Above defects are covered into the modification, LGTM. 

🍺 HAWQ-1000

zhangh43:
+1

PR #815, opened 2016-07-26
HAWQ-898. Add feature test for COPY with new test framework

jiny2:


linwen:
+1

yaoj2:
LGTM

jiny2:
Revert COPY checkinstall-good case because transactions and row_types need this to prepare data

PR #818, opened 2016-07-27
HAWQ-955. Add scriptS for feature test running in parallel.

xunzhang:


xunzhang:
cc @yaoj2 @paul-guo-

xunzhang:
Usage: `./src/test/feature/run-feature-test.sh 8 ./src/test/feature/feature-test --gtest_filter=xx`

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,74 @@
...
+  if [ $? -ne 0 ]; then
+    echo "Alter database hawq_feature_test set timezone failed."
+    exit 1
+  fi
+  $PSQL -d $HAWQ_DB -h $HAWQ_HOST -p $HAWQ_PORT \
+        -c "alter database $TEST_DB_NAME set datestyle to 'postgres,MDY';" > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+    echo "Alter database hawq_feature_test set datestyle failed."
+    exit 1
+  fi
```
yaoj2: It would be better make runsql and error check in a function

xunzhang: Good suggestion! Let me do this.

xunzhang:
also @radarwave

radarwave:
+1

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,47 @@
+#! /bin/bash
+
+if [ x$GPHOME == 'x' ]; then
+  echo "Please export GPHOME variable."
+  exit 0
+fi
```
paul-guo-: I think we need more words. How about this?

Please export the environment variable GPHOME before running.
Typically you can run " source greenplum_path.sh" under
the hawq installation path.

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,47 @@
...
+run_sql() {
+  $PSQL -d $HAWQ_DB -h $HAWQ_HOST -p $HAWQ_PORT -c "$1" > /dev/null 2>&1
+  if [ $? -ne 0 ]; then
+    echo "$1 failed."
+    exit 1
+  fi
+}
+
+init_hawq_test() {
+  source "${GPHOME}/greenplum_path.sh"
```
paul-guo-: I think "source greenplum_path.sh" should be run before running the case? What is you opinion?

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,47 @@
...
+  
+  TEST_DB_NAME="hawq_feature_test"
```
paul-guo-: I'm wondering whether change the name from hawq_feature_test to hawq_feature_test_db.

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,47 @@
...
+  run_sql "alter database $TEST_DB_NAME set lc_numeric to 'C';"
+  run_sql "alter database $TEST_DB_NAME set lc_time to 'C';"
+  run_sql "alter database $TEST_DB_NAME set timezone_abbreviations to 'Default';"
+  run_sql "alter database $TEST_DB_NAME set timezone to 'PST8PDT';"
+  run_sql "alter database $TEST_DB_NAME set datestyle to 'postgres,MDY';"
+  PGDATABASE=$TEST_DB_NAME
+}
+
+run_feature_test() {
+  if [ $# -lt 2 ] || [ $1 == "--help" ]; then
```
paul-guo-: Add a line in help output to explain what is the script for? (for running in parallel)

I suspect '$1 == "--help"' could be removed.

src/test/feature/run-feature-test.sh
```diff
@@ -0,0 +1,47 @@
...
+
+if [ x$GPHOME == 'x' ]; then
+  echo "Please export GPHOME variable."
+  exit 0
+fi
+
+PSQL=${GPHOME}/bin/psql
+HAWQ_DB=${PGDATABASE:-"postgres"}
+HAWQ_HOST=${PGHOST:-"localhost"}
+HAWQ_PORT=${PGPORT:-"5432"}
```
paul-guo-: Why ignore HAWQ_PASSWORD? It is easier to obtain in shell.

paul-guo-:
By the way, you might need to update README.MD along with the change.

paul-guo-:
I'm wondering whether need to rename the shell script file name to reflect its parallel-run feature.

xunzhang:
Actually I don't want to update README.md this time. We are not sure that the script is working well till now so users could still run these tests in old way until we are sure this works.

I like the name, I will update soon.

src/test/feature/run-feature-test.sh
xunzhang: Why it should be removed?

src/test/feature/run-feature-test.sh
xunzhang: Ok

src/test/feature/run-feature-test.sh
xunzhang: It is!

src/test/feature/run-feature-test.sh
xunzhang: Ok

src/test/feature/run-feature-test.sh
xunzhang: Some we can not run a command like psql -c "xx" -U.

xunzhang: Since

src/test/feature/run-feature-test.sh
xunzhang: In the logic, user do not need to know anything about source in running feature tests. In running this script, the only thing users need to do is set GPHOME. So I think the info is enough. Thoughts?

paul-guo-: Actually for direct run, you need to source greenplum_path.sh at first, thus "source greenplum_path.sh" could be the same prerequisite for both direct run and parallel run.
We could check GP_HOME just to simply but not strictly judge whether users have sourced that file or not, not for us to source that file ourselves in the code.

Besides, another benefit is that you could make doc neat in README.MD :-) see the existing related doc below.

"Make sure HAWQ is running correctly. If not, init or start HAWQ at first.
Load environment configuration by running source $INSTALL_PREFIX/greenplum_path.sh.
Load hdfs configuration. For example, export HADOOP_HOME=/Users/wuhong/hadoop-2.7.2 && export PATH=${PATH}:${HADOOP_HOME}/bin. Since some test cases need hdfs and hadoop command, just ensure these commands work before running. Otherwise you will get failure."

src/test/feature/run-feature-test.sh
paul-guo-: I think "program --help" could be caught by "[ $# -lt 2]"

xunzhang: Oh, I think I will change the first condition, or it is a little tricky then.

src/test/feature/run-feature-test.sh
xunzhang: GOT

src/test/feature/parallel-run-feature-test.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [ x$GPHOME == 'x' ]; then
+  echo "Please source greenplum_path.sh before running feature tests."
+  exit 0
+fi
+
+PSQL=${GPHOME}/bin/psql
+HAWQ_DB=${PGDATABASE:-"postgres"}
+HAWQ_HOST=${PGHOST:-"localhost"}
+HAWQ_PORT=${PGPORT:-"5432"}
```
paul-guo-: HAWQ_PASSWORD=${PGPASSWORD:-""}

paul-guo-: Others look good.

xunzhang: Done, please review again.

paul-guo-:
+1

xunzhang:
Merged into master.

PR #822, opened 2016-07-28
HAWQ-905.  Add feature test for temp table with new test framework

wengyanqing:
1. sqlutility to support schema/database mode
2. temp table test case with new feature framework


paul-guo-:
+1

PR #828, opened 2016-08-01
HAWQ-969. Add getting configuration from HDFS and YARN

wcl14:
Add getting configuration from HDFS and YARN, and writing xml file in xml_parser.cpp.

### DIFF
--- a/src/test/feature/lib/hdfs_config.cpp
+++ b/src/test/feature/lib/hdfs_config.cpp
@@ -19,9 +19,9 @@ void HdfsConfig::runCommand(const string &command,
                             string &result) {
   string cmd = "";
   if (ishdfsuser) {
-    cmd = "sudo -u ";
+    cmd = "/usr/bin/sudo -Eu ";
     cmd.append(getHdfsUser());
-    cmd.append(" ");
+    cmd.append(" env \"PATH=$PATH\" ");
     cmd.append(command);
   } else {
     cmd = command;
@@ -154,7 +154,7 @@ int HdfsConfig::isTruncate() {
 
 string HdfsConfig::getHadoopHome() {
   string result = "";
-  runCommand("ps -ef|grep hadoop", true, result);
+  runCommand("ps -ef|grep hadoop", false, result);
   string hadoopHome = "";
   auto lines = hawq::test::split(result, '\n');
   for (size_t i=0; i<lines.size()-1; i++) {
--- a/src/test/feature/lib/hdfs_config.cpp
+++ b/src/test/feature/lib/hdfs_config.cpp
@@ -185,7 +185,7 @@ bool HdfsConfig::getStandbyNamenode(string &standbynamenode,
 bool HdfsConfig::getHANamenode(const string &namenodetype,
                                string &namenode,
                                int &port) {
-  if (!isHA()) {
+  if (isHA() <= 0) {
     return false;
   }
   string namenodeService = "";
--- a/src/test/feature/lib/yarn_config.cpp
+++ b/src/test/feature/lib/yarn_config.cpp
@@ -19,9 +19,9 @@ void YarnConfig::runCommand(const string &command,
                             string &result) {
   string cmd = "";
   if (isyarnuser) {
-    cmd = "sudo -u ";
+    cmd = "/usr/bin/sudo -Eu ";
     cmd.append(getYarnUser());
-    cmd.append(" ");
+    cmd.append(" env \"PATH=$PATH\" ");
     cmd.append(command);
   } else {
     cmd = command;
--- a/src/test/feature/lib/yarn_config.cpp
+++ b/src/test/feature/lib/yarn_config.cpp
@@ -184,7 +184,7 @@ bool YarnConfig::getStandbyRM(string &standbyRM,
 bool YarnConfig::getHARM(const string &RMtype,
                                string &RM,
                                int &port) {
-  if (!isHA()) {
+  if (isHA() <= 0) {
     return false;
   }
   string RMService = "";
--- a/src/test/feature/testlib/test_lib.cpp
+++ b/src/test/feature/testlib/test_lib.cpp
@@ -87,8 +87,10 @@ TEST_F(TestCommonLib, TestHdfsConfig) {
 
   hc.isSafemode();
 
-  hc.getParameterValue("dfs.replication");
+  std::string defaultValue = hc.getParameterValue("dfs.replication");
   hc.setParameterValue("dfs.replication", "1");
+  hc.getParameterValue("dfs.replication");
+  hc.setParameterValue("dfs.replication", defaultValue);
 }
 
 TEST_F(TestCommonLib, TestYarnConfig) {
@@ -120,8 +122,10 @@ TEST_F(TestCommonLib, TestYarnConfig) {
   portList.clear();
   hc.getActiveNodeManagers(hostList, portList);
 
-  hc.getParameterValue("yarn.scheduler.minimum-allocation-mb");
+  std::string defaultValue =hc.getParameterValue("yarn.scheduler.minimum-allocation-mb");
   hc.setParameterValue("yarn.scheduler.minimum-allocation-mb", "1024");
+  defaultValue =hc.getParameterValue("yarn.scheduler.minimum-allocation-mb");
+  hc.setParameterValue("yarn.scheduler.minimum-allocation-mb", defaultValue);
 }
 
 TEST_F(TestCommonLib, TestCommand) {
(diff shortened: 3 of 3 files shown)


src/test/feature/lib/hdfs_config.h
```diff
@@ -0,0 +1,172 @@
...
+     * set parameter value in ./etc/hadoop/hdfs-site.xml according to parameter name
+     * @param parameterName, parameter name which used to set parameter value
+     * @param parameterValue, parameter value which to be set
+     * @return true if succeeded
+     */
+    bool setParameterValue(const std::string &parameterName, const std::string &parameterValue);
+
+  private:
+    /**
+     * @return yarn user
```
xunzhang: TYPO: `return hdfs user`. Actually, I do not suggest that we should add doxygen-style comments of private interfaces.

src/test/feature/lib/hdfs_config.h
```diff
@@ -0,0 +1,172 @@
...
+ */
+class HdfsConfig {
+  public:
+    /**
+     * HdfsConfig constructor
+     */
+    HdfsConfig(): psql(HAWQ_DB, HAWQ_HOST, HAWQ_PORT, HAWQ_USER, HAWQ_PASSWORD) {}
+
+    /**
+     * HdfsConfig desstructor
```
xunzhang: ...typo

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,284 @@
...
+  string confPath=getHadoopHome();
+  if (confPath == "")
+    return false;
+  confPath.append("/etc/hadoop/hdfs-site.xml");
+  hdfsxmlconf.reset(new XmlConfig(confPath));
+  hdfsxmlconf->parse();
+  return true;
+}
+
+bool HdfsConfig::isHA() {
```
xunzhang: [REVIEW-LINK-1] I think the implementation of these functions are problematic. Say if user modified these configuration files without restart HAWQ, the result could be confused. A better way is to get these kinds of status from running HAWQ, is that possible?

src/test/feature/testlib/test_lib.cpp
```diff
@@ -56,6 +58,69 @@ TEST_F(TestCommonLib, TestHawqConfig) {
   return;
 }
 
+TEST_F(TestCommonLib, TestHdfsConfig) {
+  hawq::test::HdfsConfig hc;
+  hc.isHA();
```
xunzhang: As I mentioned in **[REVIEW-LINK-1]**, if you implemented with loading from a configuration file, you should not just invoke the interface. You should load from a test config file and use `EXPECT_EQ`-like macro function to write your tests.

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+bool HdfsConfig::isTruncate() {
+  string cmd = "hadoop fs -truncate";
+  Command c(cmd);
+  string result = c.run().getResultOutput();
+  auto lines = hawq::test::split(result, '\n');
+  if (lines.size() >= 1) {
+    string valueLine = lines[0];
+    int find = valueLine.find("-truncate: Unknown command");
+    if (find < 0)
```
xunzhang: Unify your coding style with parentheses

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  hawqxmlconf.reset(new XmlConfig(confPath));
+  hawqxmlconf->parse();
+  return true;
+}
+
+bool HdfsConfig::LoadFromHdfsConfigFile() {
+  string confPath=getHadoopHome();
+  if (confPath == "")
```
xunzhang: Unify your coding style with parentheses

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,284 @@
...
+    return true;
+  } else {
+    return false;
+  }
+}
+
+bool HdfsConfig::isKerbos() {
+  bool ret = LoadFromHawqConfigFile();
+  if (!ret) {
+    return false;
```
xunzhang: What is the semantics of the interface returning `false`? If it means that "without Kerbos", the implementation is incorrect. You should throw exception or load default configuration instead.

src/test/feature/lib/hdfs_config.h
```diff
@@ -0,0 +1,172 @@
...
+     * whether HDFS is in HA mode
+     * @return true if HDFS is HA
+     */
+    bool isHA();
+
+    /**
+     * whether HDFS is kerbos
+     * @return true if HDFS is kerbos
+     */
+    bool isKerbos();
```
xunzhang: I am not sure kerbos is a short name of kerberos. I have googled that, I think you use a incorrect library name.

src/test/feature/lib/hdfs_config.h
```diff
@@ -0,0 +1,172 @@
...
+     * @return true if getHANamenode succeeded
+     */
+    bool getHANamenode(const std::string &namenodetype,
+                       std::string &namenode,
+                                   int &port);
+
+  private:
+    std::unique_ptr<XmlConfig> hawqxmlconf;
+    std::unique_ptr<XmlConfig> hdfsxmlconf;
+    hawq::test::PSQL psql;
```
xunzhang: You don't need psql object in this class. Remember modifying the construction too. Same problem also exists in `YarnConfig`.

src/test/feature/lib/yarn_config.cpp
```diff
@@ -0,0 +1,267 @@
...
+      port.push_back(std::stoi(hawq::test::trim(hawq::test::split(values, ':')[1])));
+    }
+  }
+}
+
+
+string YarnConfig::getParameterValue(const string &parameterName) {
+  bool ret = LoadFromYarnConfigFile();
+  if (!ret) {
+    return NULL;
```
xunzhang: return type is `string`...

src/test/feature/lib/xml_parser.h
```diff
@@ -17,10 +17,27 @@ class XmlConfig {
...
+  void closeAndSave();
+
   // parse the configuration file
   void parse();
 
   // @param key The key of the configuration item
+  // @param value The updated value
+  // @param save whether save the updated document to disk, if save is false, open() and closeAndSave() should be called additionally
+  // @ return The value of configuration item
+  bool setString(const std::string &key, const std::string &value, const bool &save);
```
xunzhang: With POD types, there is no case using const reference. For example, here we should define as `bool setString(const std::string &key, const std::string &value, bool save);`

xunzhang:
LGTM

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,293 @@
...
+  const char *env = getenv("GPHOME");
+  string confPath = env ? env : "";
+  if (confPath != "") {
+    confPath.append("/etc/hdfs-client.xml");
+  } else {
+    return false;
+  }
+
+  hawqxmlconf.reset(new XmlConfig(confPath));
+  hawqxmlconf->parse();
```
wengyanqing: Your should handle parse fail condition.

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,293 @@
...
+    int find = valueLine.find("-truncate: Unknown command");
+    if (find < 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
+string HdfsConfig::getHadoopHome() {
+  string cmd = "ps -ef|grep hadoop";
```
wengyanqing: I see lots of places have the same logic. 
1. run command
2. get command result by lines
3. process lines result
It's better to extract the common logic into function.

src/test/feature/lib/hdfs_config.cpp
```diff
@@ -0,0 +1,293 @@
...
+  if (!ret) {
+    throw GetHadoopHomeException();
+  }
+
+  return hdfsxmlconf->getString(parameterName);
+}
+
+string HdfsConfig::getParameterValue(const string &parameterName,
+                                     const string &conftype) {
+  if (conftype == "hdfs" || conftype == "HDFS")
```
wengyanqing: It's better to call lower(conftype) == "hdfs" ?

src/test/feature/lib/hdfs_config.h
```diff
@@ -0,0 +1,175 @@
...
+
+namespace hawq {
+namespace test {
+
+/**
+ * HdfsConfig common libray. Get detailed information about HDFS
+ * including checking state of namenodes and datanodes, get parameter value
+ * @author Chunling Wang
+ */
+class HdfsConfig {
```
wengyanqing: Should HdfsConfig be singleton ?

wengyanqing:
LGTM

src/test/feature/lib/hdfs_config.cpp
xunzhang: `hawq::test::lower(string)`

PR #832, opened 2016-08-03
Hawq-970. Provide More Accurate Information When LibYARN Meets an Exception

linwen:
The fix has removed catch(...) from ApplicationMasterProtocol, ApplicationClientProtocol, ContainerManagementProtocol, so when an exception is raised from RPC call(invoke() function), the exception will be thrown to outer, so that libyarn can display the exact exception information instead of "unexpected exception".

Please review. Thanks!

### DIFF
--- a/depends/libyarn/sample/Makefile
+++ b/depends/libyarn/sample/Makefile
@@ -16,8 +16,8 @@
 # under the License.
 #
 CC=gcc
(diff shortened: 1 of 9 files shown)


jiny2:
LGTM +1

xunzhang:
+1

PR #833, opened 2016-08-03
HAWQ-978. Fixed deadlock in signal handler which call asynchronous un…

liming01:
…safe functions elog()

(1) Can not call elog() in signal handler function, because elog() will cascade call some unsafe function. 
(2) When CheckStatementTimeout() called, "canceling statement due to statement timeout" will be printed out, so we don't need to re-print it.


linwen:
+1

wengyanqing:
+1

ictmalili:
LGTM  +1

PR #835, opened 2016-08-04
HAWQ-980. hawq does not handle guc value with space properly

paul-guo-:
This patch includes pg's code change for split_opts (they call pg_split_opts now).


paul-guo-:
@ictmalili @zhangh43

paul-guo-:
@wengyanqing

src/backend/cdb/executormgr.c
```diff
@@ -864,29 +864,42 @@ addOneOption(PQExpBufferData *buffer, struct config_generic * guc)
...
 				const char *str = *sguc->variable;
-				int			j,
-							start,
-							end;
-				char		temp[1024];
+				unsigned int	 j, start, size;
+				char			*temp;
 
-				end = strlen(str);
+				size = 256;
```
ictmalili: Shall we extract 256 as a macro? I think it's magic number here.

src/backend/utils/misc/guc.c
```diff
@@ -12216,7 +12216,36 @@ ProcessGUCArray(ArrayType *array, GucSource source)
 		 * GPSQL needs to dispatch the database/user config to segments.
 		 */
 		if (Gp_role == GP_ROLE_DISPATCH)
-			appendStringInfo(&MyProcPort->override_options, "-c %s=%s ", name, value);
+		{
```
wengyanqing: It's better to make a function to handle this common logic which process the special char.

src/backend/cdb/executormgr.c
```diff
@@ -864,29 +864,42 @@ addOneOption(PQExpBufferData *buffer, struct config_generic * guc)
...
-				for (start = 0; start < end; ++start)
+				for (start = 0; start < strlen(str); ++start)
 				{
-					if (str[start] == ' ')
-						continue;
+					if (j == size)
+					{
+						size *= 2;
+						temp = realloc(temp, size + 8);
+						if (temp == NULL)
```
ictmalili: What if realloc fail? Will the memory leak?

src/backend/utils/misc/guc.c
```diff
@@ -12216,7 +12216,36 @@ ProcessGUCArray(ArrayType *array, GucSource source)
...
+			char			*temp;
+
+			size = 256;
+			temp = guc_malloc(ERROR, size + 8);
+
+			j = 0;
+			for (start = 0; start < strlen(value); ++start)
+			{
+				if (j == size)
+				{
```
ictmalili: I think this part of code has kind of duplication with the code segment in executormgr.c?? Shall we extract this out?

src/backend/cdb/executormgr.c
paul-guo-: 256 does not have special meaning. It is just a rough initial size so that realloc() is seldom called.

src/backend/cdb/executormgr.c
paul-guo-: Yes. Need to take care of.

wengyanqing:
+1

ictmalili:
+1

PR #839, opened 2016-08-05
HAWQ-983. Fix bug that minirepro generates wrong dependency order of objects

hsyuan:
This patch changes gp_dump_query_oids by traversing the parsed query
tree instead of traversing the query tree struct, which has too
many node type and corner cases to consider. Even it is a little bit
risky to traverse the pased query tree string, we haven't seen any
sign that postgresql upstream is going to change the format.

Add 2 options: relation-oids and function-oids into pg_dump command
line tool to help minirepro dump all the objects in correct order.

In addition, we also fix a minirepro python script bug when column
stats has text type most common values that containing single quote,
minirepro failed to escape that text, which causes sql grammar error
and can't insert that statistics.

Update minirepro to deal with error messages correctly, set PGUSER
as default user, let output file accept relative path.

### DIFF
--- a/src/bin/gpoptutils/gpoptutils.c
+++ b/src/bin/gpoptutils/gpoptutils.c
@@ -26,104 +26,85 @@
  *	LANGUAGE C STRICT;
  */
 
+#include "postgres_fe.h"
 #include "postgres.h"
 #include "funcapi.h"
 #include "utils/builtins.h"
-#include "gpopt/utils/nodeutils.h"
 #include "rewrite/rewriteHandler.h"
-#include "c.h"
+#include "tcop/tcopprot.h"
 
-extern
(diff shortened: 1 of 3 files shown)


hsyuan:
@changleicn @liming01 Please take a look.

changleicn:
LGTM

hsyuan:
ping, will anyone work on merging this pull request?

hsyuan:
@d @oarap please take a look. This is ported from gpdb.

oarap:
LGTM

hsyuan:
Pushed to master:
https://github.com/apache/incubator-hawq/commit/19124b7da032b61d9215520483b67a109246e363
https://github.com/apache/incubator-hawq/commit/08a1c6c2af705df850b1b9bbd6188d0cd2036b42

PR #843, opened 2016-08-10
HAWQ-914. Improve user experience of HAWQ's build infrastructure

paul-guo-:
It is an endless work. This patch adds more output info for various
header/library check failures. It also makes some output info more
platform/release neutral.


huor:
+1

radarwave:
Cool +1

PR #850, opened 2016-08-16
HAWQ-980. hawq does not handle guc value with space properly

paul-guo-:
Remove and replace some thread-unsafe functions in addOneOption().


wengyanqing:
+1

ictmalili:
+1

paul-guo-:
Merged.

PR #851, opened 2016-08-16
HAWQ-838. Replace python module paramiko with pexpect

radarwave:
Since paramiko's license is conflict with Apache HAWQ's license, so we can't deliver it with HAWQ package. 

We can use pexpect(and it's submodule ptyprocess) to replace it which license is compatible with HAWQ. So user don't need to install it manually.


xunzhang:
+1

huor:
+1

PR #853, opened 2016-08-19
HAWQ-967. Extend Projection info to include filter attributes.

sansanichfb:


GodenYao:
+1

src/backend/access/external/pxffilters.c
```diff
@@ -451,6 +452,51 @@ opexpr_to_pxffilter(OpExpr *expr, PxfFilterDesc *filter)
 	return false;
 }
 
+static List*
+append_attr_from_var(Var* var, List* attrs)
+{
+	AttrNumber varattno = var->varattno;
+	/* system attr not supported */
+	if (varattno > InvalidAttrNumber)
```
kavinderd: So the `varattno` is only valid when greater than `InvalidAttrNumber`?

kavinderd:
Nice work! LGTM

src/backend/access/external/pxffilters.c
sansanichfb: @kavinderd, yes, physical attributes should have positive value, some system attributes have -1, and we are not projecting them.

PR #859, opened 2016-08-25
HAWQ-1011. Check whether the table to be registered is existed.

xunzhang:


xunzhang:
cc @ictmalili @huor

tools/bin/hawqregister
```diff
@@ -131,6 +130,18 @@ def option_parser_yml(yml_file):
 
 def create_table(dburl, tablename, schema_info, fmt, distrbution_policy, file_locations):
     try:
+        query = "select count(*) from pg_class where relname = '%s'" % tablename.split('.')[-1]
```
ictmalili: Maybe we need to consider the upper/lower cases of tablename.  All the table name stored in HAWQ are converted to lower case. So I think we need to convert tablename to lower case and then do comparison.

xunzhang:
also cc @radarwave

radarwave:
+1

ictmalili:
LGTM. +1

xunzhang:
Merged into master.

PR #870, opened 2016-08-28
HAWQ-1028. Add '-d' option for hawq state

radarwave:
This option is not used at all, just add it to be compatible with Ambari since Ambari is using this option.

After Ambari removed it, we will remove it again.


xunzhang:
+1

huor:
+1

PR #872, opened 2016-08-29
HAWQ-1030. Forward PostgreSQL spin lock peformance enhancement to HAWQ

liming01:
Hawq didn't change the implementation of spin lock which is same with postgresql, thinking that porting from new version of postgresql in the future, here we keep the code more similar to latest postgresql (version 9.6).

### DIFF
--- a/src/backend/storage/lmgr/s_lock.c
+++ b/src/backend/storage/lmgr/s_lock.c
@@ -3,13 +3,45 @@
  * s_lock.c
  *	   Hardware-dependent implementation of spinlocks.
  *
+ * When waiting for a contended spinlock we loop tightly for awhile, then
+ * delay using pg_usleep() and try again.  Preferably, "awhile" should be a
+ * small multiple of the maximum time we expect a spinlock to be held.  100
+ * iterations seems about right as an initial guess.  However, on a
+ * uniprocessor the loop is a waste of cycles, while in a multi-CPU scenario
+ * it's usually better to spin a bit longer than to call the kernel, so we try
+ * to adapt the spin loop count depending on whether we seem to be in a
+ * uniprocessor or multiprocessor.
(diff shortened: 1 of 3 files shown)


paul-guo-:
The patch is beneficial generally. hawq should include it.

I'm just curious about the upstream implementation of spin lock. This is a bit not related to this patch, but I'd express my option here.

1) pg_usleep() uses the select() way.

Not sure whether the comment still applies to some latest HW and SW, but
isn't a posix delay syscall nanosleep(), which is with high resoluation better?

+/*
+ * pg_usleep --- delay the specified number of microseconds.
+ *
+ * NOTE: although the delay is specified in microseconds, the effective
+ * resolution is only 1/HZ, or 10 milliseconds, on most Unixen.  Expect
+ * the requested delay to be rounded up to the next resolution boundary.
+ *
+ * On machines where "long" is 32 bits, the maximum delay is ~2000 seconds.
+ */
+void
+pg_usleep(long microsec)
+{
+       if (microsec > 0)
+       {
+#ifndef WIN32
+               struct timeval delay;
+
+               delay.tv_sec = microsec / 1000000L;
+               delay.tv_usec = microsec % 1000000L;
+               (void) select(0, NULL, NULL, NULL, &delay);
+#else

2) spin lock is used for cases that the resource will be released soon, so I guess a sched_yield() after some backoff pause is probably more suitable for spinlock (probably could be combined with some process priority adjustment) since it will reduce the latency, even with some additional cpu cost. If there are cases that could lock for a bit longer, a blocking lock should be implemented/used instead.

liming01:
Agree, maybe we can post it into posgresql mail list for further discussion.

paul-guo-:
+1

linwen:
+1

PR #873, opened 2016-08-29
HAWQ-992. PXF Hive data type check in Fragmenter too restrictive.

sansanichfb:
Changes in this PR:

* Updated ProtocolData to parse new attributes X-GP-ATTR-TYPEMODX-COUNT, X-GP-ATTR-TYPEMODX-Y;
* Mapped EnumHawqType to DataType;
* Removed HiveInputFormatFragmenter.toHiveType and replaced it with metadata-driven method HiveUtilities.toCompatibleHiveType;


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/ColumnDescriptor.java
```diff
@@ -26,10 +26,11 @@
  */
 public class ColumnDescriptor {
 
-	int gpdbColumnTypeCode;
-    String gpdbColumnName;
-    String gpdbColumnTypeName;
-    int gpdbColumnIndex;
+	int dbColumnTypeCode;
```
kavinderd: Indent

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,68 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+
+        if (!expectedHawqType.getDataType().equals(hawqDataType)) {
+            throw new UnsupportedTypeException("Invalid definition for column " + hawqColumnName
+                                    +  ": expected HAWQ type " + expectedHawqType.getDataType() +
+                    ", actual HAWQ type " + hawqDataType);
+        }
+
+        if ((hawqTypeMods == null || hawqTypeMods.length == 0) && expectedHawqType.isMandatoryModifiers())
+            throw new UnsupportedTypeException("Invalid definition for column " + hawqColumnName +  ": modifiers are mandatory for type " + expectedHawqType.getTypeName());
+
```
GodenYao: In the case of Hive: Char(xxx) with a fixed length to HAWQ: Char , length omitted, default to 1, we should give user information about the length from Hive: Char(xxx), so they can modify the table definition.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,68 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+     * <li>{@code SMALLINT -> smallint (tinyint is converted to smallint)}</li>
+     * <li>{@code BIGINT -> bigint}</li>
+     * <li>{@code TIMESTAMP -> timestamp}</li>
+     * <li>{@code NUMERIC -> decimal}</li>
+     * <li>{@code BYTEA -> binary}</li>
+     * <li>{@code INTERGER -> int}</li>
+     * <li>{@code TEXT -> string}</li>
+     * <li>{@code REAL -> float}</li>
+     * <li>{@code FLOAT8 -> double}</li>
+     * </ul>
```
GodenYao: based on the logic we also support:
`
               case VARCHAR:
 +            case BPCHAR:
 +            case CHAR:
`

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,68 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+                if (hawqTypeMods != null && hawqTypeMods.length > 0) {
+                    Integer[] hiveTypeModifiers = EnumHiveToHawqType
+                            .extractModifiers(hiveType);
+                    for (int i = 0; i < hiveTypeModifiers.length; i++) {
+                        if (hawqTypeMods[i] < hiveTypeModifiers[i])
+                            throw new UnsupportedTypeException(
+                                    "Invalid definition for column " + hawqColumnName
+                                            + ": modifiers are not compatible, "
+                                            + Arrays.toString(hiveTypeModifiers) + ", "
+                                            + Arrays.toString(hawqTypeMods));
```
GodenYao: same as above comment, should tell user the length for modifier , for numeric, it needs to be exactly match I suppose? for VARCHAR, BPCHAR and CHAR, if modifier exists on HAWQ side, it needs to be >= Hive same type modifier.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
```diff
@@ -110,4 +122,68 @@ public static EnumHiveToHawqType getHiveToHawqType(String hiveType) {
...
+
+        for (EnumHiveToHawqType t : values()) {
+            if (t.getHawqType().getDataType().equals(dataType)) {
+                types.add(t);
+            }
+        }
+
+        if (types.size() == 0)
+            throw new UnsupportedTypeException("Unable to find compatible Hive type for given HAWQ's type: " + dataType);
+
```
GodenYao: the error message should be reversed: 

> Unable to find compatible **HAWQ** type for given **Hive**'s type

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
```diff
@@ -29,8 +35,8 @@
  */
 public enum EnumHiveToHawqType {
 
-    TinyintType("tinyint", EnumHawqType.Int2Type),
-    SmallintType("smallint", EnumHawqType.Int2Type),
+    TinyintType("tinyint", EnumHawqType.Int2Type, (byte) 1),
+    SmallintType("smallint", EnumHawqType.Int2Type, (byte) 2),
```
GodenYao: maybe some comments about adding the byte 1/2 here to differentiate the  2 types on hawq side.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
```diff
@@ -43,37 +45,40 @@ public void serialize(EnumHawqType value, JsonGenerator generator,
...
+    Float4Type("float4", DataType.REAL),
+    Float8Type("float8", DataType.FLOAT8),
+    TextType("text", DataType.TEXT),
+    VarcharType("varchar", DataType.VARCHAR, (byte) 1, false),
+    ByteaType("bytea", DataType.BYTEA),
+    DateType("date", DataType.DATE),
+    TimestampType("timestamp", DataType.TIMESTAMP),
+    BoolType("bool", DataType.BOOLEAN),
+    NumericType("numeric", DataType.NUMERIC, (byte) 2, false),
+    BpcharType("bpchar", DataType.BPCHAR, (byte) 1, false),
```
kavinderd: Can you explain why the last parameter is changed to `false`

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
GodenYao: Another thing:
```
     CharType("char", EnumHawqType.BpcharType, "[(,)]"),
```
according to the code above, we convert Hive: char to HAWQ: Bpchar type. In that case, why do we need to check modifier as HAWQ: char (xxx) won't be compared, it's different oid of HAWQ: Bpchar

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
```diff
@@ -43,37 +45,40 @@ public void serialize(EnumHawqType value, JsonGenerator generator,
...
+    //modifier is mandatory for this type because by default it's 1
+    CharType("char", DataType.CHAR, (byte) 1, true);
```
GodenYao: this type is not used in hivetohawq conversion

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
```diff
@@ -110,4 +122,68 @@ public static EnumHiveToHawqType getHiveToHawqType(String hiveType) {
...
+     */
+    public static EnumHiveToHawqType getCompatibleHawqToHiveType(DataType dataType) {
+
+        SortedSet<EnumHiveToHawqType> types = new TreeSet<EnumHiveToHawqType>(
+                new Comparator<EnumHiveToHawqType>() {
+                    public int compare(EnumHiveToHawqType a,
+                            EnumHiveToHawqType b) {
+                        return Byte.compare(a.getSize(), b.getSize());
+                    }
+                });
```
kavinderd: Can this `types` variable be instantiated outside the method? It doesn't changed based on input right?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
GodenYao: yes, if omitting the modifier, bpchar will be dynamic length

kavinderd:
Overall LGTM, but I'd like to spend more time reviewing. This is a lot of code

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
sansanichfb: Sure, we already have explanation in EnumHiveToHawqType#getSize()

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
sansanichfb: It's actually being populated based on dataType argument.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/ColumnDescriptor.java
sansanichfb: Thanks, fixed

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
sansanichfb: Thanks, updated JavaDoc to reference to EnumHiveToHawqType where those mappings are defined so we don't have to keep docs in two places.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
sansanichfb: deleted mandatory attribute after discussions.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
sansanichfb: Deleted this check after discussions.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
sansanichfb: Deleted

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/io/GPDBWritable.java
```diff
@@ -832,40 +832,10 @@ private boolean isTextForm(int type) {
...
-            case TIME:
-                return "TIME";
-            case TIMESTAMP:
-                return "TIMESTAMP";
-            case NUMERIC:
-                return "NUMERIC";
-            default:
-                return "TEXT";
-        }
+        DataType type = DataType.get(oid);
```
shivzone: Earlier we handled the default case as "TEXT". With this change we will get "UNSUPPORTED_TYPE"

sansanichfb: DataType.get works in such way that it either returns any known data type or UNSUPPORTED_TYPE if oid is invalid. And for UNSUPPORTED_TYPE we still return TEXT.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
```diff
@@ -110,4 +122,68 @@ public static EnumHiveToHawqType getHiveToHawqType(String hiveType) {
...
+            if (t.getTypeName().toLowerCase()
+                    .equals(hiveTypeName.toLowerCase())) {
+                return result;
+            }
+        }
+        throw new UnsupportedTypeException("Unable to map Hive's type: "
+                + hiveType + " to HAWQ's type");
+    }
+
+    /**
```
shivzone: Would be useful if the first line in this comment is added when we do the field declaration in the class (private byte size)

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
```diff
@@ -43,37 +45,32 @@ public void serialize(EnumHawqType value, JsonGenerator generator,
...
+    TextType("text", DataType.TEXT),
+    VarcharType("varchar", DataType.VARCHAR, (byte) 1),
+    ByteaType("bytea", DataType.BYTEA),
+    DateType("date", DataType.DATE),
+    TimestampType("timestamp", DataType.TIMESTAMP),
+    BoolType("bool", DataType.BOOLEAN),
+    NumericType("numeric", DataType.NUMERIC, (byte) 2),
+    BpcharType("bpchar", DataType.BPCHAR, (byte) 1);
 
+    private DataType dataType;
```
shivzone: why do we need both typeName and dataType ?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,61 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+        EnumHiveToHawqType hiveToHawqType = EnumHiveToHawqType.getHiveToHawqType(hiveType);
+        EnumHawqType expectedHawqType = hiveToHawqType.getHawqType();
+
+        if (!expectedHawqType.getDataType().equals(hawqDataType)) {
+            throw new UnsupportedTypeException("Invalid definition for column " + hawqColumnName
+                                    +  ": expected HAWQ type " + expectedHawqType.getDataType() +
+                    ", actual HAWQ type " + hawqDataType);
+        }
+
+        switch (hawqDataType) {
```
shivzone: why do we have cases only for these 3 data types and not the rest ?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,61 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+            case NUMERIC:
+            case VARCHAR:
+            case BPCHAR:
+                if (hawqTypeMods != null && hawqTypeMods.length > 0) {
+                    Integer[] hiveTypeModifiers = EnumHiveToHawqType
+                            .extractModifiers(hiveType);
+                    for (int i = 0; i < hiveTypeModifiers.length; i++) {
```
shivzone: so we would allow hawq data type to have a larger value in the modifier compared to the data type in Hive ?
If this is the case, please add this above in the comments section of this validate function

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
```diff
@@ -256,4 +257,61 @@ private static boolean verifyIntegerModifiers(String[] modifiers) {
...
+                        if (hawqTypeMods[i] < hiveTypeModifiers[i])
+                            throw new UnsupportedTypeException(
+                                    "Invalid definition for column " + hawqColumnName
+                                            + ": modifiers are not compatible, "
+                                            + Arrays.toString(hiveTypeModifiers) + ", "
+                                            + Arrays.toString(hawqTypeMods));
+                    }
+                }
+                break;
```
shivzone: In the older code in compareTypes function we don't enforce the type check for this case convertedHive.equals("smallint") && hiveType.equals("tinyint")) as Hive only supports tinyint. Don't we need to take care of this case ?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
sansanichfb: We are validating modifiers only for types which could have modifiers.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
sansanichfb: Yes, we require HAWQ's type equal or bigger than Hive or no modifiers at all. Will add it to Javadoc.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/HiveUtilities.java
sansanichfb: Right now both cases are valid:

Hive type smallint, HAWQ type smallint, Hive type tinyint, HAWQ type smallint.
They are covered by tests in https://github.com/apache/incubator-hawq/pull/873/files#diff-7d6d87697d9432f93698702c36eda0d3R187

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/EnumHawqType.java
sansanichfb: typeName acts as a user-facing type alias

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
sansanichfb: I also was thinking this way, but it's private field so I've added Javadoc to public getter so it would be available to callers.

shivzone:
+1 great job with refactoring

PR #874, opened 2016-08-29
Remove Thrift 0.9.3 from source tree - HAWQ-959

edespino:
* We will need to update documentation to provide guidance on building
  supporte Thrift version (0.9.3).


paul-guo-:
Hi ed, I guess the concen comes from the binaries only. I'm wondering
whether we could just remove those binaries (probably need to hack a bit)
to workaround the issue? Besides, if we want to remove thrift, we need to
undo the related work in configure/configure.in also (e.g. the
--with-thrift option), probably there are more things to undo (you could
refer to the original patch which adds the thrift support) please
double-check.

2016-08-30 6:19 GMT+08:00 Ed Espino <notifications@github.com>:

xunzhang:
Hi ed, thanks for the pull request! Please check out this [commit](https://github.com/apache/incubator-hawq/commit/c0d7c4fdbebbf6f48be92974a8f055a02ea4c3f2) to make sure the building process passed.

I think it is not a small change doing this, so I am also wondering if these is other solution for HAWQ-959? For example, only remove the binary files in thrift or sync with the upstream thrift(it removed some binary files with latest commits).

edespino:
I chatted with Roman on this issue.  We have conflicting priorities which are causing us to rethink the inclusion of Thrift in the source base.

xunzhang:
LGTM

paul-guo-:
+1

shivzone:
+1

edespino:
This has been merged to master (https://github.com/apache/incubator-hawq/commit/72ea8afd7e17b4777bef9a4f75b8691570cc92c7)

PR #879, opened 2016-08-31
HAWQ-845. Parameterize kerberos principal service name for HAWQ

linwen:
This fix remove the check of "postgres" for kerberos service name if HAWQ is running with kerberos enable.
So that, customers can replace "postgres" with a different service name.
If user want to use a different name, below property/value should be added into hawq-site.xml,
otherwise "postgres" is used. 
<property>
<name>krb_srvname</name>
<value>gpadmin</value>
</property>

Please review, thanks!


ztao1987:
+1

ictmalili:
LGTM. +1

PR #880, opened 2016-09-01
HAWQ-1037. Modify way to get HDFS port in TestHawqRegister

wcl14:


src/test/feature/ManagementTool/test_hawq_register.cpp
```diff
@@ -3,18 +3,36 @@
...
 class TestHawqRegister : public ::testing::Test {
  public:
   TestHawqRegister() {}
   ~TestHawqRegister() {}
+  string getHdfsLocation() {
+    HdfsConfig hc;
+    string hostname = "";
+    int port = 0;
+    if (hc.isHA()) {
+      hc.getActiveNamenode(hostname, port);
```
ictmalili: Will this work? Shouldn't pass the pointer? If the params are modified inside function getActiveNamenode, I don't think they will be modified after the function exists. Your thought?

wengyanqing:
+1

amyrazz44:
LGTM!

PR #881, opened 2016-09-01
HAWQ-1032. Bucket number of new added partition is not consistent wit…

zhangh43:
…h parent table.
Failure Case
```
set deafult_hash_table_bucket_number = 12;
CREATE TABLE sales3 (id int, date date, amt decimal(10,2))         DISTRIBUTED BY (id)                                                           PARTITION BY RANGE (date)                                                     ( START (date '2008-01-01') INCLUSIVE                                            END (date '2009-01-01') EXCLUSIVE                                             EVERY (INTERVAL '1 day') );

set deafult_hash_table_bucket_number = 16;
ALTER TABLE sales3 ADD PARTITION                                   START (date '2009-03-01') INCLUSIVE                                           END (date '2009-04-01') EXCLUSIVE;
```
The newly added partition with buckcet number 16 is not consistent with parent partition.

### DIFF
--- a/src/backend/cdb/cdbpartition.c
+++ b/src/backend/cdb/cdbpartition.c
@@ -6381,7 +6381,7 @@ atpxPartAddList(Relation rel,
 								NULL /*catalogname*/, 
 								get_namespace_name(RelationGetNamespace(par_rel)),
 								RelationGetRelationName(par_rel), -1);
-	
+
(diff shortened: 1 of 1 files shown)


ictmalili:
LGTM. +1

wengyanqing:
+1

PR #884, opened 2016-09-02
HAWQ-1032. Set bucket number of child partition to parent partition o…

zhangh43:
…nly when policy type equals to POLICYTYPE_PARTITIONED.

policy type may be POLICYTYPE_ENTRY which should not be applied to reset child partition bucket number.


xunzhang:
LGTM

ztao1987:
+1

PR #886, opened 2016-09-02
HAWQ-1032. Fix bug of set bucket number of child partition to parent …

zhangh43:
…partition only when policy type equals to POLICYTYPE_PARTITIONED.


xunzhang:
+1

wengyanqing:
+1

PR #889, opened 2016-09-07
HAWQ-1007. Add the pgcrypto code into hawq

paul-guo-:


paul-guo-:
Per disscusion, @rvs is going through the Apache crypto process, and we could check in the code at first.

radarwave:
+1

huor:
+1

PR #892, opened 2016-09-07
HAWQ-1007. Add the pgcrypto code into hawq

paul-guo-:
Avoid maven license check for the pgcrypto code since it comes from upstream postgres.


wengyanqing:
+1

radarwave:
+1

PR #900, opened 2016-09-13
HAWQ-1033. Combine update and insert into one transaction, fix hadoop warning.

xunzhang:


xunzhang:
cc @ictmalili  @zhangh43 @linwen

xunzhang:
also cc @wcl14

ictmalili:
LGTM. +1

tools/bin/hawqregister
```diff
@@ -460,12 +460,13 @@ if __name__ == '__main__':
                 do_not_move, files_update, sizes_update = True, files, sizes
                 files, sizes = [], []
         else:
-            for k, f in enumerate(files):
+            files_old, sizes_old = [f for f in files], [sz for sz in sizes]
```
xunzhang: Notice here, alternative modification like `files_old, sizes_old = files, sizes` is wrong since python default copy is not a deep copy, it is just a pointer copy.

tools/bin/hawqregister
```diff
@@ -460,12 +460,13 @@ if __name__ == '__main__':
                 do_not_move, files_update, sizes_update = True, files, sizes
                 files, sizes = [], []
         else:
-            for k, f in enumerate(files):
```
xunzhang: The old code here is wrong: it is risky to iterate a changing object.

tools/bin/hawqregister
```diff
@@ -373,8 +402,8 @@ if __name__ == '__main__':
     if len(args) != 1 or ((options.yml_config or options.force or options.repair) and options.filepath) or (options.force and options.repair):
         parser.print_help(sys.stderr)
         sys.exit(1)
-    if local_ssh('hadoop', logger):
-        logger.error('command "hadoop" is not available.')
+    if local_ssh('hdfs', logger):
+        logger.error('command "hdfs" is not available.')
```
paul-guo-: Maybe you could be more accurate (e.g. "Maybe the hdfs path is not in the shell environment variable PATH" or similar).

wcl14:
LGTM.

xunzhang:
Merged into master.

tools/bin/hawqregister
xunzhang: We do not try to predict the problem in user side, 
here I only want to return to users the reason why register can not work. Thanks anyway for the volunteer review.

PR #903, opened 2016-09-16
HAWQ-1056 - improve hawq check command line help

lisakowen:
- add several missing options
- update content for accuracy and clarity

fixes the command line help part of HAWQ-1056.

### DIFF
--- a/tools/doc/gpcheck_help
+++ b/tools/doc/gpcheck_help
@@ -58,23 +58,23 @@ An example is shown below:
 OPTIONS
 *****************************************************
 
---config config_file
+--config <config_file>
 
  The name of a configuration file to use instead of the default 
  file $GPHOME/etc/hawq_check.cnf.
 
--f hostfile_hawq_check
+-f <hostfile_hawq_check>
 
  The name of a file that contains a list of hosts that hawq check 
  uses to validate platform-specific settings. This file should 
  contain a single host name for all hosts in your HAWQ system 
  (master, standby master, and segments).
 
--h/--host hostname
+-h/--host <hostname>
  Specifies a single host on which platform-specific settings will
  be validated.
 
---hadoop/--hadoop-home hadoop_home
+--hadoop/--hadoop-home <hadoop_home>
 
  Use this option to specify the full path to your hadoop installation
  location so that hawq check can validate HDFS settings. This option
@@ -121,7 +121,7 @@ OPTIONS
  performed.
 
 
---zipin hawq_check_zipfile
+--zipin <hawq_check_zipfile>
 
  Use this option to decompress and check a .zip file created with 
  the --zipout option. hawq check performs validation tasks against 
@@ -142,7 +142,7 @@ OPTIONS
 EXAMPLES
 *****************************************************
 
-Verify and validate the HAWQ platform settings by  specifying
+Verify and validate the HAWQ platform settings by specifying
 a host file and identifying the full hadoop install path:
 
  # hawq check -f hostfile_hawq_check --hadoop /usr/local/hadoop-<version>/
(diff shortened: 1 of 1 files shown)


edespino:
@lisakowen - Do you still need a review for this PR?  The PR and Jira have been open for a while without any activity (Sep 2016).

lisakowen:
yes, need a review for this PR.

wasn't sure how to rerun the tests - i think the travis failure is
transient/unrelated to the changes.  also wasn't sure how to determine if
there were any tests that compared command line help output, and where i
might look to figure that out.

thanks ed.

paul-guo-:
@radarwave

tools/doc/gpcheck_help
```diff
@@ -135,27 +142,27 @@ OPTIONS
...
- # hawq check -f hostfile_hawq_check --hadoop ~/hadoop-2.0.0/
+ # hawq check -f hostfile_hawq_check --hadoop /usr/hdp/<version>/hadoop/
 
 
 Verify and validate the HAWQ platform settings with HDFS HA
-enabled, YARN HA enabled and Kerberos enabled:
+enabled, YARN HA enabled, and Kerberos enabled:
 
- # hawq check -f hostfile_hawq_check --hadoop ~/hadoop-2.0.0/
+ # hawq check -f hostfile_hawq_check --hadoop /usr/local/hadoop-<version>/
```
radarwave: We have both '/usr/hdp/<version>/hadoop/' and '/usr/local/hadoop-<version>/'. Maybe we should make them the same.

tools/doc/gpcheck_help
```diff
@@ -135,27 +142,27 @@ OPTIONS
 EXAMPLES
 *****************************************************
 
-Verify and validate the HAWQ platform settings by 
-entering a host file and specifying the hadoop location:
+Verify and validate the HAWQ platform settings by  specifying
+a host file and identifying the full hadoop install path:
 
- # hawq check -f hostfile_hawq_check --hadoop ~/hadoop-2.0.0/
+ # hawq check -f hostfile_hawq_check --hadoop /usr/hdp/<version>/hadoop/
```
radarwave: We have both '/usr/hdp/<version>/hadoop/' and '/usr/local/hadoop-<version>/'. Maybe we should make them the same.

lisakowen: thanks for reviewing @radarwave.  i used both local and hdp in the directory paths to distinguish between a product hadoop installation and an open source version.  i agree it is kind of confusing.  i will have both examples use local.

edespino:
One thing I would suggest is to be consistent in the use of the "<" and ">" characters used in the synopsis as the detailed option description.  As an example, the string "--config config_file" should instead be "--config <config_file>" to be consistent with the synopsis. BTW: this would make it consistent with the docs as well.

edespino:
If you could refresh this PR with my comment above, I can review one last time and we can get this merged into master.

tools/doc/gpcheck_help
paul-guo-: hdp should probably not be in the open source doc. I think it is fine for the change.

lisakowen:
@edespino - thanks for reviewing.  i added <> around the option arguments  as you suggested.

liming01:
Hi @lisakowen , please add HAWQ-1056 prefix to commit message, so that we can easily track the apache jira and related fix.

If you don't open any apache jira for doc change, then maybe a simple and help prefix can be used, eg. "Doc updated: ... "

lisakowen:
@liming01 @wengyanqing @stanlyxiang - thank you for noting that i did not follow the appropriate PR conventions.  i will make sure to squash commits and include the JIRA number in the commit message in future PRs.

PR #906, opened 2016-09-16
HAWQ-964. Support for OR and NOT Logical Operators

kavinderd:
Signed-off-by: Leslie Chang <hi.lesliechang@gmail.com>

### DIFF
--- a/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/BasicFilter.java
+++ b/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/BasicFilter.java
@@ -1,0 +1,56 @@
+package org.apache.hawq.pxf.api;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+/**
+ * Basic filter provided for cases where the target storage system does not provide it own filter
+ * For example: Hbase storage provides its own filter but for a Writable based record in a
+ * SequenceFile there is no filter provided and so we need to have a default
+ */
+public class BasicFilter {
+    private FilterParser.Operation oper;
+    private FilterParser.ColumnIndex column;
+    private FilterParser.Constant constant;
+
+    /**
+     * Constructs a BasicFilter.
+     *
+     * @param oper the parse operation to perform
+     * @param column the column index
+     * @param constant the constant object
+     */
+    public BasicFilter(FilterParser.Operation oper, FilterParser.ColumnIndex column, FilterParser.Constant constant) {
+        this.oper = oper;
+        this.column = column;
+        this.constant = constant;
+    }
+
+    public FilterParser.Operation getOperation() {
+        return oper;
+    }
+
+    public FilterParser.ColumnIndex getColumn() {
+        return column;
+    }
+
+    public FilterParser.Constant getConstant() {
+        return constant;
+    }
+}
--- a/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
+++ b/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
@@ -51,18 +51,26 @@ public class FilterParser {
     private Stack<Object> operandsStack;
     private FilterBuilder filterBuilder;
 
-    private static Map<Integer, Operation> operatorTranslationMap = initOperatorTransMap();
-
     /** Supported operations by the parser. */
     public enum Operation {
+        NOOP,
         HDOP_LT,
         HDOP_GT,
         HDOP_LE,
         HDOP_GE,
         HDOP_EQ,
         HDOP_NE,
-        HDOP_AND,
         HDOP_LIKE
+    }
+
+    /**
+     * This enum was added to support filter pushdown with the logical operators OR and NOT
+     * HAWQ-964
+     */
+    public enum LogicalOperation {
+        HDOP_AND,
+        HDOP_OR,
+        HDOP_NOT
     }
 
     /**
@@ -72,15 +80,36 @@ public class FilterParser {
      */
     public interface FilterBuilder {
         /**
-         * Builds the filter.
+         * Builds the filter for an operation
          *
-         * @param operation the parse operation to perform
+         * @param operation the parsed operation to perform
          * @param left the left operand
          * @param right the right operand
          * @return the built filter
          * @throws Exception if building the filter failed
          */
         public Object build(Operation operation, Object left, Object right) throws Exception;
+
+        /**
+         * Builds the filter for a logical operation and two operands
+         *
+         * @param operation the parsed logical operation to perform
+         * @param left the left operand
+         * @param right the right operand
+         * @return the built filter
+         * @throws Exception if building the filter failed
+         */
+        public Object build(LogicalOperation operation, Object left, Object right) throws Exception;
+
+        /**
+         * Builds the filter for a logical operation and one operand
+         *
+         * @param operation the parsed unary logical operation to perform
+         * @param filter the single operand
+         * @return the built filter
+         * @throws Exception if building the filter failed
+         */
+        public Object build(LogicalOperation operation, Object filter) throws Exception;
     }
 
     /** Represents a column index. */
@@ -105,42 +134,6 @@ public class FilterParser {
         }
 
         public Object constant() {
-            return constant;
-        }
-    }
-
-    /**
-     * Basic filter provided for cases where the target storage system does not provide it own filter
-     * For example: Hbase storage provides its own filter but for a Writable based record in a
-     * SequenceFile there is no filter provided and so we need to have a default
-     */
-    static public class BasicFilter {
-        private Operation oper;
-        private ColumnIndex column;
-        private Constant constant;
-
-        /**
-         * Constructs a BasicFilter.
-         *
-         * @param oper the parse operation to perform
-         * @param column the column index
-         * @param constant the constant object
-         */
-        public BasicFilter(Operation oper, ColumnIndex column, Constant constant) {
-            this.oper = oper;
-            this.column = column;
-            this.constant = constant;
-        }
-
-        public Operation getOperation() {
-            return oper;
-        }
-
-        public ColumnIndex getColumn() {
-            return column;
-        }
-
-        public Constant getConstant() {
             return constant;
         }
     }
@@ -175,6 +168,7 @@ public class FilterParser {
     public Object parse(String filter) throws Exception {
         index = 0;
         filterString = filter;
+        int opNumber;
 
         if (filter == null) {
             throw new FilterStringSyntaxException("filter parsing ended with no result");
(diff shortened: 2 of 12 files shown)


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -396,4 +387,12 @@ private Operation reverseOp(Operation operation) {
         operatorTranslationMap.put(8, Operation.HDOP_LIKE);
         return operatorTranslationMap;
     }
+
+    static private Map<Integer, LogicalOperation> initLogicalOperatorTransMap() {
+        Map<Integer, LogicalOperation> integerLogicalOperationMap = new HashMap<>();
```
denalex: array will be a better structure for faster access of lookup of integer from 0..N to an object.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -81,6 +88,8 @@
          * @throws Exception if building the filter failed
          */
         public Object build(Operation operation, Object left, Object right) throws Exception;
+        public Object build(LogicalOperation operation, Object left, Object right) throws Exception;
+        public Object build(LogicalOperation operation, Object filter) throws Exception;
```
denalex: This is confusing -- a LogicalOperation can be assumed a type of Operation, but it is not -- a different Enum and no inheritance. Why do we need such a distinction ? 

If we really need to keep track of whether an operation is logical or not, have a boolean flag on the enum value.

If we consolidate, we can have just one Enum, with numbers for mapping and a flag for logical (if needed). This will reduce number of functions / enums and code overall.

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -154,6 +160,12 @@ private void initOperatorsMap() {
         operatorsMap.put(FilterParser.Operation.HDOP_NE, CompareFilter.CompareOp.NOT_EQUAL); // "!="
     }
 
+    private void initLogicalOperatorsMap() {
+        logicalOperatorsMap = new HashMap<>();
```
denalex: maps where keys are enums should be EnumMap for lookup optimization

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -71,14 +74,19 @@ public HBaseFilterBuilder(HBaseTupleDescription tupleDescription) {
      * @throws Exception if parsing failed
      */
     public Filter getFilterObject(String filterString) throws Exception {
-        FilterParser parser = new FilterParser(this);
-        Object result = parser.parse(filterString);
+        // First check for NOT, HBase does not support this
+        if (filterString.contains(NOT_OP)) {
+            return null;
+        } else {
```
denalex: there is no point in having else branch if you return from the if branch. Removing else branch will simplify the code and make it more readable.

kavinderd: Ok, sounds good

pxf/pxf-hive/src/test/java/org/apache/hawq/pxf/plugins/hive/HiveORCSearchArgumentExample.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.hawq.pxf.api.BasicFilter;
+import org.apache.hawq.pxf.api.LogicalFilter;
+import org.apache.hawq.pxf.api.utilities.ColumnDescriptor;
+import org.junit.Assert;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+public class HiveORCSearchArgumentExample {
```
kavinderd: Sure, will do. I had been debating whether this is just a code example or a useful test

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
kavinderd: That's true. I like that idea. @shivzone any objections to @denalex suggestion?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -220,6 +193,24 @@ public Object parse(String filter) throws Exception {
...
+                    LogicalOperation logicalOperation = logicalOperationTranslationMap.get(safeToInt(parseNumber()));
+
+                    if (logicalOperation == null) {
+                        throw new FilterStringSyntaxException("unknown op ending at " + index);
+                    }
+
+                    if (logicalOperation == LogicalOperation.HDOP_NOT) {
+                        Object exp = operandsStack.pop();
+                        result = filterBuilder.build(logicalOperation, exp);
+                    } else {
```
kavinderd: Ok, I'll add the check

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -405,24 +411,41 @@ private String buildFilterStringForHive() throws Exception {
...
+                prefix = " or ";
+                break;
+            case HDOP_NOT:
+                prefix = " not ";
+                break;
+            default:
+                prefix = "";
+        }
+
+        for (int i = 0; i < filter.getFilterList().size(); i++) {
```
denalex: for (Object f : filter.getFilterList() )

is more compact notation that collapses 2 lines into one

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
sansanichfb: +1 on using enum built-in structure instead of addition integerLogicalOperationMap.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/BasicFilter.java
```diff
@@ -0,0 +1,37 @@
+package org.apache.hawq.pxf.api;
```
sansanichfb: Missing Apache header?

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -71,14 +74,19 @@ public HBaseFilterBuilder(HBaseTupleDescription tupleDescription) {
...
+            FilterParser parser = new FilterParser(this);
+            Object result = parser.parse(filterString);
+
+            if (!(result instanceof Filter)) {
+                throw new Exception("String " + filterString + " resolved to no filter");
```
sansanichfb: Rewrite error message to something like "String couldn't be resolved to any supported filter"?

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -122,25 +130,23 @@ public Filter getFilterObject(String filterString) throws Exception {
...
     }
 
+    @Override
+    public Object build(FilterParser.LogicalOperation opId, Object leftOperand, Object rightOperand) {
+        return handleCompoundOperations(opId, (Filter) leftOperand, (Filter) rightOperand);
+    }
+
+    @Override
+    public Object build(FilterParser.LogicalOperation opId, Object leftOperand) {
+        return null;
```
sansanichfb: Does it mean that we just ignore unary logical operator for HBase?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -205,55 +217,138 @@ private boolean isFiltered(List<HivePartition> partitionFields,
...
-     * condition occurs, we return true. 2. If fieldA does not match any one of
-     * the partition fields we also return true, it means we ignore this filter
-     * because it is not on a partition field. 3. If fieldA = partittionOne and
-     * valueA != valueOne, then we return false.
-     */
-    private boolean testOneFilter(List<HivePartition> partitionFields,
-                                  Object filter, InputData input) {
-        // Let's look first at the filter
-        FilterParser.BasicFilter bFilter = (FilterParser.BasicFilter) filter;
+    private boolean testForUnsupportedOperators(List<Object> filterList) {
```
sansanichfb: Method name is a bit confusing. Could we rename it to something like hasUnsupportedOperator?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveAccessor.java
```diff
@@ -205,55 +217,138 @@ private boolean isFiltered(List<HivePartition> partitionFields,
...
-                return filterValue.equals(partition.val);
-            }
-        }
+                        boolean keepPartition = filterValue.equals(partition.val);
+
+                        /*
+                         * If the string comparison fails then we should check the comparison of
+                         * the two operands as typed values
+                         */
+                        if (!keepPartition && !partition.val.equals("__HIVE_DEFAULT_PARTITION__")){
```
sansanichfb: Make "__HIVE_DEFAULT_PARTITION__" a constant?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -404,24 +411,41 @@ private String buildFilterStringForHive() throws Exception {
...
-            buildSingleFilter(filter, filtersString, prefix);
+            buildSingleFilter(filter, filtersString, "");
         }
 
         return filtersString.toString();
     }
 
+    private void buildCompoundFilter(LogicalFilter filter, StringBuilder filterString) throws Exception{
+        String prefix;
+        switch(filter.getOperator()) {
```
sansanichfb: I would add prefix as a member value to LogicalOperation enum and just invoke getter here.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
kavinderd: I think we'll need to go with an array. I haven't found a way to do a reverse lookup of a Enum based on `int` values. Enum.valueOf() takes a String.

kavinderd: @denalex Thanks. I had stumbled on the same thing

denalex: please review http://stackoverflow.com/questions/5292790/convert-integer-value-to-matching-java-enum for discussion on dangers of relying on ordinal values. I think for us, we might want to maintain explicit number as the private member of enum, then build and cache the array once for all future access.

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -52,12 +52,15 @@
  */
 public class HBaseFilterBuilder implements FilterParser.FilterBuilder {
     private Map<FilterParser.Operation, CompareFilter.CompareOp> operatorsMap;
+    private Map<FilterParser.LogicalOperation, FilterList.Operator> logicalOperatorsMap;
     private byte[] startKey;
     private byte[] endKey;
     private HBaseTupleDescription tupleDescription;
+    private static final String NOT_OP = "l2";
```
kavinderd: Because HBase doesn't support NOT pushdown so we ignore any filters with NOT only in HBase

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
kavinderd: Yeah, HBase doesn't support NOT operations. I don't think this method will ever be invoked in the Hbase case because I already return null when filtering a Hbase filter with 'l2' as an op code

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveDataFragmenter.java
kavinderd: Since LogicalOperation is generic, do you think we should put a very Hive specific string in that enum?

sansanichfb: Makes sense to keep it generic, I agree.

PR #909, opened 2016-09-19
HAWQ-1044. Add normal path testcase for hawq register when file exist…

wcl14:
…s, file does not exist and --force


zhangh43:
+1

PR #910, opened 2016-09-19
HAWQ-1035. Map sub partition table name to the table list in .yml con…

linwen:
Please review. Thanks!

### DIFF
--- a/tools/bin/hawqregister
+++ b/tools/bin/hawqregister
@@ -181,7 +181,7 @@ class GpRegisterAccessor(object):
         rows = self.exec_query(query)
         return rows[0]['attrnums']
 
-    def get_partitioninfo_by_tablename(self, tablename):
+    def get_partition_info(self, tablename):
         ''' Get partition information from pg_partitions, return a constraint-tablename dictionary '''
         query = "SELECT partitiontablename, partitionboundary FROM pg_partitions WHERE tablename = '%s'" % tablename
         return self.exec_query(query)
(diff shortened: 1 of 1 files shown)


tools/bin/hawqregister
```diff
@@ -585,6 +596,20 @@ class HawqRegister(object):
...
+        mappings = []
+        pos = 0
+        for constraint in self.partitions_constraint:
+            if partitions.has_key(constraint):
+                for i in range(0, len(self.partitions_filepaths[pos])):
+                    value = (partitions[constraint], (self.partitions_filepaths[pos][i], self.partitions_filesizes[pos][i]))
+                    mappings.append(value)
+                    i += 1
+            pos += 1
+
```
xunzhang: useless newline

tools/bin/hawqregister
```diff
@@ -585,6 +596,20 @@ class HawqRegister(object):
         query += "end transaction;"
         return self.utility_accessor.update_catalog(query)
 
+    def _mapping_tablename_from_yml(self, partitions):
```
xunzhang: It is a little like c-style python, I recommend to rewrite it as below:

```
def _mapping_tablename_from_yml(self, partitions):
    ''' Mapping table name from yml file, return a list of (table_name,(file_path, file_size)) '''
    for pos, constraint in enumerate(self.partitions_constraint):
        if partitions.has_key(constraint):
            mappings = ([(partitions[constraint], (self.partitions_filepaths[pos][i], self.partitions_filesizes[pos][i]))
                        for i in xrange(len(self.partitions_filepaths[pos]))])
    return mappings
```

linwen: Thank you!

zhangh43:
＋1

PR #913, opened 2016-09-19
HAWQ-1048. Support OR, NOT logical operators in the HAWQ/PXF Bridge.

sansanichfb:


src/backend/access/external/pxffilters.c
```diff
@@ -153,72 +186,102 @@ Oid pxf_supported_types[] =
...
-					result = list_concat(result, inner_result);
+					ExpressionItem *ei = (ExpressionItem *) lfirst(ilc);
+					if (!ei->processed && ei->parent == node)
+					{
+						ei->processed = true;
+						childNodesNum++;
+					}
+				}
+
+				for (int i = 0; i < childNodesNum - 1; i++)
```
shivzone: Why exactly do we need to add children in first level again after we do a list_concat(result, inner_result) ?
I'm not understanding this flow for BoolExpr

PR #915, opened 2016-09-20
HAWQ-1052. Fixed extraction logic for boolean datatype.

sansanichfb:


src/backend/access/external/pxfheaders.c
```diff
@@ -66,8 +66,13 @@ void build_http_header(PxfInputData *input)
 	if (proj_info != NULL && proj_info->pi_isVarList)
 	{
 		List* qualsAttributes = extractPxfAttributes(input->quals);
-
-		add_projection_desc_httpheader(headers, proj_info, qualsAttributes);
+		/* projection information is incomplete if columns from WHERE clause wasn't extracted */
+		if (qualsAttributes !=  NIL || list_length(input->quals) == 0)
```
kavinderd: Why is the check for `list_length(input->quals)` with `0`?

kavinderd:
Aside from my confusion with my comment, +1

src/backend/access/external/pxfheaders.c
sansanichfb: extractPxfAttributes function might return NIL in two cases - when input quals are empty or some error occurred during extraction. So this condition list_length(input->quals) == 0 means no input quals has been passed thus nothing was extracted and it's safe to add projection info.

kavinderd: Ok, got it

PR #917, opened 2016-09-20
HAWQ-1035. Add ddl of partition table for hawq register.

zhangh43:
add create table statement for hawq to register a partition table.
note that, the name of sub partition need to be refined.


paul-guo-:
+1

PR #920, opened 2016-09-20
HAWQ-1063. HAWQ Python library missing import

kdunn926:
Fixing missing import for DatabaseError exception, referenced [here](https://github.com/apache/incubator-hawq/blob/master/tools/bin/hawqpylib/hawqlib.py#L555).


paul-guo-:
+1

edespino:
+1 - I have validated the fix as well. I will be merging this into master.

PR #921, opened 2016-09-20
HAWQ-1064. Additional work related to paramiko removal.

edespino:
The python module paramiko was replaced with pexpect-4.2 &
ptyprocess-0.5.1. The following additional changes are needed:

* LICENSE file needs to be updated
* The references to paramiko need to be removed from tools/bin/Makefile & tools/bin/lib/.gitignore


tools/bin/lib/.gitignore
```diff
@@ -2,7 +2,3 @@ netperf
 netserver
 stream
 Crypto
```
paul-guo-: We seem to be able to remove this file.

edespino: The tools/bin/Makefile processes the "stream" target and the executable "stream" is copied to tools/bin/lib/stream.  Thus the "stream" entry in the .gitignore file is used.

I will remove the other entries and leave the "stream" reference.

radarwave:
Thanks for fix it.
LGTM +1

radarwave:
BTW, could you help to delete pramiko in .travis.yml.

This can save a little time to get travis result. Thanks.

tools/bin/lib/.gitignore
paul-guo-: +1. Good catch for the whole issue.

By the way, tools/Makefile seems to be buggy. See the following patch.
This misleads me also. @ edespino and @radarwave ,
how about the patch below? If yes, Ed you probably could
add this change in you patch?

diff --git a/tools/Makefile b/tools/Makefile
index 172f6f6..3eb49f6 100644
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -27,6 +27,7 @@ include $(top_builddir)/src/Makefile.global

 all:
    ${MAKE} -C gpnetbench $@
+   ${MAKE} -C bin $@

 install :
    ${MAKE} -C bin $@

edespino:
Radar, I will remove paramiko from .travis.yml.  Thanks for the review.

tools/bin/lib/.gitignore
edespino: @paul-guo- I have incorporated your patch into this PR.

paul-guo-: Looks good. Thanks.

PR #922, opened 2016-09-20
HAWQ-1065. Document ppport.h license exception.

edespino:
* Declare PL/Perl's src/pl/plperl/ppport.h source file to be an Perl
  Artistic License exception.


radarwave:
LGTM +1

PR #927, opened 2016-09-22
HAWQ-1068. Fixed crash at calling get_ao_compression_ratio(null)

liming01:


karthijrk:
Quick comment : We were trying to see if we can port this fix to gpdb. In general, by design Postgres / GPDB doesn't check for null in low-level API. Instead, when the arguments get evaluated, it sets the `isNull` flag appropriately. Shouldn't the fix be to see why that flag was not set to true when the argument was evaluated ?

For.eg. if we do select texteq(null, 'dfd') it doesn't crash because first argument's `isNull` set to true and hence it didn't execute `texteq`.

liming01:
@karthijrk, I just fixed get_ao_compression_ratio_name() and get_ao_compression_ratio_oid(), if other functions need to be fixed, please open defect for them. Thanks.

PR #928, opened 2016-09-22
HAWQ-1051. Failing in reverse DNS lookup causes resource manager core…

stanlyxiang:
… dump

### DIFF
--- a/src/backend/resourcemanager/requesthandler.c
+++ b/src/backend/resourcemanager/requesthandler.c
@@ -627,7 +627,11 @@ bool handleRMSEGRequestIMAlive(void **arg)
 	fts_client_ip_len = strlen(fts_client_ip);
 	inet_aton(fts_client_ip, &fts_client_addr);
 	fts_client_host = gethostbyaddr(&fts_client_addr, 4, AF_INET);
-	Assert(fts_client_host != NULL);
+	if (fts_client_host == NULL)
+	{
+		elog(WARNING, "Failed to reverse DNS lookup for ip %s.", fts_client_ip);
+		return true;
+	}
 
 	/* Get the received machine id instance start address. */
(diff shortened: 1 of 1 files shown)


src/backend/resourcemanager/requesthandler.c
```diff
@@ -627,7 +627,11 @@ bool handleRMSEGRequestIMAlive(void **arg)
 	fts_client_ip_len = strlen(fts_client_ip);
 	inet_aton(fts_client_ip, &fts_client_addr);
 	fts_client_host = gethostbyaddr(&fts_client_addr, 4, AF_INET);
-	Assert(fts_client_host != NULL);
+	if (fts_client_host == NULL)
+	{
+		elog(WARNING, "Failed to reverse DNS lookup for ip %s.", fts_client_ip);
+		return true;
+	}
```
paul-guo-: If it is easy to modify we could use the replacement of gethostbyaddr().
By the way, why need to quit after "reverse dns lookup" failure. Should not
an IP address work also?

DESCRIPTION
       The gethostbyname*() and gethostbyaddr*()  functions  are  obsolete.   Applications
       should use getaddrinfo(3) and getnameinfo(3) instead.

stanlyxiang: Not quit after "reverse dns lookup" failure. This function handles the IMAlive message from segment resource manager process.  "return true" just ignore this message and do not continue the followup work after a lookup failure.  IP address is not enough since followup work needs the hostname.

For replace gethostbyaddr(), @jiny2 could you give some comments ?

huor:
It would be great if the commit message can be updated to include both stanlyxiang and jiny2 as author.

xunzhang:
@huor Could you give a commit example like this?

huor:
@xunzhang, here is a tool with that support: https://github.com/git-duet/git-duet. You can also find the example in page home page.

stanlyxiang:
Merged.

PR #929, opened 2016-09-22
HAWQ-1033. Fix bug in hawq register --force.

zhangh43:


tools/bin/hawqregister
```diff
@@ -583,7 +595,7 @@ class HawqRegister(object):
             logger.error("Files to be registered and the table are not in the same hdfs cluster.\nFile(s) to be registered: '%s'\nTable path in HDFS: '%s'." % (filepath, tabledir))
             sys.exit(1)
 
-    def _get_files_in_hdfs(self, filepath):
+    def _get_files_in_hdfs(self, filepath, mode):
```
xunzhang: why don't you use self.mode in this function directly?

zhangh43: fixed

tools/bin/hawqregister
```diff
@@ -469,8 +481,8 @@ class HawqRegister(object):
             self.filepath = self.files[0][:self.files[0].rfind('/')] if self.files else ''
             check_file_not_folder()
             check_database_encoding()
-            if self.mode != 'force' and self.mode != 'repair':
-                if not create_table():
+            if self.mode != 'repair':
+                if not create_table() and self.mode != 'force':
```
xunzhang: A little confused here, I think handle the edge case in _init_mode() could be much more clear. In detail, in this case set self.mode='usage2_table_non_exist' instead of 'force'.

tools/bin/hawqregister
```diff
@@ -168,6 +175,10 @@ class GpRegisterAccessor(object):
     def get_table_existed(self, tablename):
         qry = """select count(*) from pg_class where relname = '%s';""" % tablename.split('.')[-1].lower()
         return self.exec_query(qry)[0]['count'] == 1
+        
+    def get_table_column_num(self, tablename):
```
xunzhang: I like this modification here!

tools/bin/hawqregister
zhangh43: I just want to create table if not exist in --force mode 
And if mode is --force we donot change the it
if not create_table() and self.mode != 'force':
        self.mode = 'usage2_table_exist'

zhangh43: Since table is created by create_table()
I think there is no need to self.mode='usage2_table_non_exist' instead of 'force'

tools/bin/hawqregister
xunzhang: good ❤️

tools/bin/hawqregister
xunzhang: I got.

PR #930, opened 2016-09-22
HAWQ-1061. Ensure filepath is not regex-like.

xunzhang:


xunzhang:
cc @wcl14 @ictmalili

xunzhang:
Merged into master.

PR #935, opened 2016-09-23
HAWQ-1055. Fix to handle decimal with precision,scale

shivzone:


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/utilities/EnumHiveToHawqType.java
```diff
@@ -45,7 +45,7 @@
     BinaryType("binary", EnumHawqType.ByteaType),
     TimestampType("timestamp", EnumHawqType.TimestampType),
     DateType("date", EnumHawqType.DateType),
-    DecimalType("decimal", EnumHawqType.NumericType, "[(,)]"),
+    DecimalType("decimal", EnumHawqType.NumericType, "[(,)]", "(38,18)"),
```
kavinderd: (38,18) is a default value?

PR #939, opened 2016-09-26
HAWQ-991. Fix bug of yaml configuration file contains only files unde…

zhangh43:
…r table directory in --force mode.

### DIFF
--- a/tools/bin/hawqregister
+++ b/tools/bin/hawqregister
@@ -564,7 +564,12 @@ class HawqRegister(object):
                 self.failure_handler.rollback()
                 sys.exit(1)
 
+        if not self.yml:
+            check_no_regex_filepath([self.filepath])
+            self.files, self.sizes = self._get_files_in_hdfs(self.filepath)
+       
         self.do_not_move, self.files_update, self.sizes_update = False, [], []
+        self.newfiles, self.newsizes = [f for f in self.files], [sz for sz in self.sizes]
         if self.mode == 'force':
             existed_files, _ = self._get_files_in_hdfs(self.tabledir)
             if len(self.files) == len(existed_files):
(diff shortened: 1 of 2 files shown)


tools/bin/hawqregister
```diff
@@ -729,6 +728,23 @@ class HawqRegister(object):
                 for k, eof in enumerate(eofs[1:]):
                     query += ',(%d, %d, %d, %d, %d)' % (self.firstsegno + k + 1, eof, -1, -1, -1)
```
kdunn-pivotal: This is minor, but I believe the `%` syntax is being [deprecated](https://www.python.org/dev/peps/pep-3101/) in Python 3.x. Maybe consider replacing it with the new `.format` syntax.

tools/bin/hawqregister
```diff
@@ -603,9 +605,6 @@ class HawqRegister(object):
 
         self._check_files_and_table_in_same_hdfs_cluster(self.filepath, self.tabledir)
 
-        if not self.yml:
-            check_no_regex_filepath([self.filepath])
-            self.files, self.sizes = self._get_files_in_hdfs(self.filepath)
         print 'New file(s) to be registered: ', self.files
```
xunzhang: Here should print `self.newfiles` instead?

tools/bin/hawqregister
xunzhang: Thanks @kdunn-pivotal for the good advice. But I think upgrade from Python2.x to Python3.x is a big deal and since we are not familiar with Python3.x, current code is ok. For the syntax you mentioned here, you are right. TIL

kdunn-pivotal: Yes, upgrade to 3.x is a big deal - I'm definitely *not* proposing that. Rather, we should minimize the technical debt by making net-new code (like this tool) better aligned with 3.x conventions, to avoid revisiting it later on.

PR #945, opened 2016-09-28
HAWQ-991. Fix upper/lower case of tablename, use hadoop fs to avoid hdfs dfs bug in 2.7.1 hadoop.

xunzhang:


xunzhang:
cc @wcl14 @ictmalili

ictmalili:
LGTM. +1

PR #947, opened 2016-09-28
HAWQ-1067. Append hawq version number to plr-hawq rpm pakcage install location

radarwave:
Previously it will get install to /usr/local/hawq, but it's possible that hawq is installed into '/usr/local/hawq_2_0_1_0.
@paul-guo- @xunzhang


paul-guo-:
+1

linwen:
+1

PR #950, opened 2016-10-01
HAWQ-991. Refator HAWQ Register code to make sure partition table register can easy leverage existing code

wcl14:


tools/bin/hawqregister
```diff
@@ -325,250 +326,221 @@ class HawqRegister(object):
...
+            tmp_dict[d['name']] = i + 1
+        # 'DISTRIBUETD BY (1,3)' -> {1,3}
+        cols = self.distribution_policy.strip().split()[-1].strip('(').strip(')').split(',')
+        original_policy = ','.join([str(tmp_dict[col]) for col in cols])
+        if policy.strip('{').strip('}') != original_policy:
+            logger.error('Distribution policy of %s from yaml file is not consistent with the policy of existing table.' % self.dst_table_name)
+            self.failure_handler.rollback()
+            sys.exit(1)
+
+    def set_yml_dataa(self, file_format, files, sizes, tablename, schema, distribution_policy, file_locations,\
```
linwen: Why uses set_yml_dataa not set_yml_data? What does the additional "a" stand for?

tools/bin/hawqregister
```diff
@@ -847,6 +819,33 @@ class HawqRegister(object):
                 self._modify_metadata('insert')
         logger.info('Hawq Register Succeed.')
 
+class HawqRegister_Partition(HawqRegister):
```
linwen: Class name in this file is like this format:AaaBbbCcc, so HawqRegisterPartition is better here.

tools/bin/hawqregister
xunzhang: @linwen , @zhangh43 did this hah!

tools/bin/hawqregister
```diff
@@ -325,250 +326,221 @@ class HawqRegister(object):
...
-            check_policy_consistency()
-            check_no_regex_filepath(self.files)
         else:
-            if is_folder(self.filepath) and self.filesize:
-                logger.error('-e option is only supported with single file case.')
-                sys.exit(1)
             self.file_format = 'Parquet'
-            check_hash_type() # Usage1 only support randomly distributed table
+            self.check_hash_type() # Usage1 only support randomly distributed table
+        self.do_check()
```
xunzhang: The name here is a little confused I think.

Firstly, your L509 use `check_file_not_folder `, L510 `check_database_encoding `, L516 use `check_hash_type`, what's the difference between `check_file_not_folder`, `check_database_encoding`, `check_hash_type` and `do_check`. (I know the reason why you do this, but please refactor with semantic, not just code itself...)

Secondly, `do_check` is an internal interface used in `_init`, which means it can only be invoked in the right place of `_init` function.  I strongly recommend define this kind of function inside `_init`, it can not be invoked independently. Anther reason defining a function variable inside a function is simplify the invoking process for readability.

Thirdly, a internal interface in Python should be named starting with `_` sign.

tools/bin/hawqregister
```diff
@@ -325,250 +326,221 @@ class HawqRegister(object):
...
+        self.partitions_constraint = partitions_constraint
+        self.partitions_name = partitions_name
+        self.partitions_compression_level = partitions_compression_level
+        self.partitions_compression_type = partitions_compression_type
+        self.partitions_checksum = partitions_checksum
+        self.partitions_filepaths = partitions_filepaths
+        self.partitions_filesizes = partitions_filesizes
+        self.encoding = encoding
+
+    def option_parser_yml(self, yml_file):
```
xunzhang: -1 on this. Can you move the function outside? In my view, partition logic is not related to class `HawqRegister`. Besides, can you move `_mapping_tablename_from_yml` logic outside of HawqRegister class?

tools/bin/hawqregister
```diff
@@ -325,250 +326,221 @@ class HawqRegister(object):
...
+                      partitions_compression_type, partitions_checksum, partitions_filepaths, partitions_filesizes, encoding):
+        self.file_format = file_format
+        self.files = files
+        self.sizes = sizes
+        self.src_table_name = tablename
+        self.schema = schema
+        self.distribution_policy = distribution_policy
+        self.file_locations = file_locations
+        self.bucket_number = bucket_number
+        self.partitionby = partitionby
```
xunzhang: Ditto, -1 on this. Please make the logic of class `HawqRegister` clean. Please do not only refactor from coding perspective.

xunzhang:
@ictmalili @linwen , Since this pull is based on not that latest code of `hawqregister`, I do not recommend to merge this kind of pull requests before we finish the automative negative test cases for 2.0.1 `hawqregister` usage.  Otherwise, some of our manually tests effort before is in vain.

ictmalili:
@wcl14 Could you change the description for this Pull Request, since it only covers code refactor part. The partition table support can be processed by another PR. Thanks

linwen:
LGTM

PR #951, opened 2016-10-04
HAWQ-1083

sansanichfb:
Changes in patch:

If url parameter ever has "localhost" string in it when using libchurl, HAWQ replaces "localhost" with actual ip address of any loopback interface(either ipv4 or ipv6).

I think we need loopback criteria to bypass firewall if any.

### DIFF
--- a/src/backend/access/external/libchurl.c
+++ b/src/backend/access/external/libchurl.c
@@ -311,31 +311,7 @@ CHURL_HANDLE churl_init(const char* url, CHURL_HEADERS headers)
 	create_curl_handle(context);
 	clear_error_buffer(context);
 
-	/* needed to resolve localhost */
-	if (strstr(url, LocalhostIpV4) != NULL) {
-		//get loopback interface ip address
-		char* loopback_addr = get_loopback_ip_addr();
-		elog(DEBUG1, "Loopback interface IP address: %s", loopback_addr);
-		//find host start position in url
-		char* start = strstr(url, LocalhostIpV4);
-		//allocate new url with replaced host
-		char* replaced_url = palloc(strlen(url) + strlen(loopback_addr) - strlen(LocalhostIpV4) + 1);
-		//token before host
-		char* before_host = pnstrdup(url, start - url);
-		//token after host
-		char* after_host = pstrdup(url + (start - url) + strlen(LocalhostIpV4));
-		//construct replaced url using loopback interface ip
-		sprintf(replaced_url, "%s%s%s", before_host, loopback_addr, after_host);
-		set_curl_option(context, CURLOPT_URL, replaced_url);
-
-		//release memory
-		pfree(before_host);
-		pfree(after_host);
-		pfree(replaced_url);
-		pfree(loopback_addr);
-	} else
-		set_curl_option(context, CURLOPT_URL, url);
-
+	set_curl_option(context, CURLOPT_URL, url);
 	set_curl_option(context, CURLOPT_VERBOSE, (const void*)FALSE);
 	set_curl_option(context, CURLOPT_ERRORBUFFER, context->curl_error_buffer);
 	set_curl_option(context, CURLOPT_IPRESOLVE, (const void*)CURL_IPRESOLVE_V4);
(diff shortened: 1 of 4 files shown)


src/backend/access/external/pxfutils.c
```diff
@@ -133,3 +137,60 @@ static void process_request(ClientContext* client_context, char *uri)
...
+				}
+				break;
+			}
+		}
+	}
+
+	freeifaddrs(ifaddr);
+
+	if (loopback_addr == NULL)
+		elog(ERROR, "Unable to get loop back address");
```
shivzone: Should we instead just default to using 127.0.0.1 ?

src/backend/access/external/libchurl.c
```diff
@@ -312,16 +313,29 @@ CHURL_HANDLE churl_init(const char* url, CHURL_HEADERS headers)
...
-		strcat(pxf_host_entry, LocalhostIpV4Entry);
-		resolve_hosts = curl_slist_append(NULL, pxf_host_entry);
-		set_curl_option(context, CURLOPT_RESOLVE, resolve_hosts);
-		pfree(pxf_host_entry);
-	}
+		//get loopback interface ip address
+		char* loopback_addr = get_loopback_ip_addr();
+		elog(DEBUG1, "Loopback interface IP address: %s", loopback_addr);
+		//find host start position in url
+		char* start = strstr(url, LocalhostIpV4);
```
kavinderd: Can't this be inlined into the if statement
`if ((start = strstr(url, LocalhostIpV4) != NULL)`

src/backend/access/external/pxfutils.c
```diff
@@ -133,3 +137,60 @@ static void process_request(ClientContext* client_context, char *uri)
...
+			if (s != 0) {
+				elog(WARNING, "Unable to get name information for interface, getnameinfo() failed: %s\n", gai_strerror(s));
+			}
+
+			//get loopback interface
+			if (ifa->ifa_flags & IFF_LOOPBACK) {
+				if (family == AF_INET)
+				{
+					loopback_addr = palloc(strlen(host) + 1);
+					sprintf(loopback_addr, "%s", host);
```
kavinderd: If length of `host` is bigger than `loopback_addr` then there will be a buffer overflow, safer to use `snprintf()`

src/backend/access/external/pxfutils.c
```diff
@@ -133,3 +137,60 @@ static void process_request(ClientContext* client_context, char *uri)
...
+					sprintf(loopback_addr, "[%s]", host);
+				}
+				break;
+			}
+		}
+	}
+
+	freeifaddrs(ifaddr);
+
+	if (loopback_addr == NULL)
```
kavinderd: Maybe also check that `loopback_addr` is not also an empty string

src/backend/access/external/libchurl.c
shivzone: Would be good if we can move below lines till you contract the new url with the loopback_addr to a new function in pxfutils.c named replaceUri(char *uri, char *hostname) and invoke replaceUri(uri, loopback_addr)

src/backend/access/external/libchurl.c
```diff
@@ -312,16 +313,29 @@ CHURL_HANDLE churl_init(const char* url, CHURL_HEADERS headers)
...
+		//allocate new url with replaced host
+		char* replaced_url = palloc(strlen(url) + strlen(loopback_addr) - strlen(LocalhostIpV4) + 1);
+		//token before host
+		char* before_host = pnstrdup(url, start - url);
+		//token after host
+		char* after_host = pstrdup(url + (start - url) + strlen(LocalhostIpV4));
+		//construct replaced url using loopback interface ip
+		sprintf(replaced_url, "%s%s%s", before_host, loopback_addr, after_host);
```
kavinderd: Maybe null check the replaced_url, but it might just be overkill

kavinderd:
+1 aside from comments

src/backend/access/external/libchurl.c
```diff
@@ -312,16 +313,29 @@ CHURL_HANDLE churl_init(const char* url, CHURL_HEADERS headers)
...
+		set_curl_option(context, CURLOPT_URL, replaced_url);
+
+		//release memory
+		pfree(before_host);
+		pfree(after_host);
+		pfree(replaced_url);
+		pfree(loopback_addr);
+	} else
+		set_curl_option(context, CURLOPT_URL, url);
```
denalex: this statement is executed both in if and else branches -- would be cleaner to just overwrite "url" ref in if branch, drop else and have the statement after the block:

url = default
if (problem) {
  url = new_stuff
}
set_option(url)

src/backend/access/external/pxfutils.c
sansanichfb: If function returned null it means there is no loopback interfaces available, and even if we return "127.0.0.1" no interface could handle requests to PXF.

src/backend/access/external/libchurl.c
sansanichfb: Two cases are different, because in case if url contains "localhost" string I am allocating new string with replaced localhost to actual ip address thus I need to release memory in this case. In other case I do not need to free any memory because I haven't allocated any.

src/backend/access/external/libchurl.c
kavinderd: Not double check but move the allocation of `start` into the above `if` statement so we don't need to call the function twice

src/backend/access/external/pxfutils.c
kavinderd: Ok, makes sense. I overlooked the previous line

src/backend/access/external/libchurl.c
sansanichfb: Makes sense.

src/backend/access/external/libchurl.c
sansanichfb: Checked through HAWQ code - haven't found much null checks after pallocating memory. I guess palloc might check for NULL inside.

PR #955, opened 2016-10-10
HAWQ-1035. Support partition table register.

wcl14:


tools/bin/hawqregister
```diff
@@ -422,9 +462,9 @@ class HawqRegister(object):
             Format = 'AO'
         Format_FileLocations = '%s_FileLocations' % Format
         partitionby = params.get(Format_FileLocations).get('PartitionBy')
-        if partitionby:
-            logger.info('Partition table is not supported in current release of hawq register.')
-            sys.exit(0)
+        #if partitionby:
+        #    logger.info('Partition table is not supported in current release of hawq register.')
+        #    sys.exit(0)
```
ictmalili: Let's remove these 3 lines of code when submit.

tools/bin/hawqregister
```diff
@@ -855,6 +895,85 @@ class HawqRegister(object):
...
+        for ele in self.accessor.get_partition_info(self.dst_table_name):
+            dic[ele['partitionboundary']] = ele['partitiontablename']
+        return dic
+
+    def _mapping_tablename_from_yml(self, partitions):
+        ''' Mapping table name from yml file, return a list of (table_name,(file_path, file_size)) '''
+        mappings = []
+        for pos, constraint in enumerate(self.partitions_constraint):
+            if partitions.has_key(constraint):
+                mappings.extend([(partitions[constraint], (self.partitions_filepaths[pos][i], self.partitions_filesizes[pos][i]))
```
ictmalili: Do we have processing mechanism for un-expected error? For example, filepaths[pos][i] has value which filesizes[pos][i] value is invalid?

xunzhang:
https://github.com/apache/incubator-hawq/pull/971#issuecomment-256239846

tools/bin/hawqregister
```diff
@@ -123,6 +123,25 @@ def register_yaml_dict_check(D, table_column_num, src_tablename):
...
+    except yaml.scanner.ScannerError as e:
+        print e
+        system.exit(1)
+
+    if params['FileFormat'].lower() == 'parquet':
+        Format = 'Parquet'
+    else: #AO format
+        Format = 'AO'
+    Format_FileLocations = '%s_FileLocations' % Format
+    partitionby = params.get(Format_FileLocations).get('PartitionBy')
```
xunzhang: What if `params.get(Format_FileLocations) ` is None?

tools/bin/hawqregister
```diff
@@ -855,6 +888,79 @@ class HawqRegister(object):
         logger.info('Hawq Register Succeed.')
 
 
+class HawqRegisterPartition(HawqRegister):
```
xunzhang: Where is the `__init__` function of this class? Use base class 's constructor? I am confused here.

xunzhang: Also, where is the register function of this class? Can we make the interface consistent? For example, 
```
def __init__(self, options, table, utility_conn, conn, failure_handler):
    HawqRegister.__init__(self, options, table, utility_conn, conn, failure_handler)

def register(self):
    HawqRegister.register(self)
```

xunzhang:
BTW, since current implementation separate files into `self.files_update`, `self.files_same_path `, `self.files_append`, it is very tricky and hard to read/maintain. I suggest to add some unit test for some key functions for readability. 

Still I do not think delete all files and insert into ao_seg.# is a good way, it brings complexity for some edge case. If we use this implementation, we should add unit tests.

wcl14:
@linwen @zhangh43

zhangh43:
+1

tools/bin/hawqregister
```diff
@@ -123,6 +123,25 @@ def register_yaml_dict_check(D, table_column_num, src_tablename):
...
         sys.exit(1)
 
+def ispartition(yml_file):
+    import yaml
+    try:
+        with open(yml_file, 'r') as f:
+            params = yaml.load(f)
+    except yaml.scanner.ScannerError as e:
+        print e
+        system.exit(1)
```
xunzhang: Error here...

xunzhang: Can we add some negative cases for partition table registering?

tools/bin/hawqregister
```diff
@@ -939,10 +1082,41 @@ def main(options, args):
...
+        logger.info('try to connect database %s:%s %s' % (url.pghost, url.pgport, url.pgdb))
+        utility_conn = pg.connect(dbname=url.pgdb, host=url.pghost, port=url.pgport,
+                                  user=url.pguser, passwd=url.pgpass, opt='-c gp_session_role=utility')
+        conn = pg.connect(dbname=url.pgdb, host=url.pghost, port=url.pgport,
+                          user=url.pguser, passwd=url.pgpass)
+        return utility_conn, conn
+
+    # connect db
+    try:
+        utility_conn, conn = connectdb(options)
```
xunzhang: where is `options` define, your test function has 0-parameter. Have you ever tried your `test` function locally?

PR #957, opened 2016-10-12
HAWQ-963. PXF support for IS_NULL and IS_NOT_NULL conditions

denalex:


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -63,7 +63,9 @@
         HDOP_EQ,
         HDOP_NE,
         HDOP_AND,
-        HDOP_LIKE
+        HDOP_LIKE,
+        HDOP_ISNULL,
```
sansanichfb: Make them HDOP_IS_NULL, HDOP_IS_NOT_NULL?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -63,7 +63,9 @@
         HDOP_EQ,
         HDOP_NE,
         HDOP_AND,
-        HDOP_LIKE
+        HDOP_LIKE,
```
sansanichfb: Does it makes sense to separate UnaryOperation to new enum?

shivzone: I don't think we need a separate enum for unary operator. We have separate enums currently based on how HAWQ bridge separates a/c/o operators. Also we will soon introduce IN and other operators as well.

PR #959, opened 2016-10-12
HAWQ-1084. Fixed memory allocation for table printing.

sansanichfb:


src/bin/psql/describe.c
```diff
@@ -4316,10 +4316,10 @@ describePxfTable(const char *profile, const char *pattern, bool verbose)
 		if (verbose)
 		{
 			sourcefieldtype = PQgetvalue(res, i, 4);
-			total_fields = PQgetvalue(res, i, 5);
+			total_fields = atoi(PQgetvalue(res, i, 5));
```
kavinderd: Why was an OOM being thrown without `atoi`?

kavinderd: Ah, good catch

kavinderd:
+1

src/bin/psql/describe.c
GodenYao: does atoi() guarantee a positive value?  If there's a chance the char* can be interpreted as negative integer - should we use unsigned int as total_fields is for memory allocation?

PR #965, opened 2016-10-17
HAWQ-1057. Skip SARG_PUSHDOWN parameter if any of filters isn't suppo…

sansanichfb:
…rted.

### DIFF
--- a/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
+++ b/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
@@ -20,6 +20,8 @@ package org.apache.hawq.pxf.plugins.hive;
  */
 
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.hive.ql.io.orc.OrcInputFormat;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgument;
 import org.apache.hadoop.hive.ql.io.sarg.SearchArgumentFactory;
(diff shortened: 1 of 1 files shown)


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
```diff
@@ -131,15 +139,20 @@ private void buildExpression(SearchArgument.Builder builder, List<Object> filter
                         builder.startNot();
                         break;
                 }
-                buildExpression(builder, ((LogicalFilter) f).getFilterList());
-                builder.end();
+                if (buildExpression(builder, ((LogicalFilter) f).getFilterList()))
```
denalex: do we prefer to use { brackets always, even if it is a 1-liner ?

GodenYao:
+1 👍

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
```diff
@@ -131,15 +139,20 @@ private void buildExpression(SearchArgument.Builder builder, List<Object> filter
...
+                    builder.end();
+                else
+                    return false;
             } else {
-                buildArgument(builder, f);
+                if (!buildArgument(builder, f))
```
shivzone: instead just return buildArgument(builder, f)

sansanichfb: We need to return false only if buildArgument(builder, f) call returned false, in case of true we want to proceed.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
sansanichfb: good catch, will add figure brackets.

PR #967, opened 2016-10-19
HAWQ-1044. Enhance hawq register test cases for normal mode and force…

wcl14:
… mode


xunzhang:
also cc @linwen @ictmalili @zhangh43

ictmalili:
LGTM. +1
Huge work here! Thanks @wcl14

PR #968, opened 2016-10-19
HAWQ 1116

shivzone:


src/backend/access/external/test/pxffilters_test.c
```diff
@@ -323,6 +323,30 @@ OpExpr* build_op_expr(void* left, void* right, int op)
...
+	NullTest *expr = (NullTest*) palloc0(sizeof(NullTest));
+	expr->nulltesttype = nullType;
+	expr->arg = arg;
+
+	expr->xpr.type = T_NullTest;
+	return expr;
+}
+
+ExpressionItem* build_null_expression_item(int attnum, Oid attrtype, NullTestType nullType) {
+
```
kavinderd: style: the '{' should be on a new line

src/backend/access/external/test/pxffilters_test.c
```diff
@@ -507,6 +542,23 @@ void test__pxf_serialize_filter_list__oneFilter(void **state) {
 
 }
 
+void test__pxf_serialize_fillter_list__nullFilter(void **state) {
+
```
kavinderd: brace on newline

src/backend/access/external/pxffilters.c
```diff
@@ -406,6 +406,22 @@ pxf_serialize_filter_list(List *expressionItems)
 				appendStringInfo(resbuf, "%c%d", PXF_LOGICAL_OPERATOR_CODE, boolType);
 				break;
 			}
+			case T_NullTest:
+			{
+				elog(DEBUG1, "pxf_serialize_filter_list: node tag %d (T_NullTest)", tag);
+				NullTest *expr = (NullTest *) node;
+
+				/* filter expression for T_NullTest will not have any constant value */
+				if (expr->nulltesttype == IS_NULL)
```
kavinderd: Maybe put this into a conditional statement instead of an if/else block. But it's fine as is too

```
op = expr->nulltestype == IS_NULL ? PXFOP_IS_NULL :  PXFOP_IS_NOTNULL;
appendStringInfo(resbuf, "%c%d%c%d", PXF_ATTR_CODE, ((Var *) expr->arg)->varattno - 1, PXF_OPERATOR_CODE,  op);
```

denalex: +1 -- 2 lines are better than 8 and there is less repetition

shivzone: good point

src/backend/access/external/pxffilters.c
```diff
@@ -406,6 +406,16 @@ pxf_serialize_filter_list(List *expressionItems)
 				appendStringInfo(resbuf, "%c%d", PXF_LOGICAL_OPERATOR_CODE, boolType);
 				break;
 			}
+			case T_NullTest:
+			{
+				elog(DEBUG1, "pxf_serialize_filter_list: node tag %d (T_NullTest)", tag);
+				NullTest *expr = (NullTest *) node;
+
+				/* filter expression for T_NullTest will not have any constant value */
+				PxfOperatorCode o = (expr->nulltesttype == IS_NULL) ? PXFOP_IS_NULL :  PXFOP_IS_NOTNULL;
```
sansanichfb: It would be safer to explicitly check whether expr->nulltesttype == IS_NULL or expr->nulltesttype == IS_NOTNULL instead of relying to default case. As an example in exising codebase: https://github.com/apache/incubator-hawq/blob/master/src/backend/executor/execQual.c#L3828

src/backend/access/external/test/pxffilters_test.c
```diff
@@ -323,10 +323,32 @@ OpExpr* build_op_expr(void* left, void* right, int op)
 	return expr;
 }
 
-ExpressionItem* build_expression_item(int lattnum, Oid lattrtype, char* rconststr, Oid rattrtype, int op) {
+NullTest* build_null_expr(void* arg, NullTestType nullType)
```
sansanichfb: Should we have Expr *arg instead of void* arg?

shivzone: I just wanted the signature to be alligned with build_op_expr that also uses void* for the arguments. I will anyway update it to Expr * to avoid the extra casting

PR #969, opened 2016-10-20
HAWQ-1100. Support for new filter string

kavinderd:
- constant data type is now of the form "c<DATATYPE>s<LENGTH>d<DATA>"

### DIFF
--- a/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
+++ b/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
@@ -34,17 +34,25 @@ import java.util.Stack;
  * interface with two pop-ed operands.
  * <br>
  * A string of filters looks like this:
- * <code>a2c5o1a1c"abc"o2o7</code>
+ * <code>a2c23s1d5o1a1c25s3dabco2o7</code>
  * which means {@code column#2 < 5 AND column#1 > "abc"}
  * <br>
  * It is a RPN serialized representation of a filters tree in GPDB where
  * <ul>
  * <li> a means an attribute (column)</li>
- * <li>c means a constant (either string or numeric)</li>
+ * <li>c means a constant followed by the datatype oid</li>
+ * <li>s means the length of the data in bytes</li>
+ * <li>d denotes the start of the constant data</li>
  * <li>o means operator</li>
  * </ul>
- *
- * Assuming all operators are binary, RPN representation allows it to be read left to right easily.
+ * <br>
+ * For constants all three portions are required in order to parse the data type, the length of the data in bytes
+ * and the data itself
+ * <br>
+ * The parsing operation parses each element of the filter (constants, columns, operations) and adds them to a stack.
+ * When the parser sees an op code 'o' or 'l' it pops off two elements from the stack assigns them as children of the op
+ * and pushses itself onto the stack. After parsing is complete there should only be one element in the stack, the root
+ * node of the filter's tree representation which is returned from this method
  * <br>
  * FilterParser only knows about columns and constants. The rest is up to the {@link FilterBuilder} implementer.
  * FilterParser makes sure column objects are always on the left of the expression (when relevant).
--- a/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
+++ b/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
@@ -23,6 +23,7 @@ package org.apache.hawq.pxf.api;
 import org.apache.hawq.pxf.api.io.DataType;
 
 import java.sql.Date;
+import java.sql.Time;
 import java.sql.Timestamp;
 import java.util.Arrays;
 import java.util.Stack;
@@ -53,6 +54,12 @@ public class FilterParser {
     private byte[] filterByteArr;
     private Stack<Object> operandsStack;
     private FilterBuilder filterBuilder;
+    public static final char COL_OP = 'a';
+    public static final char CONST_OP = 'c';
+    public static final char CONST_LEN = 's';
+    public static final char CONST_DATA = 'd';
+    public static final char COMP_OP = 'o';
+    public static final char LOG_OP = 'l';
 
     /** Supported operations by the parser. */
     public enum Operation {
@@ -181,13 +188,13 @@ public class FilterParser {
             char op = (char) filterByteArr[index];
             ++index; // skip op character
             switch (op) {
-                case 'a':
+                case COL_OP:
                     operandsStack.push(new ColumnIndex(safeToInt(parseNumber())));
                     break;
-                case 'c':
+                case CONST_OP:
                     operandsStack.push(new Constant(parseParameter()));
                     break;
-                case 'o':
+                case COMP_OP:
                     opNumber = safeToInt(parseNumber());
                     Operation operation = opNumber < Operation.values().length ? Operation.values()[opNumber] : null;
                     if (operation == null) {
@@ -222,7 +229,7 @@ public class FilterParser {
                     operandsStack.push(result);
                     break;
                 // Handle parsing logical operator (HAWQ-964)
-                case 'l':
+                case LOG_OP:
                     opNumber = safeToInt(parseNumber());
                     LogicalOperation logicalOperation = opNumber < LogicalOperation.values().length ? LogicalOperation.values()[opNumber] : null;
 
@@ -297,7 +304,7 @@ public class FilterParser {
     }
 
     private int parseDataLength() throws Exception {
-        if (((char) filterByteArr[index]) != 's') {
+        if (((char) filterByteArr[index]) != CONST_LEN) {
             throw new FilterStringSyntaxException("data length delimiter 's' expected at " +  index);
         }
 
@@ -343,6 +350,10 @@ public class FilterParser {
                     return Date.valueOf(data);
                 case TIMESTAMP:
                     return Timestamp.valueOf(data);
+                case TIME:
+                    return Time.valueOf(data);
+                case BYTEA:
+                    return data.getBytes();
                 default:
                     throw new FilterStringSyntaxException("DataType " + dataType.toString() + " unsupported");
             }
@@ -371,7 +382,7 @@ public class FilterParser {
             throw new FilterStringSyntaxException("data size larger than filter string starting at " + index);
         }
 
-        if (((char) filterByteArr[index]) != 'd') {
+        if (((char) filterByteArr[index]) != CONST_DATA) {
             throw new FilterStringSyntaxException("data delimiter 'd' expected at " + index);
         }
 
--- a/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
+++ b/pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
@@ -32,7 +32,10 @@ import java.util.Stack;
  * Once an operation is read, the evaluate function is called for the {@link FilterBuilder}
  * interface with two pop-ed operands.
  * <br>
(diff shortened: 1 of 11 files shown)


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -279,25 +282,108 @@ int safeToInt(Long value) throws FilterStringSyntaxException {
...
+            return Integer.parseInt(digits);
+        } catch (NumberFormatException e) {
+            throw new FilterStringSyntaxException("invalid numeric argument " + digits);
+        }
+    }
+
+    private Object convertDataType(byte[] byteData, DataType dataType) throws Exception {
+        String data = new String(byteData);
+        try {
+            switch (dataType) {
```
sansanichfb: Should we also include TIME and BYTEA types?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -165,17 +168,17 @@ public FilterParser(FilterBuilder eval) {
...
             throw new FilterStringSyntaxException("filter parsing ended with no result");
         }
 
-        while (index < filterString.length()) {
-            char op = filterString.charAt(index);
+        while (index < filterByteArr.length) {
+            char op = (char) filterByteArr[index];
             ++index; // skip op character
             switch (op) {
                 case 'a':
```
sansanichfb: Should we have all conventional characters used in protocol as constants or enum?
'a','c', 's','d','o','l'

pxf/pxf-api/src/test/java/org/apache/hawq/pxf/api/FilterParserTest.java
```diff
@@ -143,32 +167,39 @@ public void parseNegativeBadOperation() {
...
         runParseNegative("filter with only column", filter, "filter parsing failed, missing operators?");
 
-        filter = "c1";
+        filter = "c20s1d1";
         runParseNegative("filter with only numeric const", filter, "filter parsing failed, missing operators?");
+    }
 
-        filter = "c\"something in the way\"";
-        runParseNegative("filter with only string const", filter, "filter parsing failed, missing operators?");
+    @Test
```
sansanichfb: Add test case for negative value?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -279,25 +282,108 @@ int safeToInt(Long value) throws FilterStringSyntaxException {
...
-                ? parseString()
-                : parseNumber();
-    }
+        DataType dataType = DataType.get(parseConstDataType());
+        if (dataType == DataType.UNSUPPORTED_TYPE) {
+            throw new FilterStringSyntaxException("invalid DataType OID at " + (index - 1));
+        }
+
+        int dataLength = parseDataLength();
+        if (dataLength < 1) {
```
shivzone: How would an equality check for "" be handled ? What would the data length be ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
kavinderd: Yeah, an enum is a good idea...I'll add it

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -229,8 +233,20 @@ private ByteArrayComparable getComparator(int type, Object data) throws Exceptio
...
-                result = new HBaseIntegerComparator((Long) data);
+                if (data instanceof Long) {
+                    result = new HBaseIntegerComparator((Long) data);
+                } else if (data instanceof Integer) {
+                    result = new HBaseIntegerComparator(((Integer) data).longValue());
+                } else {
+                    result = null;
+                }
+                break;
+            case FLOAT8:
```
kavinderd: Good point

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -279,25 +282,108 @@ int safeToInt(Long value) throws FilterStringSyntaxException {
...
+
+        try {
+            return Integer.parseInt(digits);
+        } catch (NumberFormatException e) {
+            throw new FilterStringSyntaxException("invalid numeric argument " + digits);
+        }
+    }
+
+    private Object convertDataType(byte[] byteData, DataType dataType) throws Exception {
+        String data = new String(byteData);
```
denalex: creating a string out of bytes assumes a specific charset. how will that be handled ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -318,17 +404,17 @@ private Long parseNumber() throws Exception {
     private String parseDigits() throws Exception {
         String result;
         int i = index;
-        int filterLength = filterString.length();
+        int filterLength = filterByteArr.length;
 
         // allow sign
         if (filterLength > 0) {
-            int chr = filterString.charAt(i);
+            int chr = (char) filterByteArr[i];
```
denalex: if we cast to (char), should the datatype be char as well, not int ? i know it works either way, but for consistency...

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
shivzone: it would use the platform's default charset. @kavinderd is that the interned behavior ?
If not, use "UTF-8"

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -356,8 +442,8 @@ private String parseString() throws Exception {
         int i;
 
         // starting from index + 1 to skip leading "
-        for (i = index + 1; i < filterString.length(); ++i) {
-            char chr = filterString.charAt(i);
+        for (i = index + 1; i < filterByteArr.length; ++i) {
+            char chr = (char) filterByteArr[i];
```
denalex: do we still have quotes around string values ? since we are passing length, it is no longer necessary

pxf/pxf-hbase/src/main/java/org/apache/hawq/pxf/plugins/hbase/HBaseFilterBuilder.java
```diff
@@ -89,12 +90,15 @@ private boolean filterNotOpPresent(String filterString) {
...
+        if (filterString == null)
+            return null;
+
         // First check for NOT, HBase does not support this
         if (filterNotOpPresent(filterString))
             return null;
 
         FilterParser parser = new FilterParser(this);
-        Object result = parser.parse(filterString);
+        Object result = parser.parse(filterString.getBytes());
```
denalex: this is confusing ... we already have a string here, then we convert it to byte[] to send to parser which converts each byte to char ... why ?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
```diff
@@ -162,6 +165,10 @@ private boolean buildArgument(SearchArgument.Builder builder, Object filterObj)
         ColumnDescriptor filterColumn = inputData.getColumn(filterColumnIndex);
         String filterColumnName = filterColumn.columnName();
 
+        /* Need to convert java.sql.Date to Hive's DateWritable Format */
+        if (filterValue instanceof Date)
```
shivzone: Why does the type conversion only need to happen for Date and not for other types ?

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveFilterBuilder.java
```diff
@@ -61,8 +61,11 @@ public HiveFilterBuilder(InputData input) {
      *             filter or list of basic filters
      */
     public Object getFilterObject(String filterString) throws Exception {
+        if (filterString == null)
```
denalex: style : curly brackets missing ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -279,25 +282,108 @@ int safeToInt(Long value) throws FilterStringSyntaxException {
...
+
+        if (((char) filterByteArr[index]) != 'd') {
+            throw new FilterStringSyntaxException("data delimiter 'd' expected at " + index);
+        }
+
+        index++;
 
-    private boolean senseString() {
-        return filterString.charAt(index) == '"';
+        Object data = convertDataType(Arrays.copyOfRange(filterByteArr, index, index+dataLength), dataType);
```
kavinderd: Good catch, I'll update the method

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
shivzone: For the case when TEXT field is checked with = against "", bridge uses s0do5 in the expression string.

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCAccessor.java
kavinderd: Hive doesn't support java.sql.Data so I needed to catch this explicitly and convert to DateWritable

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -279,25 +296,110 @@ int safeToInt(Long value) throws FilterStringSyntaxException {
...
+
+        try {
+            return Integer.parseInt(digits);
+        } catch (NumberFormatException e) {
+            throw new FilterStringSyntaxException("invalid numeric argument " + digits);
+        }
+    }
+
+    private Object convertDataType(byte[] byteData, int start, int end, DataType dataType) throws Exception {
+        String data = new String(byteData, start, end-start);
```
denalex: didn't we want to explicitly specify UTF-8 here, since the data was re-encoded with UTF-8 after being parsed from the header ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/FilterParser.java
```diff
@@ -30,26 +34,40 @@
  * interface with two pop-ed operands.
  * <br>
```
shivzone: Can you add this 
The filter string is of the pattern `<attcode><attnum><constcode><constval><constsizecode><constsize><constdata><constvalue><opercode><opernum>`

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveFilterBuilder.java
```diff
@@ -61,8 +61,11 @@ public HiveFilterBuilder(InputData input) {
...
+            return null;
+
         FilterParser parser = new FilterParser(this);
-        Object result = parser.parse(filterString);
+        Object result = parser.parse(filterString.getBytes());
```
denalex: and here ?

PR #979, opened 2016-10-27
HAWQ-1125. Running pl/python related feature_test cases in parallel f…

paul-guo-:
…ailed with error "cache lookup failed for language"

Looks like an apparent race condition.


wengyanqing:
+1

PR #980, opened 2016-10-27
HAWQ-1113. Fix bug and add test for hawq register force mode.

zhangh43:
force mode should use tabledir(which got from metadata) instead of filepath(extracted from yaml) to get existed file.


src/test/feature/ManagementTool/test_hawq_register_usage2_case2.cpp
```diff
@@ -107,11 +107,17 @@ void TestHawqRegister::runYamlCaseForceMode(std::string casename, std::string ym
     hc.getNamenodeHost(hdfs_prefix);
     strs_src_dst["@PORT@"]= hdfs_prefix;
     frep.replace(t_yml_tpl, t_yml, strs_src_dst);
+    if(samepathfile) {
+        for(int i=8; i<=13; i++){
```
xunzhang: This is cool.

linwen:
+1

PR #981, opened 2016-10-27
HAWQ-1114. Implement filter-push down for IN on HAWQ bridge side.

sansanichfb:


src/backend/access/external/pxffilters.c
```diff
@@ -187,13 +200,18 @@ Oid pxf_supported_types[] =
...
 	BOOLOID,
 	DATEOID,
-	TIMESTAMPOID
+	TIMESTAMPOID,
+	/* complex datatypes*/
+	INT2ARRAYOID,
+	INT4ARRAYOID,
+	INT8ARRAYOID,
+	BOOLARRAYOID,
+	TEXTARRAYOID
```
shivzone: Can you check what ANYARRAYOID means ? If we can support this datatype as well along with the other array datatypes

src/backend/access/external/pxffilters.c
```diff
@@ -357,6 +393,46 @@ pxf_serialize_filter_list(List *expressionItems)
...
+			{
+				elog(DEBUG1, "pxf_serialize_filter_list: node tag %d (T_Var)", tag);
+				PxfFilterDesc *filter = (PxfFilterDesc *) palloc0(sizeof(PxfFilterDesc));
+				Var *var = (Var *) node;
+				if (var_to_pxffilter(var, filter))
+				{
+					PxfOperand l = filter->l;
+					PxfOperand r = filter->r;
+					PxfOperatorCode o = filter->op;
+					if (pxfoperand_is_attr(l) && pxfoperand_is_scalar_const(r))
```
shivzone: Can we have a scenario which is left operand is scalar const and right operand is attirbute ?

shivzone:
Why aren't we adding PXFOP_IN to PxfOperatorCode enum ?
For the other operators, PXF service expects to see the opcode value to correspond to the PxfOperatorCode enum

src/backend/access/external/pxffilters.c
sansanichfb: Left operand constant and right operand is attribute translates in SQL:
WHERE (1,2,3) IN c1 which has incorrect syntax and should be never the case.

sansanichfb:
@shivzone good point about PXFOP_IN, added it. we should distinguis three cases:
1) WHERE scalar_column IN (1,2,3)
2) WHERE array_column = '{1,2,3}'
2) WHERE array_column IN (1,2,3)

src/backend/access/external/pxffilters.c
```diff
@@ -719,6 +1029,143 @@ const_to_str(Const *constval, StringInfo buf)
...
+	ArrayType  *arr;
+	int len;
+
+	if (constval->constisnull)
+	{
+		elog(DEBUG1, "Null constant is not expected in this context.");
+		return;
+	}
+
+	if (constval->constbyval) {
```
shivzone: This check seems a bit unnecessary as we would never get this expression down to our code

sansanichfb: This check is needed because bridge part relies on data passed by planner.

src/backend/access/external/pxffilters.c
```diff
@@ -376,6 +467,46 @@ pxf_serialize_filter_list(List *expressionItems)
 
 		switch (tag)
 		{
+			case T_Var:
```
wenxiaozxw: Which scenario is the expr type is T_Var?
In you comment is IN(single_value),but IN(single_value) is also T_OpExpr(Var + Const)

PR #986, opened 2016-10-31
HAWQ-1034. Delete code for hawq register repair mode

wcl14:


linwen:
+1

xunzhang:
Can you describe the reason why these codes should be deleted in related JIRA? This is apache community, not pivotal.

PR #988, opened 2016-10-31
HAWQ should print error message instead of python function stack when yaml file is invalid.

linwen:
Please review, thanks!


zhangh43:
+1

PR #989, opened 2016-10-31
HAWQ-1129. Install PLR into hawq home directory

radarwave:


wengyanqing:
+1

paul-guo-:
Seems that hawq_version is not needed, so it could be removed safely in Makefile.

PR #991, opened 2016-11-01
HAWQ-1131. Fix error message difference between Orca and Planner

hsyuan:
- In ORCA, due to the way exception handled previously we do warning first
and then later print error referring that message. In this commit, we
enhanced the exception handling so we just print a single error message.
- Removed 'PQO unable to generate a plan' or 'Aborting PQO plan
generation' message and make the error message as close as the planner.
- Updated error message with filename and line number from where the exception
is raised.
- Updated regression test expected output files

Patches are ported the following commits from GPDB:
https://github.com/greenplum-db/gpdb/commit/4eb5db7a6bc8ac3faeca761193dc37e2c70e1ea3
https://github.com/greenplum-db/gpdb/commit/d41ebde141104dd3db6fa29ac46b4096ee7d4825

Author: Karthikeyan Jambu Rajaraman <karthi.jrk@gmail.com>


paul-guo-:
+1

linwen:
+1

hsyuan:
Pushed to master: 
https://github.com/apache/incubator-hawq/commit/56b44f8595b4ba1ac9c49b775d5110e9be1275b2
https://github.com/apache/incubator-hawq/commit/31ea7d08ede7d5bc3108454731ba11bdcf1092d1
https://github.com/apache/incubator-hawq/commit/fed12f1a57ac64a8c7cab867f31b2930dddc2744

PR #994, opened 2016-11-02
HAWQ-1135. MADlib: Raising exception leads to database connection ter…

zhangh43:
…mination.

To solve this bug, we remove AtSubAbort_ActiveQueryResource() when subtransaction abort.

In sub transaction, ActiveQueryResource should not be set NULL.
Sub transaction inherits resource from parent.
Parent transaction may still use ActiveQueryResource to reference the original resource.


xunzhang:
🍻

huor:
Nice fix, Hubert! Is it possible to simplify the reproduction case and make it MADlib independent so that this scenario can be added to hawq regression test?

liming01:
+1

PR #996, opened 2016-11-02
HAWQ-1117.RM crash when init db after configure with param '--enable-cassert'

stanlyxiang:
please review @jiny2  @huor


wengyanqing:
+1

PR #1000, opened 2016-11-03
HAWQ-1143. Libhdfs create semantic is not consistent with posix stand…

zhangh43:
…ard.
Open a file under posix standard, if o_create flag is set to true and the file exists, no side effect except O_EXCL is also be set true.
Open a file in HDFS with hdfs::create flag will report errors if file exists.
In libhdfs, the o_create flag is interpreted to hdfs::create, which leads to errors if file exists no matter O_EXCL is set or not.


linwen:
+1

PR #1002, opened 2016-11-03
HAWQ-1130. Make HCatalog integration work with non-superusers.

sansanichfb:
Changed master_highest_used_oid logic to use CAQL instead of SPI, because some of tables require super-user access(pg_catalog.pg_authid and pg_catalog.pg_user_mapping).

@hornn

### DIFF
--- a/src/backend/access/transam/varsup.c
+++ b/src/backend/access/transam/varsup.c
@@ -16,7 +16,7 @@
 #include "access/clog.h"
 #include "access/subtrans.h"
 #include "access/transam.h"
-#include "executor/spi.h"
+#include "catalog/catquery.h"
 #include "miscadmin.h"
 #include "postmaster/autovacuum.h"
 #include "storage/pmsignal.h"
@@ -416,6 +416,11 @@ GetNewExternalObjectId(void)
 		ResetExternalObjectId();
 	}
 
+	/*
+	 * This check is needed for upgrade from old HAWQ versions, which don't support
+	 * oid pool for HCatalog objects.
+	 * In current implementation max oid will be always less than FirstExternalObjectId.
+	 */
 	if (!IsValidExternalOidRange)
 		ereport(ERROR,
 				(errcode(ERRCODE_INSUFFICIENT_RESOURCES),
@@ -474,7 +479,7 @@ ResetExternalObjectId(void)
 
 /*
  * master_highest_used_oid
- * 		Query the database to find the highest used Oid by
+ * 		Uses CAQL to find the highest used Oid by
  * 		1) Find all the relations that has Oids
  * 		2) Find max oid from those relations
  */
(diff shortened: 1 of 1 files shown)


src/backend/access/transam/varsup.c
```diff
@@ -474,73 +479,53 @@ ResetExternalObjectId(void)
...
-	if (SPI_OK_CONNECT != SPI_connect())
-	{
-		ereport(ERROR, (errcode(ERRCODE_CDB_INTERNAL_ERROR),
-				errmsg("Unable to connect to execute internal query for HCatalog.")));
-	}
-
-	int ret = SPI_execute("SELECT relname FROM pg_class where relhasoids=true", true, 0);
+	cqContext *pcqOuterCtx = caql_beginscan(
+			NULL,
+			cql("SELECT * FROM pg_class where relhasoids = :1",
```
hornn: shouldn't it be `SELECT relname ...` like it was in the previous query?

src/backend/access/transam/varsup.c
```diff
@@ -474,73 +479,53 @@ ResetExternalObjectId(void)
...
 	{
-		if (i > 0)
-		{
-			appendStringInfo(&sqlstr, " UNION ALL ");
-		}
-		appendStringInfo(&sqlstr, "SELECT max(oid) AS oid FROM %s", tableNames[i]);
-	}
-	appendStringInfo(&sqlstr, ") AS x");
+		classForm = (Form_pg_class) GETSTRUCT(tuple);
+		appendStringInfo(sqlstr, "SELECT oid FROM %s WHERE oid >= :1 ORDER BY oid", classForm->relname.data);
```
hornn: why not use `MAX(oid)` like it was before? is it because it's not supported by caql?
Also, maybe add `LIMIT 1` to the query because we only need the first row.

src/backend/access/transam/varsup.c
```diff
@@ -474,73 +479,53 @@ ResetExternalObjectId(void)
...
 
-	ret = SPI_execute(sqlstr.data, true, 1);
+		currentOid = caql_getoid_plus(NULL, &fetchCount, NULL, cql(sqlstr->data, oidMax));
```
hornn: `oidMax` is set to `InvalidOid` before the loop, so the first time this query is run it's actually `WHERE oid >= InvalidOid`. It's probably better not to rely on the actual value of `InvalidOid`, so maybe use a separate variable for the query, set it to -1 or 0 before the loop, and then update it inside the loop to oidMax.

src/backend/access/transam/varsup.c
sansanichfb: Oh, nice, will use just relname, thanks.

src/backend/access/transam/varsup.c
hornn: Also, if the query returns no rows, `currentOid` will be `InvalidOid`. It is probably better to check specifically for this condition and only update `oidMax` if it's a valid result. Something like
```
if (oidMax == InvalidOid)
    oidMax = currentOid
if (currentOid != InvalidOid)
    oidMax = currentOid > oidMax ? currentOid : oidMax;
```
What do you think?
I am also not sure the `WHERE > maxOid` is needed - logically the loop will work fine without it because we only update maxOid if the result is bigger. Would be interesting to see if there is any difference in performance without this condition.

hornn:
@sansanichfb - did you compare performance before and after the change? IIRC this function is used internally every time an hcatalog object is created, so it should be fast.
After this change we'll have 1 + [number of tables with oids] queries, where before we had 2 queries (one with [number of tables with oids] joins), so it is probably good to check that the new approach is not slower.

hornn:
Also forgot to mention - I think there are unittests around this code. You might need to revise them, I am not sure.

src/backend/access/transam/varsup.c
sansanichfb: I found out that caql doesn't support MAX and it supports only ORDER BY, but doesn't support ORDER BY ... DESC so we actually need last row. Will updated code to iterate fetchCount times and get last value.

sansanichfb:
@hornn haven't test performance yet, will test soon. As far as I understood, we invoke this function only one time per session, when importing first relation from Hive. Also before we were using SPI, which supposed to be slower(supports ANSI SQL, checks for permissions etc) and now we will be using CAQL which I hope supposed to light-weight. Haven't found any unit-tests yet, working on adding some.

hornn:
I found some regression tests that check that we get the an external oid out of the range of existing objects, maybe it will be useful, or you can enhance them.  https://github.com/apache/incubator-hawq/blob/master/src/test/regress/output/json_load.source and https://github.com/apache/incubator-hawq/blob/master/src/test/regress/json_utils.c

Didn't know the SPI is slower than CAQL, if so then this change is really good because the code is much cleaner this way.

sansanichfb:
Regarding performance comparison - I tried running HCatalog queries subsequently in new sessions - CAQL query with ORDER BY and WHERE oid > maxOid faster than SPI query ~ 12 times.

src/backend/access/transam/varsup.c
```diff
@@ -408,6 +409,9 @@ GetNewExternalObjectId(void)
 	/*
 	 * must perform check on External Oid range on
 	 * initial access of NextExternalOid
+	 *
+	 * It's needed for upgrade scenario from old version
+	 * of HAWQ which doesn't support dedicated oid pool for HCatalog objects
 	 */
 	if (!IsExternalOidInitialized)
 	{
```
shivzone: Can refactor this 
if(master_highest_used_oid() < FirstExternalObjectId)
{
    ResetExternalObjectId;
}
else
{
    ereport(....);
}

src/backend/access/transam/varsup.c
```diff
@@ -474,74 +483,128 @@ ResetExternalObjectId(void)
...
 
-	for (int i = 0; i < rows; i++)
-	{
-		HeapTuple tuple = tuptable->vals[i];
-		tableNames[i] = SPI_getvalue(tuple, tupdesc, 1);
-	}
+			while (HeapTupleIsValid(innerTuple))
+			{
+				currentOid = HeapTupleGetOid(innerTuple);
+				innerTuple = caql_getnext(pcqInnerCtx);
```
kavinderd: If the caql query orderd by oid value, why do you need to call caql_getnext repeatedly?

sansanichfb: CAQL only supports ordering in ascending order, so this logic walks over result set up to last row which supposed to be max oid for current table.

kavinderd: I see, that sucks

PR #1007, opened 2016-11-10
HAWQ-1148. Update gtest-parallel to make sure test case can run in bo…

amyrazz44:
…th parallel way and serial way

Signed-off-by: wengyanqing <wengyanqing@gmail.com>

### DIFF
--- a/src/test/feature/README.md
+++ b/src/test/feature/README.md
@@ -15,8 +15,8 @@ Before building the code of feature tests part, just make sure your compiler sup
 1. Make sure HAWQ is running correctly. If not, `init` or `start` HAWQ at first. Note please don't set locale related arguments for hawq init.
 2. Load environment configuration by running `source $INSTALL_PREFIX/greenplum_path.sh`.
 3. Load hdfs configuration. For example, `export HADOOP_HOME=/Users/wuhong/hadoop-2.7.2 && export PATH=${PATH}:${HADOOP_HOME}/bin`. Since some test cases need `hdfs` and `hadoop` command, just ensure these commands work before running. Otherwise you will get failure.
-4. Run the cases with`./parallel-run-feature-test.sh 8 ./feature-test`(in this case 8 threads in parallel), you could use `--gtest_filter` option to filter test cases(both positive and negative patterns are supported). Please see more options by running `./feature-test --help`.
-5. As for now, there are several cases which in the sequence.txt file can not parallel run. Run the cases with `./parallel-run-feature-test.sh 8 ./feature-test --gtest_filter=-`cat ./sequence.txt``.
+4. Run the cases with`./parallel-run-feature-test.sh 8 ./feature-test`(in this case 8 threads in parallel), you could use `--gtest_filter` option to filter test cases(both positive and negative patterns are supported). Please see more options by running `./feature-test --help`. 
+5.You can also run cases with `./parallel-run-feature-test.sh 8 ./feature-test --gtest_schedule` (eg. --gtest_schedule=./schedule.txt) if you want to run cases in both parallel way and serial way.The schedule file sample is schedule.txt which stays in the same directory.
 
 # Development
 In contribution to HAWQ, we suggest developers submitting feature tests related to your feature development. In writting a featurte test, you need to write a cpp file inside corresponding folders. There are two recommended way to write this cpp file:
--- a/src/test/feature/gtest-parallel
+++ b/src/test/feature/gtest-parallel
@@ -170,7 +170,7 @@ class FilterFormat:
       self.out.permanent_line("FAILED TESTS (%d/%d):"
                               % (len(self.failures), self.total_tests))
       for (binary, test) in self.failures:
-        self.out.permanent_line(" " + binary + ": " + test)
+        self.out.permanent_line(" " + binary + ": " + test + " [" + test_map[test] + "]")
     self.out.flush_transient_output()
 
 class RawFormat:
@@ -234,6 +234,69 @@ class TestTimes(object):
     except IOError:
       pass  # ignore errors---saving the times isn't that important
 
+
+def get_schedule(schedule_file):
+  "read schedule file from --gtest_schedule option"
+  try:
+    sfile = open(schedule_file, 'r')
+  except (EOFError, IOError):
+    sys.exit("Read file error")
+  
+  filter_test,  ptest, stest = [], [], []
+
+  for line in sfile.readlines():
+    if not line.strip():
+      continue
+    if line[0] == '#':
+      continue
+
+    testline = line.split('=')
+    if len(testline) != 2 or (not testline[1].strip()):
+      sys.exit("format error in schedule file")
+    
+    (key, value) = testline
+    if key == "PARALLEL":
+      ptest.append(value) 
+    elif key == "SERIAL":
+      stest.append(value)
+    else:
+      sys.exit("format error in schedule file")
+
+  filter_test = ptest + stest	
+  sfile.close()
+  return len(ptest), filter_test
+
+
+def do_gtest_filter(list_command, command, op_filter):
+    "get tests by --gtest_filter and --gtest_list_tests"
+    list_command += ['--gtest_filter=' + op_filter]
+    try:
+      test_list = subprocess.Popen(list_command + ['--gtest_list_tests'],
+                                 stdout=subprocess.PIPE).communicate()[0]
+    except OSError as e:
+      sys.exit("%s: %s" % (test_binary, str(e)))
+
+    command += additional_args
+    tests = []
+    test_group = ''
+    for line in test_list.split('\n'):
+      if not line.strip():
+        continue
+      if line[0] != " ":
+        # Remove comments for typed tests and strip whitespace.
+        test_group = line.split('#')[0].strip()
+        continue
+      # Remove comments for parameterized tests and strip whitespace.
+      line = line.split('#')[0].strip()
+      if not line:
+        continue
+      test = test_group + line
+      if not options.gtest_also_run_disabled_tests and 'DISABLED_' in test:
+        continue
+      tests.append((times.get_test_time(test_binary, test),
+                    test_binary, test, command))
+    return tests
+
 # Remove additional arguments (anything after --).
 additional_args = []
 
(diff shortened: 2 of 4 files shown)


amyrazz44:
@paul-guo- @xunzhang  pls review this pr , thank you.

src/test/feature/gtest-parallel
```diff
@@ -170,7 +170,7 @@ class FilterFormat:
       self.out.permanent_line("FAILED TESTS (%d/%d):"
                               % (len(self.failures), self.total_tests))
       for (binary, test) in self.failures:
-        self.out.permanent_line(" " + binary + ": " + test)
+        self.out.permanent_line(" " + binary + ": " + test + " [" + test_map[test] + "]")
```
xunzhang: I suggest you to use format string for readability and correctness. For example, use `" %s: %s [ %s ]" % (binary, test, test_map[test])` here.

src/test/feature/gtest-parallel
```diff
@@ -234,6 +234,71 @@ class TestTimes(object):
...
+def get_schedule(schedule_file):
+  "read schedule file from --gtest_schedule option"
+  try:
+    file = open(schedule_file, 'r')
+  except (EOFError, IOError):
+    sys.exit("Read file error")
+  
+  filter_test = []
+  ptest = []
+  stest = []
```
xunzhang: `filter_test, ptest, stest = [], [], []` is better in Python.

src/test/feature/gtest-parallel
```diff
@@ -234,6 +234,71 @@ class TestTimes(object):
     except IOError:
       pass  # ignore errors---saving the times isn't that important
 
+
+def get_schedule(schedule_file):
+  "read schedule file from --gtest_schedule option"
+  try:
+    file = open(schedule_file, 'r')
```
xunzhang: Do not use default keyword `file`, it is risky.

src/test/feature/schedule.txt
```diff
@@ -0,0 +1,6 @@
+#PARALLEL=* are the parallel tests to run, optional but should not be empty
+#SERIAL=* are the serial tests to run, optional but should not be empty
+#you can have several PARALLEL or SRRIAL
+
+PARALLEL=TestErrorTable.*:TestExternalTable.*:TestPreparedStatement.*:TestUDF.*:TestAOSnappy.*:TestAlterOwner.*:TestAlterTable.*:TestCreateTable.*:TestGuc.*:TestType.*:TestDatabase.*:TestParquet.*:TestPartition.*:TestSubplan.*:TestAggregate.*:TestCreateTypeComposite.*:TestGpDistRandom.*:TestInformationSchema.*:TestQueryInsert.*:TestQueryNestedCaseNull.*:TestQueryPolymorphism.*:TestQueryPortal.*:TestQueryPrepare.*:TestRowTypes.*:TestQuerySequence.*:TestCommonLib.*:TestToast.*:TestTransaction.*:TestCommand.*:TestCopy.*:TestExternalTable.TestExternalTableAll
```
xunzhang: Should we add some checks in the python script above to ensure the valid input schedule?

src/test/feature/gtest-parallel
```diff
@@ -234,6 +234,71 @@ class TestTimes(object):
...
+  except (EOFError, IOError):
+    sys.exit("Read file error")
+  
+  filter_test = []
+  ptest = []
+  stest = []
+
+  for line in file.readlines():
```
xunzhang: Actually I do not like the for-loop her because I think it is a little tricky. If you need a key-value dict information, could we change the file with json format?

xunzhang: I do not understand, is it just one line of total file content here?

src/test/feature/gtest-parallel
```diff
@@ -234,6 +234,71 @@ class TestTimes(object):
...
+      test_list = subprocess.Popen(list_command + ['--gtest_list_tests'],
+                                 stdout=subprocess.PIPE).communicate()[0]
+    except OSError as e:
+      sys.exit("%s: %s" % (test_binary, str(e)))
+
+    command += additional_args
+    tests = []
+    test_group = ''
+    for line in test_list.split('\n'):
+      if not line.strip():
```
xunzhang: What is the meaning of this line?

src/test/feature/gtest-parallel
```diff
@@ -285,41 +354,42 @@ else:
...
+parallel_id = 0
+
 for test_binary in binaries:
   command = [test_binary]
   if options.gtest_also_run_disabled_tests:
     command += ['--gtest_also_run_disabled_tests']
-
   list_command = list(command)
+  
+  if options.gtest_schedule != '' and options.gtest_filter != '':
```
xunzhang: `if options.gtest_schedule  and options.gtest_filter:`

src/test/feature/gtest-parallel
```diff
@@ -285,41 +354,42 @@ else:
...
+    sys.exit("Option input failure : gtest_schedule and gtest_filter can not use in the same time: \n")
+    
+  if options.gtest_schedule != '':
```
xunzhang: `if options.gtest_schedule `

src/test/feature/gtest-parallel
```diff
@@ -285,41 +354,42 @@ else:
...
+    (pcount, filter_test) = get_schedule(options.gtest_schedule)
+    for i in range(0, len(filter_test)):
```
xunzhang: It is like C-style, not pythonic. Use enumerate instead to make the code simple. You can implement this logic using less than 10 lines of code.

src/test/feature/README.md
```diff
@@ -15,8 +15,7 @@ Before building the code of feature tests part, just make sure your compiler sup
...
 3. Load hdfs configuration. For example, `export HADOOP_HOME=/Users/wuhong/hadoop-2.7.2 && export PATH=${PATH}:${HADOOP_HOME}/bin`. Since some test cases need `hdfs` and `hadoop` command, just ensure these commands work before running. Otherwise you will get failure.
-4. Run the cases with`./parallel-run-feature-test.sh 8 ./feature-test`(in this case 8 threads in parallel), you could use `--gtest_filter` option to filter test cases(both positive and negative patterns are supported). Please see more options by running `./feature-test --help`.
-5. As for now, there are several cases which in the sequence.txt file can not parallel run. Run the cases with `./parallel-run-feature-test.sh 8 ./feature-test --gtest_filter=-`cat ./sequence.txt``.
+4. Run the cases with`./parallel-run-feature-test.sh 8 ./feature-test`(in this case 8 threads in parallel), you could use `--gtest_filter` option to filter test cases(both positive and negative patterns are supported). You can also run cases with `--gtest_schedule` (eg. --gtest_schedule=./schedule.txt) if you want to run cases in both parallel way and serial way. Please see more options by running `./feature-test --help`.
```
paul-guo-: 1) use -> add the?
2) Maybe add a simple example for --gtest_filter.

src/test/feature/schedule.txt
paul-guo-: Just curious. Does that means if dev adds a test class, they must modify this file also. Maybe we should try to make the default cases run in parallel by default.

By the way, all of these in one line is really not human readable and is fragile. I'd suggest list them one line by line. You could use a custom format or yaml whatever format?

src/test/feature/gtest-parallel
```diff
@@ -260,6 +325,8 @@ parser.add_option('--gtest_color', type='string', default='yes',
                   help='color output')
 parser.add_option('--gtest_filter', type='string', default='',
                   help='test filter')
+parser.add_option('--gtest_schedule', type='string', default='',
+                  help='test schedule')
 parser.add_option('--gtest_also_run_disabled_tests', action='store_true',
                   default=False, help='run disabled tests too')
 parser.add_option('--format', type='string', default='filter',
```
paul-guo-: Well. The help info is just a copy of option string. That is not good I think. We should make users quick and easy to understand just from the help info.

src/test/feature/gtest-parallel
amyrazz44: Will consider this, thank you.

src/test/feature/gtest-parallel
amyrazz44: Will refine this, thank you.

src/test/feature/README.md
amyrazz44: Already added an example in the README.md , so can not get your points. Please be more specific , thank you

src/test/feature/gtest-parallel
amyrazz44: To deal with the blank line.

src/test/feature/gtest-parallel
amyrazz44: Will specify the help info, thank you.

src/test/feature/gtest-parallel
xunzhang: If you use `test_list.split()`, it will handle this case. For example, `'a\n\nc\n'.split()` will return `['a', 'c']`

src/test/feature/gtest-parallel
amyrazz44: Will consider your suggestion in the future, while in this script this way will be more suitable due to the deal with test_map.

src/test/feature/schedule.txt
xunzhang: Yes @paul-guo- , I agree with you. I think `json` or `yaml` format will defnitely shorten the implementation of getting these parameters. To give powerful express ability, we can let users use `*` or other regular expression. But `:` may be tricky.

amyrazz44: Will refine this, thank you

amyrazz44: Considering only two keys in the schedule.txt file and the values of test cases are too long , it will be too ugly to see. So, no need to use the json file.

amyrazz44: @paul-guo- If users want to add new cases temporary , schedule file contains the test cases in hawq-ci will be easily changed to run test case.

huor: It would be good to use yaml for the schedule which is more human readable.

huor: The name of the schedule can be more specific, i.e, scheduke_sanity, schedule_full_regression, etc.

xunzhang:
LGTM

src/test/feature/schedule.txt
amyrazz44: Thanks for all your suggestions, I have created a story to do the further enhancement.

paul-guo-: Thanks. Another JIRA would also be ok.

paul-guo-:
+1

xunzhang:
Please close this pull request after merge.

PR #1008, opened 2016-11-11
HAWQ-1152. PXF bridge endIteration function

shivzone:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/BridgeResource.java
```diff
@@ -158,6 +158,11 @@ public void write(final OutputStream out) throws IOException,
                     if (!threadSafe) {
                         unlock(dataDir);
                     }
+                    try {
+                        bridge.endIteration();
```
denalex: should we unlock after closing resources, at the very end ? lock --> start --> end --> unlock ?

shivzone: I'm about to create another PR, as this one seems to have pulled an extra commit from the hawq main git which doesn't show up on github mirror. Please wait to to the CR on another PR i'll send shortly

PR #1012, opened 2016-11-14
Revert "HAWQ-1141. Correct default values."

edespino:
This reverts commit 6f8722a6b869167494a1306e9db0ac404ddfa79a.

We need to revert the change to contrib/hawq-ambari-plugin/build.properties which incorrectly updated the hawq.repo.prefix and hawq.addons.repo.prefix properties with "hdb" values. The OSS files should reference the Apache HAWQ product name (hawq).


mithmatt:
+1

ljainpivotalio:
+1

PR #1019, opened 2016-11-22
HAWQ-870. Allocate target's tuple table slot in PortalHeapMemory during split partition

interma:
Port GPDB split-partiton optimization.

Already test it:
1, run sql manually: first do partition sql, and then select result to check.
2, run unit test.
3, run feature test: although have some failed tests, but same result as the code before my modification. 
4, run regress test: only run partition test.

All passed.


interma:
just review and checkin my commit **HAWQ-870 d83b112.**
@paul-guo- @xunzhang tks!

wengyanqing:
LGTM

interma:
Sorry for too many confusing commits here.
I will close this PR, and open a new soon.

PR #1027, opened 2016-11-25
HAWQ-1149. Fixed relcache reference leak when pg_class tuple is invalid

liming01:
Fixed below warning:
WARNING:  relcache reference leak: relation "pg_class" not closed

The test case exists when the meta tuple in pg_class in invalid.

This bug also exists in postgreSQL, although it doesn't called for some reason.


wengyanqing:
Great, +1

PR #1029, opened 2016-11-29
HAWQ-1174. double type core counter of container set has precision issue

jiny2:
This fix is to avoid double precision problem when counting available core resource for a container set.

### DIFF
--- a/src/backend/resourcemanager/include/resourcepool.h
+++ b/src/backend/resourcemanager/include/resourcepool.h
@@ -715,6 +715,7 @@ SimpStringPtr build_segment_status_description(SegStat segstat);
 #define EPSILON 1e-7
 #define IS_DOUBLE_ZERO(d)       (fabs(d) < EPSILON)
 #define IS_DOUBLE_EQ(x, y)      ((fabs((x) - (y))) <= (EPSILON))
+#define IS_DOUBLE_EQGREATER(x, y)  (((x) + EPSILON) >= (y))
 
 void validateResourcePoolStatus(bool refquemgr);
 
--- a/src/backend/resourcemanager/include/resourcepool.h
+++ b/src/backend/resourcemanager/include/resourcepool.h
@@ -715,7 +715,7 @@ SimpStringPtr build_segment_status_description(SegStat segstat);
 #define EPSILON 1e-7
 #define IS_DOUBLE_ZERO(d)       (fabs(d) < EPSILON)
 #define IS_DOUBLE_EQ(x, y)      ((fabs((x) - (y))) <= (EPSILON))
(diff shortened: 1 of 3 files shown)


jiny2:
@linwen @ictmalili please help to review and comment, thanks.

src/backend/resourcemanager/include/resourcepool.h
```diff
@@ -715,6 +715,7 @@ SimpStringPtr build_segment_status_description(SegStat segstat);
 #define EPSILON 1e-7
 #define IS_DOUBLE_ZERO(d)       (fabs(d) < EPSILON)
 #define IS_DOUBLE_EQ(x, y)      ((fabs((x) - (y))) <= (EPSILON))
+#define IS_DOUBLE_EQGREATER(x, y)  (((x) + EPSILON) >= (y))
```
linwen: I prefer to using IS_DOUBLE_EGT(equal greater than) instead of IS_DOUBLE_EQGREATER,
a little suggestion.

jiny2: How about EQGRT?

linwen: or IS_DOUBLE_GE (greater than or equal to)

some examples of abbreviation for comparison operator:
http://folk.uio.no/hpl/scripting/doc/f77/tutorial/logical.html
https://cn.perlmaven.com/comparing-scalars-in-perl
http://www.computerperformance.co.uk/powershell/powershell_comparison_operators.htm

jiny2: ok, let me follow this GE

jiny2: Fixed

jiny2:
@huor please review

zhangh43:
+1, Yi could you please tell us the failure case before? I think the failure case can help us to determine the EPSILON

jiny2:
@zhangh43  YARN mode, 2 segments, each segment has 75gb mem, 15core resource, and the default queue has 25% capacity, 80% maximum capacity. Then allocated and return 12 x 256mb vsegs, then wait more than 5 min. RM will try to return yarn containers but due to precision problem, there are always some yarn containers cannot be returned. because 0.9999999999... is less than 1. As to EPSILON, I adopted previous definition introduce by Wen. I think it should work appropriately.

(gdb) p *containerset
$6 = {Allocated = {MemoryMB = 5120, Core = 1, Ratio = 0}, Available = {MemoryMB = 5120,
Core = 0.99999999999999956, Ratio = 0}, Containers = 0x7ff07a93c870}
(gdb) quit
A debugging session is active.

zhangh43:
got it, +1

huor:
+1

PR #1031, opened 2016-12-01
HAWQ-1182. Add Macro for unused argument and variable.

paul-guo-:


src/include/postgres.h
```diff
@@ -513,6 +513,18 @@ extern void gp_set_thread_sigmasks(void);
...
+#ifndef POSSIBLE_UNUSED_VAR
+#define POSSIBLE_UNUSED_VAR(x)	((void)x)
+#endif
+
+#ifndef POSSIBLE_UNUSED_ARG
+#define POSSIBLE_UNUSED_ARG(x)	((void)x)
+#endif
+
+#ifndef UNUSED_ARG
+#define UNUSED_ARG(x)			((void)x)
```
xunzhang: ......indent

xunzhang: +1 except for the indent above

wengyanqing:
LGTM

src/include/postgres.h
paul-guo-: It is because the tab stop shown on this page is 8, while that in hawq 4.

PR #1033, opened 2016-12-02
HAWQ-1183. Writable external table with Hash distribution shows slow …

paul-guo-:
…performance

This fixes some warnings also in the affected source file.


paul-guo-:
I thought adding a case, but it seems that it is hard to add one in the current framework.

xunzhang:
ok, maybe we could test it manually for this.

paul-guo-:
@zhangh43

paul-guo-:
@huor

zhangh43:
+1 for external table bucket number change.

huor:
+1

paul-guo-:
Merged.

PR #1039, opened 2016-12-05
HAWQ-1191. Decomissioned DELIMITER from HiveOrc profile.

sansanichfb:


shivzone:
I'm a bit concerned about the impact of removing this function parseDelimiterChar completely without completely understand the impact of it w.r.t Partition filtering. Lets merge this change once we verify that Partition based filtering works fine with all the Hive related profiles

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveResolver.java
```diff
@@ -79,7 +79,7 @@
     private String serdeName;
     private String propsString;
     String partitionKeys;
-    char delimiter;
+    char delimiter = 1;
```
kavinderd: Why do we even need this value anymore?

kavinderd:
So the delimiter we were previously explicitly asking for in the DDL will just be retrieved internally?

sansanichfb:
@shivzone, @kavinderd I double checked for HiveRc and HiveText - they still need delimiter, which should be exposed to users, so in this PR I just removed DELIMITER for HiveOrc profile.

PR #1040, opened 2016-12-06
HAWQ-1195. Fixed error "Two or more external tables use the same erro…

liming01:
…r table "

The error table should be works same as normal user table on hdfs, which support multiple INSERTs.

Hi @wangzw, could you help me to review it? Thanks.


wengyanqing:
LGTM

huor:
+1

liming01:
Revert this commit because running installcheck-good failed.
It is my fault, I had thought that we have move all testcases in installcheck-good to featuretest, so I only ran feature test.

Need more time to investigate why it conflict at the same relation file number.

liming01:
When we found the same error table used for the same query, we should assign a new sets of seg files for INSERT INTO ERROR TABLE, which means different relation will insert into the different sets of segfiles.

wangzw:
Hi guys, please see the comments in the JIRA.

PR #1041, opened 2016-12-07
HAWQ-1196. Added partitons support to HiveOrc profile.

sansanichfb:


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveResolver.java
```diff
@@ -254,16 +263,17 @@ void initPartitionFields() {
                             "Unsupported partition type: " + type);
             }
             addOneFieldToRecord(partitionFields, convertedType, convertedValue);
+            numberOfPartitions = partitionFields.size();
```
shivzone: You can instead just set numberOfPartitions after the for loop is complete

PR #1043, opened 2016-12-08
HAWQ-1179. Call Bridge api with profile value read from Fragmenter call.

sansanichfb:


shivzone:
@sansanichfb please provide some context in the Jira even though its a subtask

shivzone:
Not every external table query requires Profile to be configured. You could also have queries using only fragmenter, accessor and resolver classes.

PR #1044, opened 2016-12-09
HAWQ-1178. Enhance Fragmenter api to return profile name.

sansanichfb:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ProfileFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class ProfileFactory {
+
+    private static final String HIVE_TEXT_PROFILE = "HiveText";
+    private static final String HIVE_RC_PROFILE = "HiveRC";
+    private static final String HIVE_ORC_PROFILE = "HiveORC";
+
+    public static String get(InputFormat inputFormat) throws Exception {
+        String profileName = null;
+        //if (inputFormat instanceof TextInputFormat) {
```
lescha: commented out?

sansanichfb: Yep, this PR introduces optimization just for HiveORC, and I think we will incrementally add support for HiveRC, HiveText later.

shivzone: What's preventing us from supporting profile discovery for other ones too, since you have refactored the code anyway. HAWQ-1177 and HAWQ-1178 don't specifically call out profile discovery only for ORC

shivzone:
Good job with the code refactoring/encapsulation

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/ProfileFactory.java
sansanichfb: We need to carry more information to support other profiles, i.e. delimiter needed for HiveText profile, and in case of HCatalog we don't have this information yet.

shivzone: If we do plan to rollout the hcatalog based lookup feature, we would need to deal with that info as well. We can myabe create a separate PR/Jira for that.

sansanichfb: Logged JIRA to incrementally add functionality for other profiles - https://issues.apache.org/jira/browse/HAWQ-1228

PR #1050, opened 2016-12-14
HAWQ-1221. hawq register should error out when register a yml file do…

stanlyxiang:
…esn't exist

### DIFF
--- a/tools/bin/hawqregister
+++ b/tools/bin/hawqregister
@@ -148,6 +148,14 @@ def tablename_handler(tablename):
         return tmp_lst[-2], tmp_lst[-1]
     return 'public', tablename
 
+def check_file_exist(yml_file):
+    if yml_file:
+        if not os.path.exists(yml_file):
+            logger.error('Cannot find yaml file : %s' % yml_file)
+            sys.exit(1)
+        return True
+    return False
+
 class FailureHandler(object):
     def __init__(self, conn):
         self.operations = []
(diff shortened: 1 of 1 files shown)


stanlyxiang:
cc @xunzhang @wcl14

tools/bin/hawqregister
```diff
@@ -1145,7 +1153,7 @@ def main(options, args):
 
     failure_handler = FailureHandler(conn)
     # register
-    if options.yml_config and ispartition(options.yml_config):
+    if check_file_exist(options.yml_config) and ispartition(options.yml_config):
```
wcl14: Should function check_file_exist(options.yml_config) be called in both partition table and non-partition table?

xunzhang: I think it is necessary

stanlyxiang: It will always check yaml file exist or not first.  Then check whether is partition. if partition do HawqRegisterPartition, otherwise HawqRegister. So the check works  before doing register partition table or non-partition table. So there is no need to call the function 2 times.

stanlyxiang:
also cc @linwen @interma

linwen:
+1

tools/bin/hawqregister
wcl14: I just think check_file_exist() should not be called in the if statement. In this statement, it may be confused the condition for whether is partition or not.

stanlyxiang: Yeah, I have to say that the condition might be confused in the if statement. However, this small confusion inherits from the original logic. The if statement says :  if there is yaml file and register partition table, do HawqRegisterPartition. Otherwise do HawqRegister. The else part contains 2 situations. 1) no yaml file 2) have yaml file and register non-partition table.  

Add the function call in the if statement doing nothing except a check. We didn't notice the might confused logic before because we all thought the yaml file exist by default. We could make it more clearly with more code. But I don't think it worthes the effort to change code if we take 1 min to understand the logic.

interma:
+1

PR #1051, opened 2016-12-15
HAWQ-1214. Remove resource_parameters

paul-guo-:


paul-guo-:
@zhangh43 @jiny2 @huor

jinyismilodon:
+1 

One trivial comment about output int64 or unsigned int64 data, I think it is better to use INT64_FORMAT and UINT64_FORMAT instead of %lld. Though this is not actually involved in your fix.

paul-guo-:
I agree with you. @jinyismilodon 
I've fix the code accordingly.

zhangh43:
+1

PR #1053, opened 2016-12-16
HAWQ-1002. Implement configuration GUC related to Ranger.

zhangh43:
This pr includes :
HAWQ-1001. Implement HAWQ basic user ACL check through Ranger.
HAWQ-1002. Implement configuration GUC related to Ranger.
HAWQ-1003. Implement bached ACL check through Ranger.

when GUC enable_ranger is ON, we will use ranger_acl_check to determine where user has the permission to query the data.


huor:
+1

PR #1055, opened 2016-12-19
HAWQ-1001. Bugfix and refactor ranger logic.

xunzhang:


xunzhang:
cc @linwen @ictmalili  @zhangh43

src/backend/libpq/rangerrest.c
```diff
@@ -273,10 +245,10 @@ json_object *create_ranger_request_json_batch(List *args)
  *   }
  */
 json_object* create_ranger_request_json(char* user, AclObjectKind kind, char* object,
-        List* actions, char* how)
+        List* actions, bool how)
```
ictmalili: What do we pass for "how", why change it from char* to bool?

xunzhang: https://github.com/xunzhang/incubator-hawq/blob/3568cbfb790a196e555ed067d13a7483ee872f34/src/backend/catalog/aclchk.c#L2776

paul-guo-: Even change to bool, isn't another name better like is_*?

src/backend/libpq/rangerrest.c
```diff
@@ -391,7 +363,7 @@ static size_t write_callback(char *contents, size_t size, size_t nitems,
         void *userp)
 {
     size_t realsize = size * nitems;
-    CURL_HANDLE curl = (struct curl_context *) userp;
+    CURL_HANDLE curl = (struct curl_context_t *) userp;
```
paul-guo-: Usually if a type have a _t in end, that means no struct is needed.

Either "struct curl_context" or curl_context_t.

src/backend/libpq/rangerrest.c
```diff
@@ -468,7 +440,7 @@ void call_ranger_rest(CURL_HANDLE curl_handle, char* request)
     }
     else
     {
-        elog(LOG, "%lu bytes retrieved from Ranger Restful API.",
+        elog(LOG, "%d bytes retrieved from Ranger Restful API.",
```
paul-guo-: Why not use  size_t for size in struct?

src/backend/libpq/rangerrest.c
xunzhang: Good advice!

src/backend/libpq/rangerrest.c
xunzhang: ok, thanks.

src/backend/libpq/rangerrest.c
xunzhang: It should be, but this pull request is just a hotfix and small refactor. I think ranger codes should be largely refactored later on which is not relevant to this pull request. Thanks.

radarwave:
+1

xunzhang:
Merged into master.

PR #1059, opened 2016-12-21
HAWQ-1229. Remove verbose option in hawq config

radarwave:


paul-guo-:
'-v' seem to have another usage in "hawq config"?

[-v hawq_property_value | --value=hawq_property_value

radarwave:
Yeah, '-v' is used to set the GUC value. '--verbose' is a legacy option, now it's already been removed. This time only remove the wrong help message.

paul-guo-:
I mean why not replace this time.

radarwave:
Good point, I will update it.

radarwave:
@paul-guo- Please review. Thanks.

tools/bin/hawqpylib/HAWQ_HELP.py
```diff
@@ -159,10 +159,10 @@
 
 The "options" are:
    -c --change         Changes a configuration parameter setting.
```
paul-guo-: Actually I think "--change" could better be "--parameter"? ". How do you think?

radarwave: We'd better not change it, HAWQ have many client option names from GPDB, sometimes it's good to have similar options for the users who use both. And if we change any options, we have to change it together with Ambari.

paul-guo-: How. Then +1.

PR #1061, opened 2016-12-21
HAWQ-1226. fix compile error under gcc 4.8.2

interma:
When use gcc version **4.8.5** 20150623 (Red Hat 4.8.5-4) (GCC), there is a warning:
```
be-secure.c:323:1: warning: 'report_commerror' defined but not used [-Wunused-function]
 report_commerror(const char *err_msg)
 ^
```

But it will cause gcc 4.8.2 to make error if we remove this function. So let's keep it as before.

### DIFF
--- a/src/backend/libpq/be-secure.c
+++ b/src/backend/libpq/be-secure.c
@@ -318,8 +318,6 @@ rloop:
  * This function holds an interrupt before reporting this error to avoid
  * a self deadlock situation, see MPP-13718 for more info.
  */
-
-/* no used now, avoid compile warnning
 static void
 report_commerror(const char *err_msg)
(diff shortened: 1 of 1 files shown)


linwen:
+1

interma:
@linwen @huor thanks for review.

liming01:
+1

interma:
Finally found the root cause is not compiler version, but is configure option **--with-openssl**.
So remove report_commerror() while enable this option at the same time will cause make error.

Fix the similiar problem pls see: [#1062](https://github.com/apache/incubator-hawq/pull/1062)
Thanks, @paul-guo-

PR #1063, opened 2016-12-22
[HAWQ-1232] Update versions strings for 2.1.0.0-incubating release.

edespino:
Update HAWQ and HAWQ Ambari Plugin version strings for 2.1.0.0-incubating release.

Once the release branch is created, I will be updating the "GP_VERSION" value in "getversion" to "2.1.0.0-incubating".


paul-guo-:
Thanks, ed. +1. Really easy to omit if I work on this :-)
Do we have any document so that anyone else who work on releasing in the future could easily kick off?

edespino:
FYI: I have performed a full build and all components are building properly.

PR #1064, opened 2016-12-24
HAWQ-1215. Support Complextypes with HiveORC

kavinderd:
Preliminary PR for complex type support in Hive

### DIFF
--- a/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCSerdeResolver.java
+++ b/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCSerdeResolver.java
@@ -104,7 +104,10 @@ public class HiveORCSerdeResolver extends HiveResolver {
 
         StringBuilder columnNames = new StringBuilder(numberOfDataColumns * 2); // column + delimiter
         StringBuilder columnTypes = new StringBuilder(numberOfDataColumns * 2); // column + delimiter
-        String[] hiveColTypes = typesString.split(":");
+        String[] tmp = typesString.split(":");
+        String[] hiveColTypes = new String[numberOfDataColumns];
+        parseColTypes(tmp, hiveColTypes);
+
         String delim = ",";
         for (int i = 0; i < numberOfDataColumns; i++) {
             ColumnDescriptor column = input.getColumn(i);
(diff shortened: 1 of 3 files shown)


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCSerdeResolver.java
```diff
@@ -102,11 +104,19 @@ void initSerde(InputData input) throws Exception {
 
         StringBuilder columnNames = new StringBuilder(numberOfDataColumns * 2); // column + delimiter
         StringBuilder columnTypes = new StringBuilder(numberOfDataColumns * 2); // column + delimiter
+        String[] tmp = typesString.split(":");
```
sansanichfb: maybe cols or columns instead of tmp

pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveORCSerdeResolver.java
```diff
@@ -102,11 +104,19 @@ void initSerde(InputData input) throws Exception {
...
+        String[] hiveColTypes = new String[numberOfDataColumns];
+        parseColTypes(tmp, hiveColTypes);
+
         String delim = ",";
         for (int i = 0; i < numberOfDataColumns; i++) {
             ColumnDescriptor column = input.getColumn(i);
             String columnName = column.columnName();
             String columnType = HiveUtilities.toCompatibleHiveType(DataType.get(column.columnTypeCode()), column.columnTypeModifiers());
+            //Complex Types will have a mismatch between Hive and Hawq type
```
sansanichfb: As far as we have mappings between Hive ad Hawq types in EnumHiveToHawqType, I think we can update HiveUtilities.toCompatibleHiveType method to handle complex types properly and avoid the following section.

kavinderd: The only issue I have with that approach is that whether a column is a complex type is table specific so HiveUtilities.toCompatibleHiveType would need to also take the Hive column information as an additional parameter since Hawq type text could either be a Hive string or any Hive complex type.

I think it does make sense to put the below logic into a utility method

PR #1065, opened 2016-12-24
HAWQ-1208. Porting gpdb interconnect fix to hawq.

stanlyxiang:
cc @liming01 @wengyanqing @huor


huor:
Given there are multiple fix in each of commit, it is better to merge Ming's port and your fix that span in several commit into single commit. The commit message need to contain all port and fix as well. Any better option from your mind?

liming01:
It is better to keep them separate commit, so that we can easy tracing the problem. If you think they are too many commit, you can using merge instead of rebase. 

+1.

interma:
+1

stanlyxiang:
Merged to master.

PR #1069, opened 2016-12-28
HAWQ-1241. No need of ext/python in *PATH in file greenplum_path.sh

paul-guo-:
Also remove the Solaris branch code since we do not support Solaris.


radarwave:
LGTM +1

paul-guo-:
Small issue. Merged.

PR #1070, opened 2016-12-28
HAWQ-1243. Add suffix name for ranger restful service.

zhangh43:
Except rps_addr_host and rps_addr_port, we also need rps_addr_suffix to
We will use this GUC to construct rest service address:
http://rps_addr_host:rps_addr_port/rps_addr_suffix


linwen:
+1

PR #1071, opened 2016-12-29
HAWQ-1242. hawq-site.xml default content has wrong guc variable names

jiny2:
Fixed the wrong names in hawq-site.xml and its template file.

### DIFF
--- a/src/backend/utils/misc/etc/hawq-site.xml
+++ b/src/backend/utils/misc/etc/hawq-site.xml
@@ -33,7 +33,7 @@ under the License.
 	</property>
 
 	<property>
-		<name>hawq_standby_address_host</name>
+		<name>standby_address_host</name>
 		<value>none</value>
 		<description>The host name of hawq standby master.</description>
 	</property>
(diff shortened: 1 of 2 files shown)


paul-guo-:
+1

jiny2:
@huor @radarwave Please have a review. Thanks

huor:
+1 for the fix

stanlyxiang:
Some of the concourse feature tests failed with this commit. There are some other place use these guc name with the prefix 'hawq_'. So what about keep the prefix in the hawq-site.xml and add the prefix to the guc in guc.c ? This can reduce the work to find all the places use the several guc you changed and keep the name consistently with other gucs name.

jiny2:
I find there is some python script logic using the hawq_* guc names, I will keep those hawq_* guc names to avoid changing python scripts.

jiny2:
@huor @stanlyxiang @paul-guo- 

@stanlyxiang  is right, the correct approach is to fix guc.c as all external scripts and documentation are using hawq_* names.

paul-guo-:
1) Please make sure the doc has corresponding change if needed (notify them if needed). If I remember correctly, dfs_url is still in that doc.
2) Please make the commit message begin with HAWQ-1242 since we have a script, which needs this pattern, to detect the Apache JIRA status.

paul-guo-:
Otherwise +1. Thanks.

radarwave:
LGTM.

Just wondering how hawq read these mis-match GUCs before you change them? Seems they still can get the right values.

stanlyxiang:
@radarwave  In my investigate, I found that hawq read these GUCs from a config file which set these values when init, since there is no database to store them before init.  Maybe no one config / show them using hawq config or they use option "--skipValidation". No one change and no one hit this problem, only some warnings exist in logs.  No function read "dfs_url" "master_directory" "seg_directory". 
But RM function sendIMAlive() will use "standby_addr_host" and I haven't found the reason. Maybe @jiny2 knows why.

jiny2:
In fact some gucs are not used by reading from hawq-site.xml, for example, hawq_master_directory. But for standby_addr_host, it is useful for RM as RM switches its heart-beat to standby server when master server is not achievable.

PR #1073, opened 2017-01-03
HAWQ-1248. Merge Dockerfiles for HAWQ Dev into HAWQ code base.

guofengrichard:
This commit is to merge Dockerfiles into HAWQ code base. The Dockerfiles are used to create docker images, which provide an out-of-box way for developers to setup build and test environment for HAWQ.


contrib/hawq-docker/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+* build Apache HAWQ
+```
+cd /data/hawq
+./configure --prefix=/data/hawq-dev
+make
+make install
+```
+(When you are using CentOS 6, run command `scl enable devtoolset-2 bash` before
+configuring hawq and run command `exit` after installing hawq.)
```
paul-guo-: Why "exit" command is needed? exit the container?

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/make all
```
paul-guo-: Please add License header into all files which could add.

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
...
+
+all: 
+	@echo " Usage:"
+	@echo "    To setup a build and test environment:         make run"
+	@echo "    To start all containers:                       make start"
+	@echo "    To stop all containers:                        make stop"
+	@echo "    To remove hdfs containers:                     make clean"
+	@echo "    To remove all containers:                      make distclean"
+	@echo ""
+	@echo "    To build images locally:                       make build"
```
paul-guo-: What is the cleanup command for "make build"?  (make distclean or make clean?)

paul-guo-:
By the way, what is the difference between hawq-dev and hawq-test? (One for dev environment and another includes hdfs and/or hawq? It seems that there is no comment for this), And it seems many configuration files comes from hdfs. Why not use the default configuration files and configure them later in CLI?

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
+#!/usr/bin/make all
+
+THIS_MAKEFILE_PATH := $(abspath $(lastword $(MAKEFILE_LIST)))
+TOP_DIR := $(abspath $(dir ${THIS_MAKEFILE_PATH}))
+NDATANODES := 3
+CUR_DATANODE := 1
```
stanlyxiang: I think NDATANODES is the total number of datanodes. And CUR_DATANODE is a iteration placeholder during the loop for start/stop/run/other operations for the hdfs.  

I think a configurable number of datanodes are more reasonable. And you can set 3 as default value.

contrib/hawq-docker/Makefile
guofengrichard: That makes sense. Will add License header.

contrib/hawq-docker/Makefile
guofengrichard: @stanlyxiang is right about NDATANODES and CUR_DATANODE.

Good suggestion on configuring the number. For now user can define the value via parameters to 'make', for example
 $ make NDATANODES=6

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
...
+OS_VERSION := centos7
+# Do not use underscore "_" in CLUSTER_ID
+CLUSTER_ID := $(OS_VERSION)
+# Monut this local directory to /data in data container and share with other containers
+LOCAL :=
```
guofengrichard: Yes. '/data' is in containers and will be shared among all the containers.
'LOCAL' is supposed to be a local directory on the host. If it is defined, the directory will be mounted to '/data' in all containers.

contrib/hawq-docker/Makefile
guofengrichard: No cleanup command for 'make build'. Removing docker images is not encouraged.

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
...
+	@echo "    To remove hdfs containers:                     make clean"
+	@echo "    To remove all containers:                      make distclean"
+	@echo ""
+	@echo "    To build images locally:                       make build"
+	@echo "    To pull latest images:                         make pull"
+
+build:
+	@make -f $(THIS_MAKEFILE_PATH) build-hawq-dev-$(OS_VERSION)
+	@make -f $(THIS_MAKEFILE_PATH) build-hawq-test-$(OS_VERSION)
+	@echo "Done!"
```
guofengrichard: That's right.

contrib/hawq-docker/README.md
guofengrichard: The 'exit' command is to exit from the bash environment setup by previous 'scl' command.

contrib/hawq-docker/centos6-docker/hawq-test/conf/slaves
```diff
@@ -0,0 +1 @@
+localhost
```
stanlyxiang: This is the hdfs slaves file. If we start a hdfs cluster, this slaves should contains the 3 hostname , right ? Could you point me when do you update this slaves from "localhost" to "datanode1 datanode2 datanode3"  ? Maybe I missed it.  Thanks

contrib/hawq-docker/centos7-docker/hawq-test/conf/capacity-scheduler.xml
```diff
@@ -0,0 +1,134 @@
+<!--
```
stanlyxiang: What's the difference between these configuration xml files under centos7-docker and centos6-docker ?  If they are the same, please use one copy and point out the right path when COPY them to the docker image. DRY. If they have little differences, we could use most of the common files and just store the different copies. Otherwise you could use 2 copies for every conf files.

contrib/hawq-docker/centos6-docker/hawq-test/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+
+if [ ! -f /etc/profile.d/hadoop.sh ]; then
+  echo '#!/bin/bash' | sudo tee /etc/profile.d/hadoop.sh
+  echo "export NAMENODE=${NAMENODE}" | sudo tee -a /etc/profile.d/hadoop.sh
+  sudo chmod a+x /etc/profile.d/hadoop.sh
+fi
+
+sudo start-hdfs.sh
+sudo sysctl -p
+sudo ln -s /usr/lib/libthrift-0.9.1.so /usr/lib64/libthrift-0.9.1.so
```
stanlyxiang: This link can only be done at this time  ? If not , we could move it to the dockerfile after install libthrift. If we move it, this entrypoint shell script is the same with the entrypoint.sh under centos7-docker.

stanlyxiang:
cc @radarwave  . For Radar used to build docker image for concourse pipeline, maybe he has some comments.

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
...
+
+build-hawq-dev-$(OS_VERSION): $(TOP_DIR)/$(OS_VERSION)-docker/hawq-dev/Dockerfile
+	@echo build hawq-dev:$(OS_VERSION) image
+	docker build -t hawq/hawq-dev:$(OS_VERSION) $(TOP_DIR)/$(OS_VERSION)-docker/hawq-dev/
+
+build-hawq-test-$(OS_VERSION): $(TOP_DIR)/$(OS_VERSION)-docker/hawq-test/Dockerfile
+	@echo build hawq-test:$(OS_VERSION) image
+	docker build -t hawq/hawq-test:$(OS_VERSION) $(TOP_DIR)/$(OS_VERSION)-docker/hawq-test/
```
guofengrichard: Correct. For now two repos are needed.

contrib/hawq-docker/Makefile
```diff
@@ -0,0 +1,205 @@
...
+	@if [ -z "`docker network ls 2>/dev/null`" ]; then \
+                make -f $(THIS_MAKEFILE_PATH) NETWORK=default create-data-container && \
+                make -f $(THIS_MAKEFILE_PATH) NETWORK=default run-hdfs; \
+	else \
+	        if [ -z "`docker network ls 2>/dev/null | grep $(NETWORK)`" ]; then \
+	                echo create network $(NETWORK) && \
+	                docker network create --driver bridge $(NETWORK); \
+	        fi && \
+	        make -f $(THIS_MAKEFILE_PATH) create-data-container && \
+                make -f $(THIS_MAKEFILE_PATH) run-hdfs; \
```
guofengrichard: Thanks for reminding.

contrib/hawq-docker/README.md
```diff
@@ -0,0 +1,96 @@
...
+Take CentOS 7 as an example below.
+
+# Install docker
+* following the instructions to install docker.
+https://docs.docker.com/
+
+# Setup build and test environment
+* clone this repository
+```
+git clone https://github.com/guofengrichard/hawq-docker.git .
```
guofengrichard: Right. I forgot to change this link. Will correct it.

contrib/hawq-docker/README.md
paul-guo-: You could export the new gcc path in PATH instead.

contrib/hawq-docker/README.md
```diff
@@ -0,0 +1,96 @@
...
+```
+* Get the docker images
+```
+  cd hawq-docker
+  make pull (recommended)
+OR
+  make build
```
guofengrichard: Agree with Paul.

guofengrichard:
Based on the comments, I have added in license header and also made amendments to Makefile and README. Other improvements will be made in the future.

stanlyxiang:
Richard, other improvements could be made in the future, but at least remove the installation of libhdfs3 in the dockerfile. Thanks.

guofengrichard:
Installation of libhdfs3 has been removed for CentOS 6.

interma:
What's the problem of libhdfs3's license? 
Very curious~

paul-guo-:
Not an issue of License. Hawq code has included libhdfs3 already.

stanlyxiang:
＋1

paul-guo-:
+1. I think we may need to refine this a bit more.

huor:
@guofengrichard, please close this pr as it is merged to master.

guofengrichard:
Removed unnecessary configure files.

guofengrichard:
These commits have been checked into master of 'apache/incubator-hawq'. So close this pull request.

PR #1074, opened 2017-01-05
HAWQ-1249. Don't do ACL checks on segments

wcl14:


src/backend/catalog/aclchk.c
```diff
@@ -224,16 +224,17 @@ restrict_and_check_grant(bool is_grant, AclMode avail_goptions, bool all_privs,
 	 * If we found no grant options, consider whether to issue a hard error.
 	 * Per spec, having any privilege at all on the object will get you by
 	 * here.
+	 * Superusers or on QE bypass all permission checking.
```
paul-guo-: Why do we still need to check for superusers specially in the code below?

src/backend/catalog/aclchk.c
```diff
@@ -224,16 +224,17 @@ restrict_and_check_grant(bool is_grant, AclMode avail_goptions, bool all_privs,
...
 	 */
-	if (avail_goptions == ACL_NO_RIGHTS)
+	if (avail_goptions == ACL_NO_RIGHTS && GP_ROLE_EXECUTE != Gp_role)
```
paul-guo-: I think you  should change to "Gp_role != GP_ROLE_EXECUTE"
This is more human-understandable and follows the code convention.

src/backend/catalog/aclchk.c
```diff
@@ -3006,8 +3007,8 @@ pg_database_aclmask(Oid db_oid, Oid roleid,
 	Oid			ownerId;
 	cqContext  *pcqCtx;
 
-	/* Superusers or on QE bypass all permission checking. */
-	if (GP_ROLE_EXECUTE == Gp_role || superuser_arg(roleid))
```
paul-guo-: By the way, I know many people like to write "GP_ROLE_EXECUTE == Gp_role" similar code to avoid typo like "GP_role = GP_ROLE_EXECUTE" but gcc will warn against this kind of error.

src/backend/executor/execMain.c
```diff
@@ -1912,45 +1912,10 @@ InitPlan(QueryDesc *queryDesc, int eflags)
 	 * rangetable here --- subplan RTEs will be checked during
 	 * ExecInitSubPlan().
 	 */
-	if (operation != CMD_SELECT ||
-			(Gp_role != GP_ROLE_EXECUTE &&
-			 !(shouldDispatch && cdbpathlocus_querysegmentcatalogs)))
```
paul-guo-: Are you sure these parts of code could be removed? Please double-check to see whether those conditions still apply, following the comment.

wcl14: Yes, we think these parts of code should be removed for there is no need to do ACL checks on segments.

src/backend/catalog/aclchk.c
wcl14: For ranger, we should still do checks but for normal ACL checks, we should bypass all permission checking for  superusers. I should change the comment, and thanks.

wcl14:
@zhangh43 @linwen @stanlyxiang Please review, thanks.

paul-guo-:
I mean, this part of code seems to be called on master surely. I'm not sure
they could be called on segments.

2017-01-09 12:53 GMT+08:00 Chunling Wang <notifications@github.com>:

src/backend/catalog/aclchk.c
linwen: agree

src/backend/executor/execMain.c
wcl14: I have debugged these codes, and am sure they could be called on segments.

src/backend/catalog/aclchk.c
wcl14: Ok, I will change these, thanks.

src/backend/executor/execMain.c
paul-guo-: OK, so will your change introduce regression on master?

src/backend/catalog/aclchk.c
```diff
@@ -3935,6 +3972,10 @@ pg_extprotocol_aclcheck(Oid ptcid, Oid roleid, AclMode mode)
 AclResult
 pg_filesystem_aclcheck(Oid fsysid, Oid roleid, AclMode mode)
 {
+	/* Bypass all permission checking on QE. */
+	if (GP_ROLE_EXECUTE == Gp_role)
+		return ACLCHECK_OK;
+
```
linwen: I think the indent should be same with previous, otherwise in same function, there are two different indents.

src/backend/catalog/aclchk.c
```diff
@@ -224,16 +224,17 @@ restrict_and_check_grant(bool is_grant, AclMode avail_goptions, bool all_privs,
...
+	if (avail_goptions == ACL_NO_RIGHTS && GP_ROLE_EXECUTE != Gp_role)
 	{
 	  if (enable_ranger && !fallBackToNativeCheck(objkind, objectId, grantorId)) {
 	    if (pg_rangercheck(objkind, objectId, grantorId,
 	        whole_mask | ACL_GRANT_OPTION_FOR(whole_mask),
 	        ACLMASK_ANY) != ACLCHECK_OK)
 	      aclcheck_error(ACLCHECK_NO_PRIV, objkind, objname);
 	  }
-	  else {
+	  else if (!superuser_arg(grantorId)){
```
linwen: Is it necessary?

src/backend/executor/execMain.c
wcl14: I think it will not introduce regression on master, but I haven't completed all feature tests yet.

src/backend/catalog/aclchk.c
interma: @wcl14 if you use vim, try it:
`set tabstop=2
set softtabstop=2
set expandtab`

zhangh43:
+1

stanlyxiang:
+1

PR #1075, opened 2017-01-06
HAWQ-762. Login to kerberos if credentials are no longer valid

kavinderd:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/SecuredHDFS.java
```diff
@@ -53,6 +53,10 @@
     public static void verifyToken(ProtocolData protData, ServletContext context) {
         try {
             if (UserGroupInformation.isSecurityEnabled()) {
+                UserGroupInformation loginUser = UserGroupInformation.getLoginUser();
+                if (!loginUser.hasKerberosCredentials()) {
```
sansanichfb: Any chance that loginUser might be null?

kavinderd: I don't think so if we pass the above check `UserGroupInformation.isSecurityEnabled()` but I can add a null check to be safe

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/SecuredHDFS.java
```diff
@@ -53,6 +53,10 @@
...
+                    SecureLogin.login();
```
kavinderd: Instead of `SecureLogin` we can just directly use `UserGroupInformation.getLoginUser().reloginFromKeytab()` seems more lightweight since it leverages all the kerberos information that's already in memory

PR #1077, opened 2017-01-10
HAWQ-1264. Add KEYS file for Code Signing purpose.

edespino:
These are the public KEYS used to validate signed artifacts.  They are use are described in the team's release process wiki section:

https://cwiki.apache.org/confluence/display/HAWQ/Release+Process%3A+step+by+step+guide

### DIFF
--- a/KEYS
+++ b/KEYS
@@ -1,0 +1,72 @@
+This file contains the PGP keys of various developers.
+Please don't use them for email unless you have to. Their main
+purpose is code signing.
+
+Users: pgp < KEYS
+       gpg --import KEYS
+Developers:
+        pgp -kxa <your name> and append it to this file.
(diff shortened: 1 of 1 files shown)


paul-guo-:
Ed, could you please elaborate on the usage and purpose? Thanks.

edespino:
@paul-guo- I added a description to the PR which describes the use of this file.

paul-guo-:
Ok, assuming this is for release, I'm wondering whether we should put in in the release branch only since different release developers could have different keys. And if possible we should add some comments in the file (at least includes the link to the guide above).

edespino:
I see your point, I'll close this PR and document the process to have the release manager update the PGP KEYS file in the release subversion repositories for the release.  Thanks for the great feedback.

PR #1078, opened 2017-01-10
HAWQ-1258. segment resource manager does not switch back when it cannot resolve standby host name

jiny2:
If use does not configure standby host, segment will not switch to standby server; If segment cannot resolve target server no matter the master server or the standby server, the target server for sending IMAlive heart-beat is switched as well.

### DIFF
--- a/src/backend/resourcemanager/communication/rmcomm_RMSEG2RM.c
+++ b/src/backend/resourcemanager/communication/rmcomm_RMSEG2RM.c
@@ -107,8 +107,10 @@ int sendIMAlive(int  *errorcode,
 	if ( res != FUNC_RETURN_OK )
 	{
 		rm_pfree(AsyncCommContext, context);
-		elog(WARNING, "Fail to register asynchronous connection for sending "
-					  "IMAlive message. %d", res);
+		elog(LOG, "failed to register asynchronous connection for sending "
+			      "IMAlive message. %d", res);
+		/* Always switch if fail to register connection here. */
+		switchIMAliveSendingTarget();
 		return res;
 	}
 
@@ -140,7 +142,7 @@ void receivedIMAliveResponse(AsyncCommMessageHandlerContext  context,
 		 buffersize != sizeof(RPCResponseIMAliveData) ) {
 		elog(WARNING, "Segment's resource manager received wrong response for "
 					  "heart-beat request.");
-		DRMGlobalInstance->SendToStandby = !DRMGlobalInstance->SendToStandby;
+		switchIMAliveSendingTarget();
 	}
 	else
 	{
@@ -165,7 +167,7 @@ void sentIMAliveError(AsyncCommMessageHandlerContext context)
 	else
 		elog(WARNING, "Segment's resource manager sending IMAlive message "
 					  "switches from master to standby");
-	DRMGlobalInstance->SendToStandby = !DRMGlobalInstance->SendToStandby;
+	switchIMAliveSendingTarget();
 }
 
 void sentIMAliveCleanUp(AsyncCommMessageHandlerContext context)
--- a/src/backend/utils/misc/guc.c
+++ b/src/backend/utils/misc/guc.c
@@ -8200,7 +8200,7 @@ static struct config_string ConfigureNamesString[] =
 			NULL
 		},
 		&standby_addr_host,
-		"localhost", NULL, NULL
+		"none", NULL, NULL
 	},
 
 	{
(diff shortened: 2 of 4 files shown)


jiny2:
@paul-guo- @linwen Please review, thanks!

linwen:
LGTM

src/backend/utils/misc/guc.c
```diff
@@ -8200,7 +8200,7 @@ static struct config_string ConfigureNamesString[] =
 			NULL
 		},
 		&standby_addr_host,
-		"localhost", NULL, NULL
+		"none", NULL, NULL
```
paul-guo-: How if the user really configures the stands host as "none"? Maybe ""?

src/backend/resourcemanager/communication/rmcomm_RMSEG2RM.c
```diff
@@ -107,8 +107,10 @@ int sendIMAlive(int  *errorcode,
 	if ( res != FUNC_RETURN_OK )
 	{
 		rm_pfree(AsyncCommContext, context);
-		elog(WARNING, "Fail to register asynchronous connection for sending "
-					  "IMAlive message. %d", res);
+		elog(LOG, "failed to register asynchronous connection for sending "
+			      "IMAlive message. %d", res);
+		/* Always switch if fail to register connection here. */
+		switchIMAliveSendingTarget();
```
paul-guo-: How often is switchIMAliveSendingTarget()? I mean if it is called often, we should have a parameter or bit mask to save the info that standby exists or not to avoid frequent string comparison which is obviously heavier than simple logical operations.

src/backend/utils/misc/guc.c
jiny2: That needs @radarwave 's comment, I have similar question, but currently from script implementation, "none" means no standby.

src/backend/resourcemanager/communication/rmcomm_RMSEG2RM.c
jiny2: The interval should be 30 second,  a very low frequency.

paul-guo-: OK. That's fine.

stanlyxiang:
+1

src/backend/utils/misc/guc.c
paul-guo-: I do not think the code should be bundled with the python utility code. That will possibly introduce some potential bugs in the future.

jiny2: If we support "" instead of "none", we have segment keep connecting to "none" as a server. I think if necessary, we need another fix for this.

jiny2: This will cause behavior inconsistency, when "none" is set script thinks there is no standby, but rm segment will try to resolve "none", which is not necessary.

jiny2: So that's why we need another fix if necessary. the fix should fix script and code together. @paul-guo-

paul-guo-: I suspect this does not affect your code.

See the code below,

def get_all_values(self):
        if 'hawq_standby_address_host' in self.hawq_dict:
            if self.hawq_dict['hawq_standby_address_host'].lower() in ['none', '', 'localhost']:
                del self.hawq_dict['hawq_standby_address_host']

However hawq_ctl should remove the "none" case, I think.

@radarwave Please confirm the logic.

jiny2: No, segment rm process does not care what are saved in python code. So, it only knows there is a standby host named "none".  So deleting from hawq_dict does not help.

radarwave: Yeah, it's not so good to use 'none' to indicate there is no standby. But I don't think use '' or '  ' or '             '  are good too. Still let user feel confusion. Maybe we can add a new guc to indicate if we have a standby master defined, the value can be yes/no.

Better to start a new Jira to talk about this.

paul-guo-: OK. Agree, we could start another JIRA to track the hostname "none" issue.

paul-guo-:
+1

src/backend/utils/misc/guc.c
jiny2: https://issues.apache.org/jira/browse/HAWQ-1266 this jira is created for discussion. Thanks @paul-guo- 's thorough think.

PR #1080, opened 2017-01-11
HAWQ-1257. Prompt all tables which user doesn't have right once

interma:


interma:
wcl's commit also be included, very strange, I will open a new PR.

PR #1082, opened 2017-01-11
HAWQ-1256. non-superuser connect to db will do aclcheck through range…

stanlyxiang:
…r and curl handle haven't been initialized.


zhangh43:
+1

interma:
+1

PR #1087, opened 2017-01-16
HAWQ-1275. Check build-in catalogs, tables and functions in native aclcheck.

zhangh43:
We plan to do privilege check in hawq side for build-in catalogs, tables and functions. The reason is that Ranger mainly manage the user data, but build-in catalogs and tables are not related to user data(note that some of them contain statistics information of user data such as catalog table pg_aoseg_*).

### DIFF
--- a/src/backend/catalog/aclchk.c
+++ b/src/backend/catalog/aclchk.c
@@ -2669,6 +2669,13 @@ List *getActionName(AclMode mask)
 
 bool fallBackToNativeCheck(AclObjectKind objkind, Oid obj_oid, Oid roleid)
 {
+  /* get the latest information_schema_namespcace_oid. Since caql access heap table
+   * directly without aclcheck, this function will not be called recursively
+   */
+  if (information_schema_namespcace_oid == 0)
+  {
+	  information_schema_namespcace_oid = (int)get_namespace_oid("information_schema");
+  }
   /*for heap table, we fall back to native check.*/
   if (objkind == ACL_KIND_CLASS)
   {
@@ -2681,7 +2688,7 @@ bool fallBackToNativeCheck(AclObjectKind objkind, Oid obj_oid, Oid roleid)
   else if (objkind == ACL_KIND_NAMESPACE)
   {
 	/*native check build-in schemas.*/
-    if (obj_oid == PG_CATALOG_NAMESPACE || obj_oid == PG_INFORMATION_SCHEMA_NAMESPACE
+    if (obj_oid == PG_CATALOG_NAMESPACE || obj_oid == information_schema_namespcace_oid
     		|| obj_oid == PG_AOSEGMENT_NAMESPACE || obj_oid == PG_TOAST_NAMESPACE
 			|| obj_oid == PG_BITMAPINDEX_NAMESPACE)
     {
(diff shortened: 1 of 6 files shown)


src/backend/catalog/aclchk.c
```diff
@@ -2670,28 +2670,50 @@ List *getActionName(AclMode mask)
...
-    if(relstorage == 'h')
+    if (relstorage == 'h')
     {
       return true;
     }
   }
+  else if (objkind == ACL_KIND_NAMESPACE)
+  {
+	//native check build-in schemas.
+    if (obj_oid == PG_CATALOG_NAMESPACE || obj_oid == PG_INFORMATION_SCHEMA_NAMESPACE
```
interma: Is there need to add gp_toolkit schema? Seems need?

_`gp_toolkit` is an administrative schema that contains external tables, views, and functions that you can access with SQL commands. All database users can access gp_toolkit to view and query the system log files and other system metrics._

refer: http://hdb.docs.pivotal.io/201/hawq/ddl/ddl-schema.html

src/backend/catalog/aclchk.c
```diff
@@ -2670,28 +2670,50 @@ List *getActionName(AclMode mask)
...
     char relstorage = get_rel_relstorage(obj_oid);
-    if(relstorage == 'h')
+    if (relstorage == 'h')
     {
       return true;
     }
   }
+  else if (objkind == ACL_KIND_NAMESPACE)
+  {
+	//native check build-in schemas.
```
linwen: should use this kind of comments: /* */

src/backend/catalog/aclchk.c
```diff
@@ -2670,28 +2670,50 @@ List *getActionName(AclMode mask)
...
+    if (obj_oid == PG_CATALOG_NAMESPACE || obj_oid == PG_INFORMATION_SCHEMA_NAMESPACE
+    		|| obj_oid == PG_AOSEGMENT_NAMESPACE || obj_oid == PG_TOAST_NAMESPACE
+			|| obj_oid == PG_BITMAPINDEX_NAMESPACE)
+    {
+      return true;
+    }
+  }
+  else if (objkind == ACL_KIND_PROC)
+  {
+	//native check functions under build-in schemas.
```
linwen: should use this kind of comments: /* */ and fix indent.

src/backend/catalog/aclchk.c
```diff
@@ -2670,28 +2670,50 @@ List *getActionName(AclMode mask)
...
+  }
+
   return false;
 }
 
 bool fallBackToNativeChecks(AclObjectKind objkind, List* table_list, Oid roleid)
 {
-  //for heap table, we fall back to native check.
-  if(objkind == ACL_KIND_CLASS)
+  //we only have range table here
```
linwen: should use this kind of comments: /* */

src/include/catalog/pg_namespace.h
```diff
@@ -123,7 +123,7 @@ DESCR("Standard public schema");
 DATA(insert OID = 6104 ( "pg_aoseg" PGUID _null_ 0));
 DESCR("Reserved schema for Append Only segment list and eof tables");
 #define PG_AOSEGMENT_NAMESPACE 6104
-
+#define PG_INFORMATION_SCHEMA_NAMESPACE 10671
```
ictmalili: Is the oid for information_schema schema fixed?  I thought it was created during hawq init procedure, so the oid might be not fixed at the value 10671.   Can we get it firstly as a global variable by running "select oid from pg_namespace where nspname = 'information_schema'"  ?

src/backend/catalog/aclchk.c
ictmalili: Yes, I think so

linwen:
+1

src/backend/catalog/aclchk.c
zhangh43: I think gp_toolkit should be managed by Ranger.

src/backend/catalog/aclchk.c
zhangh43: fixed

src/include/catalog/pg_namespace.h
zhangh43: I use caql to get the latest information_schema_namespace oid instead

interma:
+1

stanlyxiang:
+1

PR #1088, opened 2017-01-16
HAWQ-1276. The error message is not friendly when ranger plugin servi…

stanlyxiang:
…ce is unavailable.

### DIFF
--- a/src/backend/libpq/rangerrest.c
+++ b/src/backend/libpq/rangerrest.c
@@ -353,18 +353,18 @@ static size_t write_callback(char *contents, size_t size, size_t nitems,
 	int original_size = curl->response.buffer_size;
 	while(curl->response.response_size + realsize >= curl->response.buffer_size)
 	{
-		/*double the buffer size if the buffer is not enough.*/
+		/* double the buffer size if the buffer is not enough.*/
 		curl->response.buffer_size = curl->response.buffer_size * 2;
 	}
 	if(original_size < curl->response.buffer_size)
 	{
-		/* our repalloc is not same as realloc, repalloc's first param(buffer) can not be NULL */
+		/* repalloc is not same as realloc, repalloc's first parameter cannot be NULL */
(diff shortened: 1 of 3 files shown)


stanlyxiang:
cc @interma @zhangh43 @linwen @ictmalili @wcl14

src/backend/libpq/rangerrest.c
```diff
@@ -366,7 +365,7 @@ static size_t write_callback(char *contents, size_t size, size_t nitems,
 	if (curl->response.buffer == NULL)
 	{
 		/* out of memory! */
-		elog(WARNING, "not enough memory for Ranger response");
+		elog(WARNING, "cannot allocate memory for ranger response");
```
wcl14: Why do you modify the out of memory error message. I think the original one is more specified, and may be better.

interma:
+1

src/backend/libpq/rangerrest.c
stanlyxiang: hmm, IMO , not enough memory is not the only error that will occur. There maybe other errors  in the repalloc implementation. So I changed the messages.

wcl14: So should the comment be consistent with the code?

stanlyxiang: sure. fixed.

src/backend/libpq/rangerrest.c
```diff
@@ -294,7 +294,7 @@ static json_object *create_ranger_request_json(List *request_list, List *result_
 				break;
 			}
 			default:
-				elog(ERROR, "unrecognized objkind: %d", (int) kind);
+				elog(ERROR, "unrecognized object kind : %d", (int) kind);
```
linwen: since this is a message to users, we should map this "kind" to string, so that users can know which kind is not supported.

stanlyxiang:
merged into master. closed.

PR #1089, opened 2017-01-16
HAWQ-1243. Add suffix name for ranger restful service.

zhangh43:
change the default RPS RESTFUL service suffix to rps


stanlyxiang:
+1

linwen:
+1

PR #1092, opened 2017-01-17
HAWQ-1203. Ranger Plugin Service Implementation. (with contributions …

denalex:
…by Lav Jain and Leslie Chang)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,12 @@ objfiles.txt
 .pydevproject
 .cproject
 .settings
+.classpath
+
+# IDEA Project
+*.iml
+.idea
+
 # Ctags
 **/tags
 
@@ -54,6 +60,7 @@ env.sh
 ext/
 plr.tgz
 autom4te.cache/
+**/target
 
 # coverage
 *.gcda
--- a/ranger-plugin/admin-plugin/pom.xml
+++ b/ranger-plugin/admin-plugin/pom.xml
@@ -1,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.hawq</groupId>
+    <artifactId>ranger-plugin-admin</artifactId>
+    <packaging>jar</packaging>
+    <name>HAWQ Ranger Admin Plugin</name>
+    <description>HAWQ Ranger Admin Plugin</description>
+    <parent>
+        <groupId>org.apache.hawq</groupId>
+        <artifactId>ranger-plugin</artifactId>
+        <version>2.1.0.0</version>
+        <relativePath>..</relativePath>
+    </parent>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>copy-dependencies</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/lib</outputDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+
+        <!-- Compilation Dependencies -->
+        <dependency>
+            <groupId>org.apache.ranger</groupId>
+            <artifactId>ranger-plugins-common</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>log4j</groupId>
+            <artifactId>log4j</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>postgresql</groupId>
+            <artifactId>postgresql</artifactId>
+        </dependency>
+
+        <!-- Test Dependencies -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-module-junit4</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.powermock</groupId>
+            <artifactId>powermock-api-mockito</artifactId>
+        </dependency>
+
+    </dependencies>
+
+</project>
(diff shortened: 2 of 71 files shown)


ictmalili:
@linwen @zhangh43 @stanlyxiang @interma @wcl14 @xunzhang Please review this PR and provide some comments. Thanks :)

ranger-plugin/admin-plugin/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.hawq</groupId>
+    <artifactId>ranger-plugin-admin</artifactId>
+    <packaging>jar</packaging>
+    <name>HAWQ Ranger Admin Plugin</name>
+    <description>HAWQ Ranger Admin Plugin</description>
+    <parent>
+        <groupId>org.apache.hawq</groupId>
+        <artifactId>ranger-plugin</artifactId>
+        <version>2.1.0.0</version>
```
ictmalili: +1 for this comment.  I guess this is the version number for apache hawq? If it is, I think we don't need record this version?

.gitignore
```diff
@@ -41,6 +41,12 @@ objfiles.txt
 .pydevproject
 .cproject
 .settings
+.classpath
+
+# IDEA Project
+*.iml
+.idea
```
ictmalili: Are these necessary for each one's .gitignore file?

ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/HawqClient.java
```diff
@@ -0,0 +1,314 @@
...
+        PreparedStatement preparedStatement = conn.prepareStatement(query);
+        preparedStatement.setString(1, userInput.equals(WILDCARD) ? "%" : (userInput + "%"));
+        return preparedStatement;
+    }
+
+    private void closeResultSet(ResultSet rs) {
+        try {
+            if (rs != null) rs.close();
+        } catch (Exception e) {
+            // ignore
```
ictmalili: What shall we do for these exceptions? Do we need log it out?

ranger-plugin/admin-plugin/pom.xml
denalex: we want to version RPS (and it RPM) with the same version as HAWQ as RPS is not a standalone product. Maven requires version in pom.xml files, so we choose the one that HAWQ has and will need to update it after Apache releases.

.gitignore
denalex: this removes IDEA editor artifacts from the list of potential GIT changes as those artifacts should never be committed.

ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/HawqClient.java
denalex: There is nothing we can do, this is a standard practice to ignore closing errors. We can log them, but that would not be of a great value to anybody.

ranger-plugin/conf/ranger-servicedef-hawq.json
```diff
@@ -0,0 +1,287 @@
...
+  [
+    {
+      "itemId": 1,
+      "name": "database",
+      "type": "string",
+      "level": 10,
+      "parent": "",
+      "mandatory": true,
+      "lookupSupported": true,
+      "recursiveSupported": false,
```
ictmalili: What are the meanings of recursiveSupported and excludesSupported? Why do we set the values to false and true respectively?

ranger-plugin/integration/admin/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+    <groupId>org.apache.hawq</groupId>
+    <artifactId>ranger-plugin-integration-admin</artifactId>
+    <packaging>jar</packaging>
+    <name>HAWQ Ranger Plugin - Integration Tests for Admin Plugin</name>
+    <description>HAWQ Ranger Plugin - Integration Tests for Admin Plugin</description>
+
+    <parent>
+        <groupId>org.apache.hawq</groupId>
+        <artifactId>ranger-plugin-integration</artifactId>
+        <version>2.1.0.0</version>
```
ictmalili: Again for the version number here, which version number we should keep it?

ranger-plugin/scripts/register_hawq.sh
```diff
@@ -0,0 +1,217 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+function usage() {
+  echo "USAGE: register_hawq.sh -r ranger_host:ranger_port -u ranger_user -p ranger_password -h hawq_host:hawq_port -w hawq_user -q hawq_password"
```
ictmalili: What's the meaning for this script? Will it be exposed directly out to users or is it just an internal script?  If it will be directly exposed out to users, can we change it to a more meaningful name?  It looks a little confusing now.

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqAuthorizer.java
```diff
@@ -0,0 +1,263 @@
...
+     */
+    private boolean authorizeResourcePrivilege(RangerAccessResource rangerResource, String accessType, String user, Set<String> userGroups) {
+
+        Map<String, String> resourceMap = rangerResource.getAsMap();
+        String database = resourceMap.get(HawqResource.database.name());
+        String schema = resourceMap.get(HawqResource.schema.name());
+        int resourceSize = resourceMap.size();
+
+        // special handling for non-leaf policies
+        if (accessType.equals(HawqPrivilege.create.name()) && database != null && schema == null && resourceSize == 1) {
```
ictmalili: What's the meaning of resourceSize valued as 1 and 2? Could you add some comments in the code here? Thanks

ictmalili:
+1 except above comments I have made.

zhangh43:
+1

ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/HawqClient.java
ictmalili: OK. Make sense

linwen:
+1

ranger-plugin/conf/ranger-servicedef-hawq.json
denalex: @ictmalili -- these are not well documented, we copied the definition from Hive. My understanding is that:
-- recursiveSupported would allow to define a policies as recursive which is used for HDFS access policies where a policy on a given folder, if declared as recursive, will be applied to all the files and subfolders inside a given folder. This is not applicable to us, so we do not allow creation of "recursive" policies for HAWQ.
-- excludesSupported will allow a resource specified in the policy to be "excluded". Ranger Admin UI will render the box to the right of the resource to be toggled between "included" and "excluded". I am not exactly sure what effect will that have on the policy.

ranger-plugin/integration/admin/pom.xml
denalex: We will keep the version the same as for the rest of HAWQ code. Perhaps a separate discussion will be useful to decide how we should manage it between releases.

ranger-plugin/scripts/register_hawq.sh
denalex: @ictmalili -- this script must be run to register HAWQ service definition with Ranger Admin and create default instance of HAWQ service within Ranger that points to HAWQ Active master for the lookup. For now it will have to be run manually by users once, later we can think of automating it with Ambari.
What name would you suggest ?

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqAuthorizer.java
denalex: resourceSize reflects how many tuples of key:value were received from HAWQ in resource element of JSON request. For example: "resource" : {"database":"foo"} is only 1 entry while "resource" : {"database":"foo", "schema":"bar"} are 2 entries. We have to analyze this for our special handling of create-schema and usage-schema permissions.

denalex:
Thank you everyone for reviewing the PR. I will commit it tomorrow morning (PST time) and will create separate tasks for:
- running shell scripts via the tool, once all shell scripts are finalized
- discussing version management (which should be in master, how to change them on the branch before release, how to change in master after branching)

Please create other JIRAs if needed. The work on RPS codebase will continue after this merge via further additional JIRAs.

ranger-plugin/scripts/register_hawq.sh
ictmalili: @denalex May we change it adding indication the meaning to register to Ranger, such as  register_hawq_to_ranger.sh, or anything else?   Because HAWQ now has a management tool named as hawq register

denalex: @ictmalili -- ok, makes sense, will rename the script as part of management scripts JIRA.

PR #1096, opened 2017-01-19
HAWQ-1282. Shared Input Scan may result in endless loop.

zhangh43:
There are residual process after running some queries. Through the call stack, we find that there is an endless loop in function writer_wait_for_acks() in shared input scan.
This PR includes:
1 add max retry times to avoid this problem.
2 fix file handler leak in retry_read() and retry_write() of shared input scan.

### DIFF
--- a/src/include/executor/nodeShareInputScan.h
+++ b/src/include/executor/nodeShareInputScan.h
@@ -31,7 +31,6 @@
 
 #include "nodes/execnodes.h"
 
-#define MAX_RETRY_TIMES 5
 extern int ExecCountSlotsShareInputScan(ShareInputScan* node);
 extern ShareInputScanState *ExecInitShareInputScan(ShareInputScan *node, EState *estate, int eflags);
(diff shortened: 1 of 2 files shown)


src/include/executor/nodeShareInputScan.h
```diff
@@ -30,6 +30,8 @@
 #define NODESHAREINPUTSCAN_H
 
 #include "nodes/execnodes.h"
+
+#define MAX_RETRY_TIMES 5
```
ictmalili: Is it appropriate to set the value as 5? What if the network environment is bad?

zhangh43: I think writer_wait_for_acks is not a must one. and for orca is on, writer_wait_for_acks is not called at all.

linwen:
+1

src/include/executor/nodeShareInputScan.h
zhangh43: I remove timeout retry max times.

PR #1097, opened 2017-01-19
HAWQ-1285. resource manager outputs uninitialized string as host name

jiny2:
When the segstat instance is created from YARN cluster report, the hostname field is not set, so should now output it directly into log. This bug causes unreadable string output in log file.


jiny2:
@wcl14 @stanlyxiang please review.

jiny2:
@huor for review

stanlyxiang:
+1

jiny2:
Applied comment from @huor for the log content

PR #1098, opened 2017-01-20
HAWQ-1286. Reduce unnecessary calls of namespace check when run \d

interma:
## Backgroud:
After HAWQ-1279 is done, current schema is no cached in current session.
But it cause too many calls of namespace check to send in run \d , most of them are unnecessary.

## Example
```
\d:
    select version()
    SELECT n.nspname as \"Schema\",\n  c.relname as \"Name\",\n  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' END as \"Type\",\n  pg_catalog.pg_get_userbyid(c.relowner) as \"Owner\"\nFROM pg_catalog.pg_class c\n     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace\nWHERE c.relkind IN ('r','v','S','')\n      AND n.nspname <> 'pg_catalog'\n      AND n.nspname <> 'information_schema'\n      AND n.nspname !~ '^pg_toast'\n  AND pg_catalog.pg_table_is_visible(c.oid)\nORDER BY 1,2;
        recomputeNamespacePath()
        recomputeNamespacePath()
        .... recompute many times in this long select sql
```

## Solution
Store the last sql query, only recompute if the query is changed.

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -48,8 +48,10 @@
 #include "utils/memutils.h"
 #include "utils/syscache.h"
 #include "utils/guc.h"
+#include "utils/hsearch.h"
 #include "cdb/cdbvars.h"
 #include "tcop/utility.h"
+
 
 /*
  * The namespace search path is a possibly-empty list of namespace OIDs.
@@ -113,6 +115,8 @@
  * namespaceUser is the userid the path has been computed for.
  */
 
+extern const char *debug_query_string;
+
 static List *namespaceSearchPath = NIL;
 
 static Oid	namespaceUser = InvalidOid;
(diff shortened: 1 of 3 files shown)


interma:
@linwen @zhangh43 @stanlyxiang to review, thanks!

linwen:
Would you please copy these information to its jira? Thanks!

interma:
@linwen Done! 
Thanks for reminder.

zhangh43:
+1.

linwen:
+1

PR #1099, opened 2017-01-24
HAWQ-1291. Fix the name of privilege when create temp table.

zhangh43:
When we do ranger check of query "create temp table", we need to PASS privilege "TEMP" to RPS instead of "CREATE_TEMP". Or RPS can not recognise it.


linwen:
+1

PR #1106, opened 2017-02-04
HAWQ-1312. Forbid grant/revoke command in HAWQ side once Ranger is co…

linwen:
HAWQ-1312. Forbid grant/revoke command in HAWQ side once Ranger is configured.

### DIFF
--- a/src/backend/catalog/aclchk.c
+++ b/src/backend/catalog/aclchk.c
@@ -298,6 +298,14 @@ ExecuteGrantStmt(GrantStmt *stmt)
 	bool		added_objs = false;
 
 	/*
+	 * Don't allow GRANT/REVOKE if in ranger mode.
+	 */
+	if (aclType == HAWQ_ACL_RANGER)
+	{
+		elog(ERROR, "GRANT/REVOKE is not allowed in ranger mode");
+	}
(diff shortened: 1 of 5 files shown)


src/backend/catalog/aclchk.c
```diff
@@ -298,6 +298,14 @@ ExecuteGrantStmt(GrantStmt *stmt)
...
 
 	/*
+	 * Don't allow GRANT/REVOKE if in ranger mode.
+	 */
+	if (aclType == HAWQ_ACL_RANGER)
+	{
+		elog(ERROR, "GRANT/REVOKE is not allowed in ranger mode");
+	}
+
+	/*
```
ictmalili: I don't think we should forbid all the objects for grant/revoke. Given that we decide to leave catalog table/owner check in HAWQ side, we should allow grant/revoke for these operations.  What we need to forbid is what we have supported in Ranger UI configuration.

linwen: Thank you for pointing out!

zhangh43:
+1

interma:
+1

stanlyxiang:
+1

PR #1108, opened 2017-02-08
HAWQ-1317. Port "Fix some regex issues with out-of-range characters and large char ranges" from pg

amyrazz44:
@huor  @paul-guo- , could you please review this pull request for me, thank you.

### DIFF
--- a/src/backend/regex/regc_lex.c
+++ b/src/backend/regex/regc_lex.c
@@ -792,13 +792,13 @@ lexescape(struct vars * v)
 			break;
 		case CHR('u'):
 			c = lexdigits(v, 16, 4, 4);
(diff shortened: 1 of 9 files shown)


jiny2:
+1 LGTM

huor:
+1

amyrazz44:
We just port part of this pg bug fix for now which is fix out-of-range characters part, about the cancel_requested part, we will port in the future. 
Because postgres 9.6 adds some extra functions in regex code, e.g. cancel_requested logic. And this pg bug fix includes cancel_requested code. 
So, need more efforts to investigate whether we should add extra functions to regex code in the other commit.

PR #1109, opened 2017-02-08
HAWQ-1298 Refactored RPS management scripts

denalex:


ranger-plugin/scripts/rps.sh
```diff
@@ -20,41 +20,71 @@
...
+    if [ $? -ne 0 ]; then
+      fail ${1}
+    fi
+}
+
+function wait_until_server_started() {
+    echo -n "Waiting for Hawq Ranger Plugin Service to start ."
+    local retries="20"
+    local n=0
+    until $(curl -s --output /dev/null --fail ${RPS_URL}/version); do
```
zhangh43: The --fail option should be removed, its return code is 22.
Since the logic here is to repeat waiting if rps is not ready. When RPS is ready, return code should be 0, and until statement returns.
curl -s --output /dev/null --fail ${Right_RPS_URL}/version
echo $? => 22
curl -s --output /dev/null  ${Right_RPS_URL}/version
echo $? => 0
curl -s --output /dev/null --fail ${Wrong_RPS_URL}/version
echo $? => 7
curl -s --output /dev/null ${Right_RPS_URL}/version
echo $? => 7

ranger-plugin/integration/service/src/test/java/org/apache/hawq/ranger/integration/service/tests/common/ServiceTestBase.java
```diff
@@ -67,7 +67,8 @@
     private static final String RANGER_URL = String.format("http://%s:%s/service/public/v2/api", RANGER_HOST, RANGER_PORT);
     private static final String RANGER_POLICY_URL = RANGER_URL + "/policy";
 
-    private static final int POLICY_REFRESH_INTERVAL = 6000;
+    private static final String POLICY_WAIT_INTERVAL_PROP_NAME = "policy.wait.interval.ms";
```
zhangh43: could you explain what is the difference between ranger.plugin.hawq.policy.pollIntervalMs and policy.wait.interval.ms?  And in which xml file to config policy.wait.interval.ms?

ranger-plugin/scripts/rps.sh
denalex: It works, I think when you tested you had RPS shut down. Curl returns error 7 if it can't connect to host. When RPS is up, the return is 0. See the trace below:

bash-4.1# curl -s --output /dev/null --fail http://localhost:8432/rps/version
bash-4.1# echo $?
7
bash-4.1# curl --fail http://localhost:8432/rps/version
curl: (7) couldn't connect to host
bash-4.1# ps aux | grep rps
root      2340  0.0  0.0   6500   464 ?        R+   04:58   0:00 grep rps
bash-4.1# ./rps.sh start
Using CATALINA_BASE:   /usr/local/hawq_2_1_0_0/ranger/plugin-service
Using CATALINA_HOME:   /usr/lib/bigtop-tomcat
Using CATALINA_TMPDIR: /usr/local/hawq_2_1_0_0/ranger/plugin-service/temp
Using JRE_HOME:        /usr/lib/jvm/java-1.7.0-openjdk.x86_64
Using CLASSPATH:       /usr/local/hawq_2_1_0_0/ranger/etc:/usr/lib/bigtop-tomcat/bin/bootstrap.jar
Using CATALINA_PID:    /usr/local/hawq_2_1_0_0/ranger/plugin-service/work/rps.pid
Waiting for Hawq Ranger Plugin Service to start ..
Hawq Ranger Plugin Service is available at http://localhost:8432/rps
bash-4.1# curl -s --output /dev/null --fail http://localhost:8432/rps/version
bash-4.1# echo $?
0
bash-4.1#

ranger-plugin/integration/service/src/test/java/org/apache/hawq/ranger/integration/service/tests/common/ServiceTestBase.java
denalex: ranger.plugin.hawq.policy.pollIntervalMs is a real property used by the service and determines the frequency of policy refreshes performed by Ranger Plugin.

policy.wait.interval.ms is the property used only by integration test suite to wait between updating policy in Ranger and trying to evaluate the request. In our tests we currently set wait.interval = 3 * pollInterval, even that sometimes causes random test failures when the system is under heavy load. We can optimize this further, if needed.

ranger-plugin/scripts/rps.sh
zhangh43: Got it. When server is down, curl --fail will return 22. When server is up, it return 0 successfully.

ranger-plugin/integration/service/src/test/java/org/apache/hawq/ranger/integration/service/tests/common/ServiceTestBase.java
zhangh43: Thanks

zhangh43:
+1

PR #1110, opened 2017-02-09
HAWQ-1318. Fix the bug of cannot start/stop master successfully if ranger is enable and with a wrong RPS address

linwen:
Please review, thanks!


zhangh43:
+1

PR #1111, opened 2017-02-09
HAWQ-1319. Expand pom.xml file for src/test directory structure.

edespino:
This has two goals:
* Catch missing ASF header from java files.
* Identify archive (jar) files. NOTE: If found, this does not cause
  RAT to fail.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,51 @@
               <exclude>src/pl/**/*</exclude>
               <exclude>src/port/*</exclude>
               <exclude>src/template/*</exclude>
-              <exclude>src/test/**/*</exclude>
+              <exclude>src/test/**/*.c</exclude>
+              <exclude>src/test/**/*.cpp</exclude>
+              <exclude>src/test/**/*.h</exclude>
+              <exclude>src/test/**/*.in</exclude>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -129,7 +129,51 @@
               <exclude>src/pl/**/*</exclude>
               <exclude>src/port/*</exclude>
               <exclude>src/template/*</exclude>
-              <exclude>src/test/**/*</exclude>
+              <exclude>src/test/**/*.c</exclude>
```
paul-guo-: I'm wondering whether we should not exclude some of the text files (source files, makefiles, etc) which we could add header files.  I remember we have a auto-script for header adding. @linwen

huor:
@edespino, the source file in feature suite should be included in pom. Please create another jira to add the license header for the source file once the pom is updated.

huor:
@edespino, please close this pr and we will update pom and add license header in source file in https://issues.apache.org/jira/browse/HAWQ-1323

edespino:
Closing this PR as @huor and team will be updating the pom.xml.  They will be adding ASF headers as appropriate to additional source files.

-=e

PR #1112, opened 2017-02-10
HAWQ-1319. Add ASF header to java files and remove jar files in function suite

huor:


huor:
@edespino and @paul-guo-, please review.

src/test/feature/UDF/TestUDF.cpp
```diff
@@ -200,6 +218,10 @@ TEST_F(TestUDF, TestUDFPljava)
 	// run test if pljava language is enabled
 	if (util.getQueryResult("SELECT lanname FROM pg_language WHERE lanname = 'java'") == "java")
 	{
+		// prepare jar files
+		hawq::test::Command cmd("cd " + d_feature_test_root + "/UDF/sql/; javac PLJavaAdd.java; jar cf PLJavaAdd.jar PLJavaAdd.class");
+		cmd.run();
+
```
paul-guo-: I'm wondering we should add a chapter in README.md for requirement, e.g. we need java dev package for test running here. Others look good.

huor: java is dependency for hawq as well, it is checked by configure.

PR #1114, opened 2017-02-10
HAWQ-1320. Remove reference to PXF rpm version in source

shivzone:


pxf/Makefile
```diff
@@ -38,6 +38,10 @@ ifneq "$(VENDOR)" ""
     BUILD_PARAMS+= -Pvendor="$(VENDOR)"
 endif
 
+ifneq "$(PXF_VERSION)" ""
+    BUILD_PARAMSi+= -Pversion="$(PXF_VERSION)"
```
ljainpivotalio: There is an extra i after BUILD_PARAMS.

ljainpivotalio:
Please feel free to merge when done

pxf/Makefile
shivzone: good catch

pxf/Makefile
```diff
@@ -38,6 +38,10 @@ ifneq "$(VENDOR)" ""
     BUILD_PARAMS+= -Pvendor="$(VENDOR)"
 endif
 
+ifneq "$(PXF_VERSION)" ""
```
sansanichfb: Shouldn't we fail if there is no value for PXF_VERSION available?

shivzone: Not really. We should still be able to run simple dev builds which don't require PXF version

PR #1115, opened 2017-02-13
HAWQ-1324. Fixed crash at query cancel, signal hanler cannot call uns…

liming01:
…afe elog()

1) The fix cannot completely fix this kinds of crash, it just reduce the possibility of this kind of crash. The complete fix is to move ProcessInterrupts() calling out of StatementCancelHandler().
2) The log can be removed because afterward in all 3 branches, hawq will print out similar message using ereport().

Thanks.


wengyanqing:
LGTM

PR #1119, opened 2017-02-14
HAWQ-1328. Add deny and exclude policy template for hawq service in ranger

zhangh43:
add option "enableDenyAndExceptionsInPolicies": "true" in ranger-servicedef-hawq.json to open deny and exclude policy template by default.

### DIFF
--- a/ranger-plugin/conf/ranger-servicedef-hawq.json
+++ b/ranger-plugin/conf/ranger-servicedef-hawq.json
@@ -4,6 +4,7 @@
   "label": "HAWQ",
   "description": "HAWQ",
   "guid": "1ebb27ee-549a-401d-85ab-818342ca54af",
+  "options": {"enableDenyAndExceptionsInPolicies": "true”},
   "resources":
   [
     {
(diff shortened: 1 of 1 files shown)


zhangh43:
@denalex @linwen Please review it.

zhangh43:
Thanks @denalex , the quote typo is fixed.
According to the Ranger document, Ranger will traverse all the deny conditions firstly to determine whether to disallow the access. And then traverse all the positive conditions.

ictmalili:
@denalex What do you mean by "even though with deny conditions access might not be determined, the RPS will continue to interpret it as access not allowed"?  Doesn't it inherit the processing the Ranger Basic Plugin?

denalex:
@ictmalili -- yes, it inherits, but RPS does not check isAccessDetermined() flag on the BasePluginResponse. It only checks isAccessAllowed() flag and only if it is "true" it reports to HAWQ that access is allowed, all other cases are treated as access is not allowed.

ictmalili:
OK. Got your point, Thanks @denalex 

This commit LGTM. +1

PR #1122, opened 2017-02-14
HAWQ-1314. Added upgrade for function pxf_get_item_fields

sansanichfb:
Added upgrade section for function pxf_get_item_fields, to upgrade its signature to latest.
@zhangh43

### DIFF
--- a/tools/bin/hawqupgrade
+++ b/tools/bin/hawqupgrade
@@ -49,13 +49,25 @@ def option_parser():
     parser.add_option('-U', '--user', help='username of the target DB')
     return parser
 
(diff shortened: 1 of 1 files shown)


tools/bin/hawqupgrade
```diff
@@ -49,30 +49,21 @@ def option_parser():
     parser.add_option('-U', '--user', help='username of the target DB')
     return parser
 
-def register_func(options, args):
+# function which runs sql statements, passed in args[1], args[2] in all databases
```
shivzone: Just update the comment as function that runs upgrade related sql statements

PR #1123, opened 2017-02-15
HAWQ-1331. Update hawq and pxf versions

shivzone:


ljainpivotalio:
Please create 2.2.json in tools/bin/gppylib/data as a copy of 2.1.json

shivzone:
@ljainpivotalio i'll update it. ideally we shouldn't be maintaining version specific files for releases in git

PR #1124, opened 2017-02-15
HAWQ-1330. More ASF header updates (including pom.xml update)

edespino:
Here are more ASF license header updates.  I have verified the change in a dev pipeline ()https://hawq.ci.pivotalci.info/teams/main/pipelines/hdb-espino).  I would like to push this change for the upcoming Apache 2.2.0.0-incubating RC1 release.

-=e

Reviewer requests: @amyrazz44 @paul-guo- @radarwave @stanlyxiang @huor

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,39 +129,88 @@
               <exclude>src/pl/**/*</exclude>
               <exclude>src/port/*</exclude>
               <exclude>src/template/*</exclude>
-              <exclude>src/test/**/*.in</exclude>
-              <exclude>src/test/**/*.mdp</exclude>
+
+              <!-- src/test/bench -->
+
+              <exclude>src/test/bench/query*</exclude>
+
+              <!-- PostgreSQL src/test/regress -->
+              
+              <exclude>src/test/regress/**/*.in</exclude>
+              <exclude>src/test/regress/**/*.mdp</exclude>
+              <exclude>src/test/regress/**/*.xml</exclude>
+              <exclude>src/test/regress/data/fixedwidth/fixedwidth_3chars_line_delimiter.tbl</exclude>
+              <exclude>src/test/regress/data/fixedwidth/fixedwidth_big.tbl</exclude>
+              <exclude>src/test/regress/data/fixedwidth/fixedwidth_no_line_delimiter.tbl</exclude>
+              <exclude>src/test/regress/data/fixedwidth/fixedwidth_out.tbl</exclude>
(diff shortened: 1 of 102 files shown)


pom.xml
```diff
@@ -129,39 +129,88 @@
...
+
+              <exclude>src/test/feature/ExternalSource/data/nation.tbl</exclude>
+              <exclude>src/test/feature/ExternalSource/data/nation_error50.tbl</exclude>
+              <exclude>src/test/feature/ExternalSource/data/region.tbl</exclude>
+              <exclude>src/test/feature/ManagementTool/test_hawq_register_hawq.paq</exclude>
+              <exclude>src/test/feature/ManagementTool/test_hawq_register_hive.paq</exclude>
+              <exclude>src/test/feature/ManagementTool/data/parquet200/dat.paq</exclude>
+              <exclude>src/test/feature/ManagementTool/data/parquet200sum/a.paq</exclude>
+              <exclude>src/test/feature/ManagementTool/data/parquet200sum/b.paq</exclude>
+              <exclude>src/test/feature/ManagementTool/test_hawq_register_data_types.paq</exclude>
```
paul-guo-: Maybe we should use wildcat for .paq and .tbl files here. I've been thinking of having an extension standard in our dev/test code.

edespino: I am trying to limit our use of the wildcards thus forcing us to be as explicit as possible. The dev community will come to understand the importance of maintaining ASF headers as needed. Additionally, we have Apache RAT in place to allow a developer to catch potential issues as early as possible.

paul-guo-: I think we should limit using wildcard for directory, etc, but not for file extensions.

edespino: I'm afraid we will not question files which do not have ASF headers if we use directory wild cards so freely. As an example, the following was excluded but should have had a header:

<exclude>src/test/**/init_file</exclude>

Thank you for helping us think through this. It helps us all in the end.  For now, I would like to be explicit in our exclusions. We can revisit if it becomes a significant maintenance issue.

huor:
@edespino, is there any rule indicate which type of files need to be excluded in the RAT?

edespino:
@huor - there are no such rules. Our use is dependent on our needs. My main goal is to avoid accidentally excluding files from being scanned.

ljainpivotalio:
LGTM

huor:
@edespino, thanks for the update. +1

PR #1126, opened 2017-02-15
HAWQ-1334. QD thread should set error code if failing so that the mai…

paul-guo-:
…n process for the query could exit soon


huor:
+1 while the func name need to be revised.

paul-guo-:
Merged. (Discussed with ruilong. The filename executormgr_seterrcode_if_needed does not need a modification while move the NULL pointer check into the function.

PR #1128, opened 2017-02-16
HAWQ-1338. Fixed writer process doesn't exit nicely in some case

liming01:
(1) From the bt in the core file, it seems that SuppressPanic is true, we still report panic even at process exist function. We should suppress the panic if there is no side effect. 
(2) Now hawq stop utility stop standby after master stopped successfully or failed to stop with 10 minutes timeout, so it seems no need to change.


wengyanqing:
LGTM

amyrazz44:
LGTM +1

PR #1132, opened 2017-02-17
HAWQ-1343. - Install pip and minor formatting change.

edespino:
The latest Travis CI OS X builds are failing (https://travis-ci.org/apache/incubator-hawq/builds/202425654) because "pip" command is not found.  This change will install it.

Error:
$ sudo pip install pycrypto
sudo: pip: command not found

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,24 +27,26 @@ before_install:
 
 install:
   - brew reinstall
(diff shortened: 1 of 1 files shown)


ljainpivotalio:
brew install python

PR #1134, opened 2017-02-20
HAWQ-1345. Fixed relfile path bug: catalog and hdfs relfile have diff…

liming01:
…erent rule

Add FormRelfilePath() to generate the rel file name which support not only catalog relfile but also relfile on hdfs, it can detect according :
- if is on hdfs (which has prefix "hdfs://"), the rel file rule is(relation_oid as directory):  relation_oid/seg_num  .
- if is on local disk, the rel file rule is (relation_oid as file name):  relation_oid.seg_num or relation_oid (when seg_num==0)

### DIFF
--- a/src/backend/access/appendonly/aomd.c
+++ b/src/backend/access/appendonly/aomd.c
@@ -101,12 +101,7 @@ FormatAOSegmentFileName(
 	
 	*fileSegNo = pseudoSegNo;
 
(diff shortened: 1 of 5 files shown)


amyrazz44:
LGTM +1

wengyanqing:
LGTM

PR #1138, opened 2017-02-23
HAWQ-1356. Add waring when user does not have usage privilege of namespace

zhangh43:
Currently, in Ranger mode, when a user doesn't have usage privilege on public schema, she cannot see any tables, functions, schemas on public schema. When she login into database and run a query "select * from a"; The error message would be "table a doesn't exist", which makes user confuse.
This fix add a warning to tell user that she does not have the usage privilege of a schema.


linwen:
+1

PR #1139, opened 2017-02-23
HAWQ-1357. Super user also need to check create privilege of public schema from Ranger.

zhangh43:
HAWQ-1318 add create|usage privilege of public schema to superuser to fix hawq stop BUG caused by Resource Manager. 
But RM only need the usage privilege of public schema to query HAWQ, So create privilege need to be removed.


linwen:
+1

PR #1140, opened 2017-02-23
HAWQ-1358. Refactor gpfdist library in featuretest.

zhangh43:
extract gpfdist code(which exists in exttable test case) as a common library, to be used by both exttable and ranger feature test.
Main functions: 
gpfdist_init: to setup gpfdist environment and start server.
gpfdist_finalize: to stop gpfdist server.

### DIFF
--- a/src/test/feature/lib/gpfdist.cpp
+++ b/src/test/feature/lib/gpfdist.cpp
@@ -1,0 +1,71 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


linwen:
+1

src/test/feature/lib/gpfdist.cpp
```diff
@@ -0,0 +1,71 @@
...
+			"CREATE EXTERNAL WEB TABLE gpfdist_start (x text) "
+					"execute E'((%s/bin/gpfdist -p 7070 -d %s  </dev/null >/dev/null 2>&1 &); sleep 2; echo \"starting\"...) ' "
+					"on SEGMENT 0 "
+					"FORMAT 'text' (delimiter '|');";
+	std::string path = util->getTestRootPath() + "/ExternalSource/data";
+	util->execute(hawq::test::stringFormat(sql, GPHOME, path.c_str()));
+
+	util->execute(
+			"CREATE EXTERNAL WEB TABLE gpfdist_stop (x text) "
+					"execute E'(/bin/pkill gpfdist || killall gpfdist) > /dev/null 2>&1; echo stopping...' "
```
paul-guo-: Do not hardcode path for those utilities. e.g. On my linux box pkill is under /usr/bin

zhangh43: remove the hardcode path. using pkill instead.

PR #1141, opened 2017-02-23
HAWQ-1342. Fixed QE process hang in shared input scan on segment node

liming01:
The basic idea for this kinds of hung problem is to:
(1) The error thrown segment will invoke rollback the whole transaction, and all related fd will be closed during transaction end.
(2) The other segment just act as before, when wait for select(), it will loop until the specific fd is closed, then the code will run until process interrupt (the rollback transaction will send cancel signal) again in other place afterward.

So some previous fix (HAWQ-166,  HAWQ-1282) will be changed accordingly.
(1) HAWQ-166: we don't need to skip sending info
(2) HAWQ-1282:
  - we don't need to close the fd, it will be closed automatically during transaction end.
  - we just end loop if we find the related FD has already been closed.

Signed-off-by: Amy Bai <abai@pivotal.io>

### DIFF
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -40,7 +40,6 @@
 
 #include "postgres.h"
 
-#include "access/xact.h"
 #include "cdb/cdbvars.h"
 #include "executor/executor.h"
 #include "executor/nodeShareInputScan.h"
@@ -641,10 +640,6 @@ read_retry:
 		goto read_retry;
 	else
 	{
-		if(fd >= 0)
-		{
-			gp_retry_close(fd);
-		}
 		elog(ERROR, "could not read from fifo: %m");
 	}
 	Assert(!"Never be here");
@@ -664,10 +659,6 @@ write_retry:
 		goto write_retry;
 	else
 	{
-		if(fd >= 0)
-		{
-			gp_retry_close(fd);
-		}
 		elog(ERROR, "could not write to fifo: %m");
 	}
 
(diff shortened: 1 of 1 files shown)


src/backend/executor/nodeShareInputScan.c
```diff
@@ -925,9 +923,12 @@ writer_wait_for_acks(ShareInput_Lk_Context *pctxt, int share_id, int xslice)
...
 					share_id, currentSliceId, save_errno);
-			/*if error(except EINTR) happens in select, we just return to avoid endless loop*/
-			if(errno != EINTR){
-				return;
+			if(save_errno == EBADF)
+			{
+				/* The file description is invalid, maybe this FD has been already closed by writer in some cases
+				 * we need to break here to avoid endless loop and continue to run CHECK_FOR_INTERRUPTS.
+				 */
+				break;
```
paul-guo-: The code change looks fine. Just I think for errno other than EINTR, we should log(ERROR, ....)

Same comment for similar changes.

ictmalili: Agree! Can we log out the information?

liming01: select() ERRORS :
EBADF  -- break;
EINTR   -- loop
EINVAL -- programming error, should not occurs

on Linux:
ENOMEM -- loop, wait for runaway to choose one transaction to rollback, or OS choose one process to kill

On macos:
EAGAIN -- loop

EAGAIN -- loop

So we just process the EBADF only, others are loop again or impossible to occurs. Thanks.

ictmalili: Thanks @liming01.  Good explanation. 

Do we need log the error information out before "break".  For example, adding log to indicate that the fd is no longer valid.

zhangh43:
+1

src/backend/executor/nodeShareInputScan.c
paul-guo-: Normally EAGAIN should not happen (I did not see this on Linux manage however saw this on mac manage), but if it happens yes we should continue just as EINTR.

For ENOMEM: It is debatable to quit myself or just keep trying and then risk letting os kill one or more processes. For me I'd quit the query to risk the kill of any process. For this err code, continue or error out, up to you.

EINVAL and even possible other errno code ( that is a bug in either manage or kernel), we should not trust or that depend on os because the cost could be high (hang), given we do not have retry time limit here (By the way we should really have such mechanism for all hang-possible while) .

liming01: @ictmalili ,  already printed log there, so need to to add code.
@paul-guo- I just process the error code as doc described, if there are some bug is OS, then we should check log when the hung problem occur again. At that time, we can find out the root cause. We just need to make sure not to hung by hawq code bug here. Thanks.

ictmalili:
LGTM. +1

PR #1143, opened 2017-02-23
HAWQ-1350. Add --enable-rps option to build ranger-plugin when build …

stanlyxiang:
…hawq.

please help to review, thanks. @ictmalili @linwen @zhangh43 @interma @wcl14


paul-guo-:
I have a question: Is this for Ranger support only? If yes, should not enable-ranger be enough or better?

By the way, Shouldn't contrib/ranger-plugin be a better place for ranger-plugin?

stanlyxiang:
Yes. This is for ranger support only. In our work, we avoid to confuse users with ranger and ranger-plugin. Because users should install ranger (apache ranger) before use HAWQ integration with Ranger. We should avoid the misleading that ranger has been installed&enabled if users configure hawq with option "enable-ranger".

For the plugin folder, agree with you as Postgresql says "contrib directory include porting tools, analysis utilities, and plug-in features that are not part of the core PostgreSQL system".  This part was committed by @denalex , we could discuss and file a jira to move if we reach a consensus.

config/tomcat.m4
```diff
@@ -0,0 +1,59 @@
...
+AC_DEFUN([PGAC_CATALINA_HOME],
+[
+  AC_MSG_CHECKING([CATALINA_HOME])
+
+  dnl Use $CATALINA_HOME if specifed, otherwise look for java in the likely places.
+  dnl /usr/local/Cellar/tomcat@6/6.0.45/libexec 
+  dnl /usr/lib/bigtop-tomcat
+  if test -x "${CATALINA_HOME}/bin/catalina.sh"; then
+    TOMCAT="${CATALINA_HOME}"
+  elif test -x "/usr/local/Cellar/tomcat@6/6.0.45/libexec/bin/catalina.sh"; then
```
ictmalili: Are we bound to this tomcat version: 6.0.45?

stanlyxiang: Oh. No, the version should be 6.0.x. I will remove this else option since I just found it's too specific to the path.

linwen:
+1

stanlyxiang:
Merged into master. Close this PR.

PR #1144, opened 2017-02-23
HAWQ-1360. Check privilege of sequence pass the wrong type to RPS.

zhangh43:
When check privilege of query with ranger enabled, a json request will be send to RPS, including information about: privilege type, resource type and role name. 
Currently, checking sequence privilege will send resource type "table" instead of  "sequence".


linwen:
+1

PR #1151, opened 2017-02-28
HAWQ-1359. Add policy test for HAWQ with Ranger enabled.

zhangh43:
Policy test includes different json requests(used by ranger) for different quries, For example, for query "select * from a;", it needs usage privilege of schema public, and select privilege of table a.
There are also queries can only be executed by superuser, we also test them in policy test.

Signed-off-by: Chuling Wang <wangchunling14@126.com>


paul-guo-:
Maybe a.txt could  be a more meaningful file name?
Is __init__.py useful? (If yes, please add license header).

2017-02-28 11:22 GMT+08:00 Huan Zhang <notifications@github.com>:

paul-guo-:
The shell scripts: Please add license header.
The tar.gz file: I'd suggest generating it dynamically.

Please make sure license check passes.

2017-02-28 11:47 GMT+08:00 Paul Guo <paulguo@gmail.com>:

zhangh43:
Using rat to add licenses and remove some redundant file.

zhangh43:
This PR use different users to run different queries, which is a consideration of performance. Since adding a policy in ranger will lead to several secs delay, and we use sleep() to ensure policy is ready. If we add policy for each query, the total sleep time is linear to the case number. So we choose to use run all the queries by different users and add policies firstly and then sleep only once, finally we run all the queries again with added policies to check whether policies work.

src/test/feature/Ranger/rangeruser.py
```diff
@@ -0,0 +1,117 @@
...
+
+def foo_callback(option, opt, value, parser):
+  setattr(parser.values, option.dest, value.split(','))
+
+def option_parser():
+    '''option parser'''
+    parser = OptionParser()
+    parser.remove_option('-h')
+    parser.add_option('-?', '--help', action='help')
+    parser.add_option('-h', '--host', dest="host", help='host of the target DB', \
```
linwen: this option specifies the host of ranger server not "host of the target DB"

src/test/feature/Ranger/rangeruser.py
```diff
@@ -0,0 +1,117 @@
...
+                      default='localhost')
+    parser.add_option('-p', '--port', dest="port", \
```
linwen: should be the port of Ranger Policy Manager

linwen:
+1

PR #1152, opened 2017-02-28
Hawq 1365. Print out detailed schema information for tables which the user doesn't have privileges

interma:
1. Print out detailed schema information for tables which the user doesn't have privileges.
2. Fixed duplicate tmp_folder name of TestUsage1FolderFailure of register test.
3. Refine the _last_query_sign_ logic of HAWQ-1355.

Detailed output information after modified is here:
https://issues.apache.org/jira/browse/HAWQ-1365

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -1982,6 +1982,13 @@ recomputeNamespacePath(void)
 			elog(DEBUG3, "recompute search_path[%s] when acl_type is ranger", namespace_search_path);
 		}
 	}
+	else 
+	{
+		if (debug_query_string != NULL)
+		{
+			last_query_sign = string_hash(debug_query_string, strlen(debug_query_string));
+		}
+	}
 
 	/* Need a modifiable copy of namespace_search_path string */
 	rawname = pstrdup(namespace_search_path);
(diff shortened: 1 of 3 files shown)


interma:
@linwen @stanlyxiang @wcl14 pls to review, thx!

src/backend/catalog/namespace.c
```diff
@@ -1982,6 +1982,13 @@ recomputeNamespacePath(void)
 			elog(DEBUG3, "recompute search_path[%s] when acl_type is ranger", namespace_search_path);
 		}
 	}
+	else 
+	{
+		if (aclType == HAWQ_ACL_RANGER && debug_query_string != NULL)
+		{
+			last_query_sign = string_hash(debug_query_string, strlen(debug_query_string));
+		}
+	}
```
wcl14: How about set current_qurey_sign before the if statement 'if (namespaceSearchPathValid && namespaceUser == roleid)' and set last_query_sign after the if statement? Can this make the return condition more clear?

interma: Thx!
If there are other code should be changed also, I will try to clear the logic.

interma:
commited.

PR #1156, opened 2017-03-01
HAWQ-1370. Misuse of regular expressions in init_file of feature test.

zhangh43:
in global_init_file of feature test, we want to skip expressions which include file and line number, e.g.(aclchk.c:123), or (aclchk.cpp:134).
But currently, the regular expressions is `(.*c[p]+:\d+) `which need to be replaced by `(.*c[p]*:\d+) `


linwen:
+1

paul-guo-:
+1

PR #1162, opened 2017-03-06
HAWQ-1378. Elaborate the "invalid command-line arguments for server p…

paul-guo-:
…rocess" error.

This patch mostly follows the pg commit below.

commit 86947e666d39229558311d7b0be45608fd071ed8
Author: Peter Eisentraut <peter_e@gmx.net>
Date:   Sun Mar 11 01:52:05 2012 +0200

    Add more detail to error message for invalid arguments for server process

    It now prints the argument that was at fault.

    Also fix a small misbehavior where the error message issued by
    getopt() would complain about a program named "--single", because
    that's what argv[0] is in the server process.


huor:
LGTM +1

paul-guo-:
Yes, and then print the immediate errorrous argument.

paul-guo-:
Merged. Closing.

PR #1164, opened 2017-03-07
HAWQ-1362. Add RPS start/stop script in HAWQ start/stop script

stanlyxiang:
please review, thanks.

### DIFF
--- a/ranger-plugin/conf/rps.properties
+++ b/ranger-plugin/conf/rps.properties
@@ -30,3 +30,6 @@ RANGER_HAWQ_INSTANCE=hawq
 
 # version of the software
 RPS_VERSION=${project.version}
+
+# home path of Tomcat/6.0.x for Ranger Plugin Service
+CATALINA_HOME=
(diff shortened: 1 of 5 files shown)


ranger-plugin/conf/rps.properties
```diff
@@ -29,4 +29,7 @@ RPS_SHUTDOWN_PORT=8405
 RANGER_HAWQ_INSTANCE=hawq
 
 # version of the software
-RPS_VERSION=${project.version}
\ No newline at end of file
+RPS_VERSION=${project.version}
+
+# home path of Tomcat/6.0.x for Ranger Plugin Service
+CATALINA_HOME=
```
denalex: should this line not be here by default and only be added by users configuring the system from the build script ? Users that install from RPM do not need this property and might be confused why it is empty in this file.

stanlyxiang: We could document this for community users and other users that install from RPM. 
But it's okay to make a change. 
I will move the definition of CATALINA_HOME to rps.properties and leave a placeholder for community users to add their own path.

linwen:
+1

stanlyxiang:
Merged into master. Close this PR.

PR #1168, opened 2017-03-09
HAWQ-1354. Externalized RPS heap size parameter

denalex:


ranger-plugin/conf/rps.properties
```diff
@@ -31,6 +31,9 @@ RANGER_HAWQ_INSTANCE=hawq
 # version of the software
 RPS_VERSION=${project.version}
 
+# max heap size for RPS (Tomcat) process
+RPS_HEAP_SIZE=128m
```
sansanichfb: Is it enough for RPS?

denalex: did load test, with 10 concurrent requests the mean response time is 4ms, 50 concurrent requests -- 18ms. min memory is about 40-50MB and GC can handle collections under load with about 100ms / GC So, this seems to be sufficient for anticipated workloads and out-of-the box sizing, it can be increased by customers, if needed.

PR #1170, opened 2017-03-10
HAWQ-1371. Fix QE process hang in shared input scan.

amyrazz44:
Due to the complication of this bug. We split the fix to two parts. First of all, fix reader/writer process waits for writer/read's acks while writer/reader process may abort at the same time. Secondly, fix select endless loop.  The other fix will be committed in the other pull request.

### DIFF
--- a/src/backend/executor/nodeShareInputScan.c
+++ b/src/backend/executor/nodeShareInputScan.c
@@ -40,7 +40,6 @@
 
 #include "postgres.h"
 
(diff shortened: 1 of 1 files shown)


amyrazz44:
@huor @liming01 , please feel free to review, thank you.

liming01:
+1

huor:
+1 for the fix

paul-guo-:
+1

PR #1171, opened 2017-03-14
HAWQ-1359. Add test cases for Ranger support, combinations of differe…

linwen:
Please review, thanks!


paul-guo-:
sql_util.cpp: I think those are debugging code, right?
Do it pass the rat license check?

linwen:
This file already has license header.

paul-guo-:
Sorry. For license check, I mean those new added files.

linwen:
It has passed jenkins and travis check, I think it can pass license check.

zhangh43:
+1

ictmalili:
+1

src/test/feature/lib/sql_util.cpp
```diff
@@ -224,7 +225,9 @@ const string SQLUtility::generateSQLFile(const string &sqlFile, bool usingDefaul
     EXPECT_TRUE(false) << "Error opening file " << newSqlFile;
   }
   out << "-- start_ignore" << std::endl;
+  printf("dd2d%s\n",schemaName.c_str());
   if (!usingDefaultSchema) {
+	  printf("ddd%s\n",schemaName.c_str());
```
stanlyxiang: what does "dd2d" and "ddd" mean ?

src/test/feature/Ranger/test_ranger.cpp
```diff
@@ -24,19 +24,34 @@
...
 using hawq::test::SQLUtility;
 using hawq::test::Command;
 
+TestHawqRanger::TestHawqRanger()
+{
+	initfile = hawq::test::stringFormat("Ranger/sql/init_file");
+	rangerHost = getRangerHost();
+}
+
+std::string& TestHawqRanger::getRangerHost()
```
stanlyxiang: Better to get guc value by connecting to db and using sql from pg_settings, not hawq config. Using sql could avoid string find and hardcoded number length. And hawq config also connect to db and use sql to get the value.

stanlyxiang: And there is function "getGUCValue" to get the "hawq_rps_address_host".
Besides, "hawq_rps_address_host" is the rps address host, not ranger host, so what do you wanna to get ? If ranger server host, then they are not the same.

ranger-plugin/.gitignore
```diff
@@ -1,2 +1,3 @@
 target/
 Makefile.global
+/bin/
```
stanlyxiang: is there any dir "bin" generated by building ? "/bin/" only ignore directory "bin" in current directory, e.g ranger-plugin/bin. Recommend "bin/" to ignore all "bin" directory from current directory to all sub-directories.

src/test/feature/Ranger/test_ranger.cpp
stanlyxiang: I checked rangeruser.py and found that the "-h" option stands for "ranger server host". Since rps won't be installed on ranger server host that usually, so I don't think this is a right way to get ranger server host.

src/test/feature/lib/sql_util.cpp
linwen: Thanks, I will check it.

src/test/feature/Ranger/test_ranger.cpp
linwen: The RPS and HAWQ master are required to installed on a same node. This test case is to add/remove policy from Ranger Admin, but HAWQ only knows and interacts with RPS, it need not configure Ranger Admin address/hostname. 

I am considering still use a environment value to control it.

stanlyxiang: I mean the python script "rangeruser.py" need ranger server host, but you gave it rps host.

PR #1173, opened 2017-03-15
HAWQ-1386. Mask some generated files for pljava feature testing for git.

paul-guo-:


huor:
Good to have the improvement. +1

PR #1176, opened 2017-03-16
HAWQ-1389. Fixed test case problem

liming01:
This fix is not valid any more, wait for new fix please.

I had thought that the root cause is the testing relation name conflict with other test case.
However the feature test will automatically add schema name in form of "class_func". So it should be other cause, need more time for investigating.


paul-guo-:
Looks good. Just one thing, the commit title should align with the Apache JIRA title, which is better IMO.

liming01:
This PR is closed for no valid any more.

PR #1178, opened 2017-03-20
HAWQ-1393. 'hawq stop cluster' failed when rps.sh have some path erro…

interma:
…rs (e.g. CATALINA_HOME)

Ignore the `rps.sh stop` result when **try** to stop RPS. 
Only check result on definitely known RPS is running.

Example:
When master is down and CATALINA_HOME is invalid, `hawq stop cluster -a` calls `rps.sh stop`, it only prompts error message, and continue to stop segments.


interma:
@ictmalili @linwen @stanlyxiang help to review, thanks!

PR #1179, opened 2017-03-21
HAWQ-1395. PXF build script to support passing additional Gradle options.

sansanichfb:


pxf/Makefile
```diff
@@ -42,6 +42,10 @@ ifneq "$(PXF_VERSION)" ""
     BUILD_PARAMS+= -Pversion="$(PXF_VERSION)"
 endif
 
+ifneq "$(GRADLEW_OPTIONS)" ""
```
shivzone: The purpose of Makefile is to abstract the user from gradle options/commands. Should we instead call this BUILD_OPTIONS ?

sansanichfb: Another option - add EXCLUDE_PROJECTS, which will take comma-separated list of projects.

shivzone: Yes, that would be good

sansanichfb:
This specific change just excludes buildRpm task, all other tasks still remain.

PR #1183, opened 2017-03-22
HAWQ-1400. An improvement in SQL Util destructor function for feature test

jiny2:


jiny2:
This is an external improvement from test tool side to make more stable CI testing.

jiny2:
@huor @paul-guo- @stanlyxiang Please review, thanks.

paul-guo-:
Thanks. I think at least we should add comment here at first.

jiny2:
comment is added to explain the sleep

huor:
+1 for the fix. The clarification looks better to have root cause

PR #1184, opened 2017-03-22
HAWQ-1401. Fix HAWQ make install with DESTDIR.

radarwave:
Need to handle pygresql install as it depend on the pre-installed hawq components.

### DIFF
--- a/tools/bin/pythonSrc/PyGreSQL-4.0/setup.py
+++ b/tools/bin/pythonSrc/PyGreSQL-4.0/setup.py
@@ -62,6 +62,10 @@ def pg_config(s):
         raise Exception("pg_config tool is not available.")
     if not d:
         raise Exception("Could not get %s information." % s)
+
+    DESTDIR = os.getenv('DESTDIR')
+    if DESTDIR:
+        d = DESTDIR + d
     return d
 
 def mk_include():
--- a/src/backend/Makefile
+++ b/src/backend/Makefile
@@ -305,13 +305,13 @@ endif
 	$(INSTALL_DATA) $(srcdir)/libpq/pg_ident.conf.sample '$(DESTDIR)$(datadir)/pg_ident.conf.sample'
 	$(INSTALL_DATA) $(srcdir)/utils/misc/postgresql.conf.sample '$(DESTDIR)$(datadir)/postgresql.conf.sample'
 	$(INSTALL_DATA) $(srcdir)/access/transam/recovery.conf.sample '$(DESTDIR)$(datadir)/recovery.conf.sample'
-	${INSTALL_SCRIPT} -d ${sysconfdir}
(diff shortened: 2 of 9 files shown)


radarwave:
@edespino @paul-guo- @huor @xunzhang 
Please help to review, thanks.

tools/bin/pythonSrc/PyGreSQL-4.0/setup.py
```diff
@@ -62,6 +62,11 @@ def pg_config(s):
         raise Exception("pg_config tool is not available.")
     if not d:
         raise Exception("Could not get %s information." % s)
+
+    DESTDIR = os.getenv('DESTDIR')
+    if DESTDIR:
+        print "DESTDIR is %s" % DESTDIR
```
huor: Need to remove the debug info here.

huor:
Great to have the fix for build. +1

tools/bin/pythonSrc/PyGreSQL-4.0/setup.py
radarwave: Thank, fixed.

paul-guo-:
Cool work! I just have one concern.
 
I saw a lot of fix like this, however is is possible the later one is a relative path while DESTDIR has not a tailing / and this could lead to a bug, so maybe you could add / between them to prevent this, besides: 1) This will make code easier to read 2) even there is // in your path, it is not a problem also. How do you think?

$(DESTDIR)${sysconfdir}

radarwave:
@paul-guo- 
I don't think the --prefix option can set a relative path, so there is definitely no need to add a '/' between DESTDIR and PREFIX-DIR.

paul-guo-:
If you can make sure all of the later paths are absolute path, it will be fine, e.g.
${DESTDIR}${docdir}
${DESTDIR}$(bindir)
${DESTDIR}${libdir}
$(DESTDIR)${sysconfdir}

I'm just a bit worried about the could be a bug in the future.

radarwave:
They are all generated by 'pg_config', they should not be relative PATH. 


# Extension makefiles should set PG_CONFIG, but older ones might not
ifndef PG_CONFIG
PG_CONFIG = pg_config
endif

bindir := $(shell $(PG_CONFIG) --bindir)
datadir := $(shell $(PG_CONFIG) --sharedir)
sysconfdir := $(shell $(PG_CONFIG) --sysconfdir)
libdir := $(shell $(PG_CONFIG) --libdir)
pkglibdir := $(shell $(PG_CONFIG) --pkglibdir)
includedir := $(shell $(PG_CONFIG) --includedir)
pkgincludedir := $(shell $(PG_CONFIG) --pkgincludedir)
mandir := $(shell $(PG_CONFIG) --mandir)
docdir := $(shell $(PG_CONFIG) --docdir)
localedir := $(shell $(PG_CONFIG) --localedir)

paul-guo-:
OK, so +1.

PR #1185, opened 2017-03-22
HAWQ-1402. Updated comments for Hive-related profiles.

sansanichfb:
@lisakowen

### DIFF
--- a/pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
+++ b/pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
@@ -46,7 +46,7 @@ under the License.
         <description>
             This profile is suitable for using when connecting to Hive.
             Supports GPDBWritable output format, as specified in FORMAT header parameter.
-            Based on storage format, it uses one of optimized profiles(HiveRC, HiveText, HiveORC) whenever it's possible.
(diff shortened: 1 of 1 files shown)


pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -46,6 +46,7 @@ under the License.
         <description>
             This profile is suitable for using when connecting to Hive.
             Supports GPDBWritable output format, as specified in FORMAT header parameter.
+            Based on storage format, it uses one of optimized profiles(HiveRC, HiveText, HiveORC) whenever it's possible.
```
denalex: I'd say: "It auto-detects actual file storage format and uses an optimized profile (HiveRC, HiveText, HiveORC) if applicable."

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -46,6 +46,7 @@ under the License.
...
         </description>
```
denalex: Does it also need to be changed in Ambari since there is a corresponding template there ?

sansanichfb: Yes, will update it in Ambari as well.

PR #1186, opened 2017-03-24
HAWQ-1405. 'hawq stop --reload' should not stop RPS

interma:


interma:
@ictmalili @linwen @stanlyxiang help to review, thanks!

tools/bin/hawq_ctl
```diff
@@ -1041,14 +1041,14 @@ class HawqStop:
             logger.info("Segments %s successfully" % self.stop_action_past)
         return total_return_flag
 
-    def _stop_rps(self):
-        cmd_str = "%s/ranger/bin/rps.sh stop" % (self.GPHOME)
-        result = remote_ssh(cmd_str, self.master_host_name, self.user)
-        return result
 
-    def stop_rps(self, check_ret = True):
```
stanlyxiang: You can simply add "if reload" in this function and no need to remove _stop_rps && update function call. 

Anyway, this change is also acceptable.

interma: I saw other functions ( e.g. _stop_segment(), _stop_master() ）) in HawqStop class all have '_' prefix, so merge stop_rps() logic to _stop_rps() for unifying them. 
Sounds reasonable?

interma:
merged

PR #1187, opened 2017-03-24
HAWQ-1408. Fixed crash when alloc not enough segno for write

liming01:
In case of there is no reproduce steps, I only went through the code to find out possible root cause. Now I get one possible cause.  The possible reason is related https://issues.apache.org/jira/browse/HAWQ-642.

This fix includes below changes:
(1) When keepHash, we can not guarantee that only generate remaining_num is enough to alloc seg file for all segment_num. Because the next new generated seg file number may have same hash key id with the old one.
(2) When !keepHash, now the remaining_num returned from addCandidateSegno() is not precise. So we need to fix it to meet the need of HAWQ-642.
(3) At the final call to addCandidateSegno(), we should keep monitoring the remaining_num instead of at the beginning because of the reason (1). So that even the new seg file is not actually used by this query ( maybe for hash key conflict), we can continue to alloc enough seg files for this query.

@ictmalili @zhangh43, please review it. Thanks.

### DIFF
--- a/src/backend/access/appendonly/appendonlywriter.c
+++ b/src/backend/access/appendonly/appendonlywriter.c
@@ -403,7 +403,7 @@ AORelCreateHashEntry(Oid relid)
 	aoHashEntry->head_rel_segfile.xid = InvalidTransactionId;
 	aoHashEntry->head_rel_segfile.latestWriteXid = InvalidTransactionId;
 	aoHashEntry->head_rel_segfile.isfull = true;
-  aoHashEntry->head_rel_segfile.needCheck = true;
+	aoHashEntry->head_rel_segfile.needCheck = true;
 	aoHashEntry->head_rel_segfile.tupcount = 0;
 	aoHashEntry->head_rel_segfile.tupsadded = 0;
 	aoHashEntry->max_seg_no = 0;
@@ -923,9 +923,9 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
  */
 int
 addCandidateSegno(AOSegfileStatus **maxSegno4Seg, int segment_num, AOSegfileStatus *candidate_status, bool isKeepHash)
-{  
+{
+	int remaining_num = segment_num;
 	if(isKeepHash){
-		int remaining_num  = segment_num;
 		int idx = (candidate_status->segno+segment_num-1) % segment_num;	// from 1 to segment_num-1, then 0(segment_num)
 		if (NULL==maxSegno4Seg[idx] || maxSegno4Seg[idx]->segno > candidate_status->segno)  //using the min seg no firstly.
         	maxSegno4Seg[idx] = candidate_status;
@@ -933,20 +933,24 @@ addCandidateSegno(AOSegfileStatus **maxSegno4Seg, int segment_num, AOSegfileStat
 		for(int i=0; i<segment_num; i++)
     	{
     		if(NULL!=maxSegno4Seg[i] && maxSegno4Seg[i]->segno > 0){
-            remaining_num--;
+				remaining_num--;
           	}
     	}
-    	return remaining_num;
 	}else{
+		bool assigned = false; // candidate_status assigned?
+		remaining_num  = 0;
 		for(int i=0; i<segment_num; i++){
-			if (NULL==maxSegno4Seg[i]){
-				maxSegno4Seg[i]= candidate_status;
-				return segment_num-i-1;	
+			if (NULL==maxSegno4Seg[i] || maxSegno4Seg[i]->segno <= 0){
+				if(!assigned){
+					maxSegno4Seg[i]= candidate_status;
+					assigned = true;
+				}else{
+					remaining_num++;
+				}
 			}
 		}
 	}
-   Assert(false);
-   return 0;
+	return remaining_num;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -923,30 +923,34 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
...
 		for(int i=0; i<segment_num; i++)
     	{
     		if(NULL!=maxSegno4Seg[i] && maxSegno4Seg[i]->segno > 0){
-            remaining_num--;
+				remaining_num--;
           	}
     	}
-    	return remaining_num;
 	}else{
+		int assigned = false; // candidate_status assigned?
```
hsyuan: Why not bool?

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -923,30 +923,34 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
  */
 int
 addCandidateSegno(AOSegfileStatus **maxSegno4Seg, int segment_num, AOSegfileStatus *candidate_status, bool isKeepHash)
-{  
+{
+	int remaining_num  = segment_num;
```
hsyuan: The space of this chunk of code is messy.
`{}` placement is not consistent, even in a single function.

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -1136,26 +1140,22 @@ List *SetSegnoForWrite(List *existing_segnos, Oid relid, int segment_num,
...
-                Assert(remaining_num==0);//make sure all segno got a candidate
-            }
+			while(remaining_num>0)
+			{
+				//generate new segment_num to make sure that in keepHash mode, all segment node has at least one segfile is writable
+				int new_status = AORelGetSegfileStatus(aoentry);
+				if (new_status == NEXT_END_OF_LIST)
+				{
+					LWLockRelease(AOSegFileLock);
+					ereport(ERROR, (errmsg("cannot open more than %d append-only table segment files cocurrently",
```
hsyuan: `cocurrently` typo

liming01: Thanks @hsyuan! I will correct all your comments.

src/backend/access/appendonly/appendonlywriter.c
liming01: Sorry for my mistake.

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -923,30 +923,34 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
  */
 int
 addCandidateSegno(AOSegfileStatus **maxSegno4Seg, int segment_num, AOSegfileStatus *candidate_status, bool isKeepHash)
```
hsyuan: The indentation of this function is horrible.

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -923,30 +923,34 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
...
     		if(NULL!=maxSegno4Seg[i] && maxSegno4Seg[i]->segno > 0){
-            remaining_num--;
+				remaining_num--;
           	}
     	}
-    	return remaining_num;
 	}else{
+		bool assigned = false; // candidate_status assigned?
+		remaining_num  = 0;
 		for(int i=0; i<segment_num; i++){
```
hsyuan:
```c
for(int i=0; i<segment_num; i++)
``` 
To
```c
for (int i = 0; i < segment_num; i++) 
````

File: src/backend/access/appendonly/appendonlywriter.c [hunk 6]
```diff
@@ -923,30 +923,34 @@ usedByConcurrentTransaction(AOSegfileStatus *segfilestat)
...
 		
 		for(int i=0; i<segment_num; i++)
     	{
     		if(NULL!=maxSegno4Seg[i] && maxSegno4Seg[i]->segno > 0){
-            remaining_num--;
+				remaining_num--;
           	}
     	}
-    	return remaining_num;
 	}else{
```
hsyuan:
```c
}else{
```
To
```c
} else {
```
Or
```c
}
else
{
```

Can you check for this whole function?

src/backend/access/appendonly/appendonlywriter.c
```diff
@@ -1136,26 +1140,22 @@ List *SetSegnoForWrite(List *existing_segnos, Oid relid, int segment_num,
...
-                                        MaxAORelSegFileStatus)));
-                    }
-                    AOSegfileStatusPool[new_status].segno = ++aoentry->max_seg_no;
-                    AOSegfileStatusPool[new_status].next = aoentry->head_rel_segfile.next;
-                    aoentry->head_rel_segfile.next = new_status;
-                    remaining_num = addCandidateSegno(maxSegno4Segment, segment_num,  &AOSegfileStatusPool[new_status], keepHash);
-                }
-                Assert(remaining_num==0);//make sure all segno got a candidate
-            }
+			while(remaining_num>0)
```
hsyuan: and here

src/backend/access/appendonly/appendonlywriter.c
liming01: Please append '?ts=4' to web URL when code review.

hsyuan: Can you give me an example? I don't get it.

liming01: Because github default ts=8, which may a lot indentation looks strange. 
However our code convention is to set ts=4.
e.g. When you review code, please append '?ts=4' to web URL as below:
https://github.com/apache/incubator-hawq/pull/1187/files?ts=4

wengyanqing:
+1

liming01:
Merged into code master branch.

PR #1189, opened 2017-03-24
HAWQ-1409. Send AGG-TYPE header to PXF

kavinderd:
This change is mean to be a proof of concept that pushing down
aggregate function information from HAWQ to the underlying external
storage layer does indeed improve performance

### DIFF
--- a/src/backend/access/external/fileam.c
+++ b/src/backend/access/external/fileam.c
@@ -459,12 +459,20 @@ external_stopscan(FileScanDesc scan)
  *	----------------
  */
 ExternalSelectDesc
-external_getnext_init(PlanState *state) {
+external_getnext_init(PlanState *state, ExternalScanState *es_state) {
 	ExternalSelectDesc desc = (ExternalSelectDesc) palloc0(sizeof(ExternalSelectDescData));
+	Plan *rootPlan;
 
 	if (state != NULL)
 	{
 		desc->projInfo = state->ps_ProjInfo;
+		/*
+		 * If we have an agg type then our parent is an Agg node
+		 */
+		rootPlan = state->state->es_plannedstmt->planTree;
+		if (rootPlan->type == T_Agg && es_state->parent_agg_type) {
+			desc->agg_type = es_state->parent_agg_type;
+		}
 	}
 	return desc;
 }
--- a/src/backend/access/external/fileam.c
+++ b/src/backend/access/external/fileam.c
@@ -470,7 +470,7 @@ external_getnext_init(PlanState *state, ExternalScanState *es_state) {
 		 * If we have an agg type then our parent is an Agg node
 		 */
 		rootPlan = state->state->es_plannedstmt->planTree;
-		if (rootPlan->type == T_Agg && es_state->parent_agg_type) {
+		if (IsA(rootPlan, Agg) && es_state->parent_agg_type) {
 			desc->agg_type = es_state->parent_agg_type;
 		}
 	}
--- a/src/backend/access/external/pxfheaders.c
+++ b/src/backend/access/external/pxfheaders.c
@@ -110,6 +110,20 @@ void build_http_header(PxfInputData *input)
 	else
 		churl_headers_append(headers, "X-GP-HAS-FILTER", "0");
 
+	/* Aggregate information */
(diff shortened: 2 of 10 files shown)


src/backend/access/external/pxfheaders.c
```diff
@@ -110,6 +110,20 @@ void build_http_header(PxfInputData *input)
 	else
 		churl_headers_append(headers, "X-GP-HAS-FILTER", "0");
 
+	/* Aggregate information */
+	if (input->agg_type) {
+		char agg_groups_str[sizeof(int32)];
```
hsyuan: Where is this var used?

src/backend/executor/nodeAgg.c
```diff
@@ -1950,6 +1950,21 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
...
+	if (outerPlan->type == T_ExternalScan) {
+		/*
+		 * Hack to indicate to PXF when there is an external scan
+		 */
+		if (aggstate->aggs && list_length(aggstate->aggs) == 1) {
+			foreach(l, aggstate->aggs) {
+				AggrefExprState *aggrefstate = (AggrefExprState *) lfirst(l);
+				Aggref	   *aggref = (Aggref *) aggrefstate->xprstate.expr;
+				//Only dealing with one agg
+				if (aggref->aggfnoid == 2147 || aggref->aggfnoid == 2803) {
```
hsyuan: Would be better use `COUNT_ANY_OID` and `COUNT_STAR_OID`...

src/backend/access/external/fileam.c
```diff
@@ -459,12 +459,20 @@ external_stopscan(FileScanDesc scan)
...
+	Plan *rootPlan;
 
 	if (state != NULL)
 	{
 		desc->projInfo = state->ps_ProjInfo;
+		/*
+		 * If we have an agg type then our parent is an Agg node
+		 */
+		rootPlan = state->state->es_plannedstmt->planTree;
+		if (rootPlan->type == T_Agg && es_state->parent_agg_type) {
```
hsyuan: Recommend to use `IsA(rootPlan, Agg)`

src/backend/executor/nodeAgg.c
```diff
@@ -1950,6 +1950,21 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
 	 * initialize child nodes
 	 */
 	outerPlan = outerPlan(node);
+	if (outerPlan->type == T_ExternalScan) {
```
hsyuan: Better use `IsA(outerPlan, ExternalScan)`

src/backend/executor/nodeAgg.c
```diff
@@ -1950,6 +1950,21 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
...
+		/*
+		 * Hack to indicate to PXF when there is an external scan
+		 */
+		if (aggstate->aggs && list_length(aggstate->aggs) == 1) {
```
hsyuan: Do we need check `aggstate->aggs` is null or not? if it is null, `list_length` will return 0.

src/backend/executor/nodeAgg.c
```diff
@@ -1950,6 +1950,21 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
...
+			foreach(l, aggstate->aggs) {
```
hsyuan: if the size of `aggstate->aggs` is already 1, do we need use `foreach` to iterate?

src/backend/executor/nodeAgg.c
```diff
@@ -1950,6 +1950,21 @@ ExecInitAgg(Agg *node, EState *estate, int eflags)
...
+				AggrefExprState *aggrefstate = (AggrefExprState *) lfirst(l);
```
hsyuan: just by changing `lfirst(l)` to `linitial(aggstate->aggs)`, you can remove `foreach` loop.

src/backend/executor/nodeAgg.c
kavinderd: Oh ok, I thought `list_length()` on a null value would cause an error

PR #1195, opened 2017-03-27
HAWQ-326. Support RPM package for Apache HAWQ.

radarwave:
1. Add a RPM spec file to support building HAWQ rpm package from source code tarball.
2. Add 'make rpm' command to build HAWQ/PXF/Ranger-plugin rpm packages and make into a binary tarball from source code tarball.

### DIFF
--- a/contrib/hawq-package/build_hawq_rpm.sh
+++ b/contrib/hawq-package/build_hawq_rpm.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+CUR_DIR=$(pwd)
+
+if [ -f /etc/redhat-release ] ; then
+    DISTRO_VERSION=$(cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*//)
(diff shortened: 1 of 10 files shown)


radarwave:
@shivzone @denalex @huor @paul-guo- @edespino 

Please help to review, including PXF/Ranger-plugin changes.

huor:
+1 for making the rpm build available for hawq!

contrib/hawq-package/build_hawq_rpm.sh
```diff
@@ -0,0 +1,77 @@
...
+    DISTRO_VERSION=$(cat /etc/redhat-release | sed s/.*release\ // | sed s/\ .*//)
+    DISTRO_MAJOR_VERSION=$(echo ${DISTRO_VERSION} | awk -F '.' '{print $1}')
+    OS_TYPE="el${DISTRO_MAJOR_VERSION}"
+elif [ -f /etc/SuSE-release ] ; then
+    DISTRO_VERSION=$(cat /etc/SuSE-release | tr "\n" ' ' | sed s/.*=\ //)
+    DISTRO_MAJOR_VERSION=$(echo ${DISTRO_VERSION} |awk -F '.' '{print $1}')
+    OS_TYPE="suse${DISTRO_MAJOR_VERSION}"
+fi
+
+HAWQ_RELEASE_VERSION=$(cat ../../getversion| grep ^GP_VERSION | cut -d '=' -f2 | sed 's|"||g' | cut -d '-' -f1)
```
paul-guo-: I think we should allow HAWQ_RELEASE_VERSION (and document it) to be predefined (if not predefined then fallback to call getversion) so that users do not need to build a rpm whose version is different than the codebase.

paul-guo-: i.e. I expect the rpm build scripts and spec could be quite independent.

contrib/hawq-package/build_hawq_rpm.sh
```diff
@@ -0,0 +1,77 @@
...
+    exit 1
+fi
+
+pushd rpmbuild > /dev/null
+RPM_TOP_DIR=$(pwd)
+
+echo 'rpmbuild --define "_topdir ${RPM_TOP_DIR}" \
+               --define "_hawq_version ${HAWQ_RELEASE_VERSION}" \
+               --define "_rpm_os_version ${OS_TYPE}" \
+               -bb SPECS/hawq.spec'
```
paul-guo-: You do not need to do like this. You could use the switch: "set -x" and "set +x"

radarwave:
Made some changes base on @paul-guo- 's comments.

paul-guo-:
+1

PR #1198, opened 2017-03-28
HAWQ-1417. Fixed crash when ANALYZE after COPY because of invalid res…

liming01:
…ource owner

The root cause is:  when ANALYZE is running, it will commit previous transaction, and begin one transaction for each relation to process analyze. However some code added before begining transactions, which means in this time, no transaction, and no resource owner neither.

The solution is to: create one resource owner for analyzeStmt(), and delete it after processing it.

### DIFF
--- a/src/backend/commands/analyze.c
+++ b/src/backend/commands/analyze.c
@@ -308,6 +308,7 @@ void analyzeStmt(VacuumStmt *stmt, List *relids, int preferred_seg_num)
 	ListCell				*le1 = NULL;
 	int 					successCount = 0, failCount = 0;
 	StringInfoData 			failNames;
+	ResourceOwner owner, oldOwner;
 
 	/**
 	 * Ensure that an ANALYZE is requested.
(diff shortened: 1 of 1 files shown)


wengyanqing:
+1

src/backend/commands/analyze.c
```diff
@@ -476,6 +477,11 @@ void analyzeStmt(VacuumStmt *stmt, List *relids, int preferred_seg_num)
 		MemoryContextSwitchTo(analyzeStatementContext);
 	}
 
+	// Create a resource owner to keep track of our resources even not in trasaction block
```
huor: Use /* */ to keep it consistent

huor:
The fix looks good. Is it possible to add unit test for it in src/test/unit/mock/backend/commands/analyze_mock.c?

liming01:
(1) no file in master branch: src/test/unit/mock/backend/commands/analyze_mock.c
(2) Do we run CI on these mock test now?

huor:
@liming01, the mock file for analyze is generated. The way to add mock is at: https://github.com/apache/incubator-hawq/blob/master/src/test/unit/README.txt. Please take a look and see if it is good to leverage mock for your fix.

The mock test is covered in hawq ci. You can do that in local environment also: cd src/backend; make unittest-check

If it is hard to reproduce the issue with sql, I think mock is an option we can try.

liming01:
After investigation, it seems we can not use cmock to reproduce HAWQ-1417.
The reason is: we need to mock all functions called in these functions, and also we need make one relation with catalog in memory, and data on disk.
I looked into some unit tests exists in backend, all tests are testing simple logic, and calling some independent modules, so the functions need to be mocked is very limited.

huor:
+1 for the fix as the mock of the verification is not feasible

PR #1210, opened 2017-04-05
HAWQ-1423. cmock framework does not recognize __MAYBE_UNUSED.

paul-guo-:


huor:
+1 for the improving cmock

liming01:
+1.

PR #1211, opened 2017-04-07
HAWQ-1425. Print error message if ssh connect failed.

radarwave:


huor:
+1 for the fix

tools/bin/hawqpylib/hawqlib.py
```diff
@@ -203,6 +203,17 @@ def check_hostname_equal(remote_host, user = ""):
...
+    if result_remote == 255:
+        print "Create ssh connection to %s failed." % remote_host
+        print "Please check the ssh connection and make sure passwordless ssh enabled."
+        sys.exit(result_remote)
+    elif result_remote > 0:
+        print "Execute remote command failed."
+        sys.exit(result_remote)
+    else:
+        pass
+
```
paul-guo-: Judge via 255 seems to be wrong since the cmd could return 255, e.g.
[pguo@host67:/home/pguo]$ ssh 127.0.0.1 'exit 255'
[pguo@host67:/home/pguo]$ echo $?
255

I'm wondering whether we could show the error info in hawq utility output or log file, and we still check result_remote>0 and log for this condition like this:

ssh command returns with error, either ssh connection fails or cmd exits with error. see more in log or.....

radarwave: Thanks @paul-guo- , print error message is good idea. Fixed.

tools/bin/hawqpylib/hawqlib.py
```diff
@@ -203,6 +203,13 @@ def check_hostname_equal(remote_host, user = ""):
     cmd = "hostname"
     result_local, local_hostname, stderr_remote  = local_ssh_output(cmd)
     result_remote, remote_hostname, stderr_remote = remote_ssh_output(cmd, remote_host, user)
+    if result_remote != 0:
+        print stderr_remote
+        print "Execute remote command failed on %s." % remote_host
```
paul-guo-: " with return code %d"  % result_remote?

Do we have error log or output in log file or hawq utility output?

tools/bin/hawqpylib/hawqlib.py
```diff
@@ -203,6 +203,13 @@ def check_hostname_equal(remote_host, user = ""):
...
+        print "Either ssh connection fails or command exits with error."
+        print "For ssh connection issue, please check connection and make sure passwordless ssh enabled."
```
huor:
```
if result_remote != 0:
    print "Failed to execute remote command on %s. Either ssh connection issue or command error. Details:" % remote_host
    print stderr_remote
    print "Please check connection and make sure passwordless ssh is enabled, or check remote host"
```

tools/bin/hawqpylib/hawqlib.py
radarwave: For Q1, added command line and error code in the error message.
For Q2, there are enough error information inside stderr_remote. User can read and know what's happening.

tools/bin/hawqpylib/hawqlib.py
```diff
@@ -203,6 +203,13 @@ def check_hostname_equal(remote_host, user = ""):
     cmd = "hostname"
     result_local, local_hostname, stderr_remote  = local_ssh_output(cmd)
     result_remote, remote_hostname, stderr_remote = remote_ssh_output(cmd, remote_host, user)
+    if result_remote != 0:
+        print "Execute command '%s' failed with return code %d on %s." % (cmd, result_remote, remote_host)
+        print "Either ssh connection fails or command exits with error. Details:"
+        print stderr_remote
+        print "For ssh connection issue, please make sure passwordless ssh is enabled or check remote host."
```
huor: The remote command only check if the host exist.
```
if result_remote != 0:
    print "Execute command '%s' failed with return code %d on %s." % (cmd, result_remote, remote_host)
    print "Please check ssh connection and make sure passwordless ssh is enabled."
```

radarwave: Although a simple command, it's possible the remote OS have some issue can cause failures. So not only the ssh connection error.

PR #1216, opened 2017-04-12
HAWQ-1429. Do not use AggBridge when WHERE clause specified.

sansanichfb:


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/utilities/Utilities.java
```diff
@@ -209,7 +209,8 @@ public static boolean useAggBridge(InputData inputData) {
             LOG.error("Unable to load accessor class: " + e.getMessage());
             return false;
         }
-        return (inputData != null) && (inputData.getAggType() != null)
+        return (inputData != null) && !inputData.hasFilter()
```
kavinderd: Maybe add a comment for why so many checks are required

sansanichfb: Sure, added

shivzone: Add an additional check to filter out queries that have aggregate that is not count(*)

PR #1218, opened 2017-04-13
HAWQ-1431: Do not use StatsAccessor when column specified in SELECT c…

sansanichfb:
…lause.


paul-guo-:
@sansanichfb If the code change has been in main, please close this PR. Thanks.

PR #1222, opened 2017-04-20
HAWQ-1438. Support resource owner beyond transaction boundary

liming01:
(1) Add TopResourceOwner to trace resource used beyond the transaction boundary.
(2) When to auto alloc a new TopResourceOwner for each process?
    Generate a singleton TopResourceOwner when setting CurrentResourceOwner to NULL.
(3) If crashed at some code beyond the transaction boundary, we should manually create
    a new resource owner, because we don't manually create TopResourceOwner for each process.
(4) When to call ResourceOwnerDelete(TopResourceOwner) automatically for each process?
    Register it at on_proc_exit().

### DIFF
--- a/src/backend/access/transam/xact.c
+++ b/src/backend/access/transam/xact.c
@@ -2458,7 +2458,7 @@ CommitTransaction(void)
 	//AtEOXact_Snapshot(true);
 	pgstat_report_xact_timestamp(0);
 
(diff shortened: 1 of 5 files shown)


wengyanqing:
+1

paul-guo-:
+1

PR #1223, opened 2017-04-24
HAWQ-1439. tolerate system time being changed to earlier point when checking resource context timeout

jiny2:


jiny2:
@paul-guo- @liming01  please review

paul-guo-:
This is a workaround. A better solution seems to use the relative time based on H/W tick. I'd test to use clock_gettime() to replace gettimeofday() if clock_gettime() exists.

liming01:
+1. LGTM.

jiny2:
I think as we discussed, we will do a simple fix firstly for this problem. Changing gettime_microsec() may have side effect which is not clear yet.

jiny2:
Seems clock_gettime() does not work for C99. 

network_utils.c:52:11: warning: implicit declaration of function 'clock_gettime' is invalid in C99 [-Wimplicit-function-declaration]
        status = clock_gettime(CLOCK_MONOTONIC, &ts);
                 ^
network_utils.c:52:25: error: use of undeclared identifier 'CLOCK_MONOTONIC'
        status = clock_gettime(CLOCK_MONOTONIC, &ts);
                               ^

jiny2:
Applied @paul-guo- 's comment to change the time getting implementation

src/backend/resourcemanager/utils/network_utils.c
```diff
@@ -41,9 +42,23 @@ static void cleanupSocketConnectionPool(int code, Datum arg);
...
+	newTime.tv_sec = ts.tv_sec;
+	newTime.tv_usec = ts.tv_nsec / 1000;
+#endif
+
+	if (status != 0)
+	{
+		gettimeofday(&newTime, NULL);
+	}
+	t = ((uint64_t)newTime.tv_sec) * USECS_PER_SECOND + newTime.tv_usec;
+    return t;
```
paul-guo-: spaces -> tab.

src/backend/resourcemanager/utils/network_utils.c
```diff
@@ -29,6 +29,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <time.h>
```
paul-guo-: You might add the following header file to suppress the warning about missing definition of on_proc_exit().

#include "storage/ipc.h"

jiny2:
@paul-guo-  please double check the latest changes

PR #1224, opened 2017-04-26
HAWQ-1440. Support ANALYZE for all Hive External Tables

kavinderd:


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveDataFragmenter.java
```diff
@@ -466,7 +466,14 @@ private boolean buildSingleFilter(Object filter,
...
     @Override
     public FragmentsStats getFragmentsStats() throws Exception {
-        throw new UnsupportedOperationException(
-                "ANALYZE for Hive plugin is not supported");
+        Metadata.Item tblDesc = HiveUtilities.extractTableFromName(inputData.getDataSource());
+        Table tbl = HiveUtilities.getHiveTable(client, tblDesc);
+        Metadata metadata = new Metadata(tblDesc);
+        HiveUtilities.getSchema(tbl, metadata);
+
+        long split_count = Long.parseLong(tbl.getParameters().get("numFiles"));
```
sansanichfb: I think split size is configured by "mapred.max.split.size" property of a job, so if a file happens to be bigger than split size - we might have multiple splits per one file.

kavinderd: Hmm, ok I'll use that metadata stat instead

shivzone: Yes, can you please test against a rather large ORC file to see the number of splits/blocks we return ?

kavinderd: @sansanichfb @shivzone It is possible to have files that are larger than an hdfs block/split size. However, I think this is an anomaly especially with ORC where creating many small files is preferred to increase concurrency and parallelism. So based on the fact that this is an edge case and the number of splits is only used by HAWQ in calculating its sampling ratio for statistic collection is the current implementation acceptable?

I personally don't think that getting an accurate number of splits for the ANALYZE case is worth running a function like https://github.com/apache/incubator-hawq/blob/master/pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveDataFragmenter.java#L284 just to get a handle on the number of splits

PR #1226, opened 2017-05-02
HAWQ-1447. Fix ranger build failure

stanlyxiang:
@interma @zhangh43 @linwen @ictmalili  pls review, thanks

### DIFF
--- a/ranger-plugin/Makefile
+++ b/ranger-plugin/Makefile
@@ -57,6 +57,7 @@ install:
 	cp $(top_builddir)$(subdir)/conf/server.xml                                        $(ranger_dir)/plugin-service/conf
 	cp $(top_builddir)$(subdir)/conf/catalina.properties                               $(ranger_dir)/plugin-service/conf
 	cp $(top_builddir)$(subdir)/service/target/ranger-plugin-service-*/WEB-INF/lib/*   $(ranger_dir)/plugin-service/lib/
+	rm $(ranger_dir)/plugin-service/lib/jersey-*-1.9.jar
 	cp $(top_builddir)$(subdir)/service/target/ranger-plugin-service-*.war             $(ranger_dir)/plugin-service/webapps/rps.war
 
 uninstall:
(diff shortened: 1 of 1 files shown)


zhangh43:
+1

edespino:
@stanlyxiang - What is the status of this PR?  It's been idle for over a month.

stanlyxiang:
I tested on my rhel7. After build rps rpm, there are 2 commands after the rpm generate.
`
[INFO] + cd /home/xsheng/hawq/incubator-hawq/ranger-plugin/target/rpm/hawq-ranger-plugin/BUILD
[INFO] + /usr/bin/rm -rf /home/xsheng/hawq/incubator-hawq/ranger-plugin/target/rpm/hawq-ranger-plugin/buildroot
`
Since previous install method is copying the ranger folder under buildroot, but now the content of BuildRoot is deleted automatically on rhel7. Because the rpmbuild version is 4.11.3. 
I searched on the internet and found that the old version of rpmbuild will keep the folder and we can copy the files to installdir. But on rhel7 the rpmbuild is high version and will remove the buildroot. So the installation will fail. 

There are some workarounds I haven't verified because they are required to modify the spec file when build rpm. But the spec file for build ranger-plugin rpm is generated automatically. 

I checked rpmbuild version on my rhel6 and rhel7. RPM version 4.8.0 and RPM version 4.11.3. They all deleted the buildroot folder after build. So there is the pr. If you have any solution, please commets. Thanks in advance. 

On my mac the rpm is "rpm (RPM) 5.4.15" and "rpmbuild (RPM) 5.4.15", it won't remove "buildroot". I think it's rpm5. 

Concourse ci build and install the ranger-plugin.rpm separately so it's green. If we use rpm to install the ranger-plugin during hawq installation, we should use "rpm --prefix=/hawq/install/path  ranger-plugin.rpm". But I tried and got an error: 
`rpm: --prefix may only be used when installing new packages`. So it not reality for a community user or developer to rpm uninstall and rpm install everytime they have changes.

interma:
Thanks a lot, @stanlyxiang 
I see. +1

PR #1227, opened 2017-05-02
HAWQ-1448. Fixed postmaster process hung at recv () on segment

liming01:
Hi all,

I don't know is there any side effect of this fix, but I think from theory it should be okay, so prefer to to checkin and monitor for a period of time. Please help me to review it. Thanks.

### DIFF
--- a/tools/bin/hawq_ctl
+++ b/tools/bin/hawq_ctl
@@ -909,9 +909,16 @@ class HawqStop:
                       (source_hawq_env, self.GPHOME, self.segment_data_directory, log_filename)
             return cmd_str
         else:
+            # If stop cluster, firstly master stopped, then segment maybe hung at recv() from master.
+            # So here change segment stop mode from smart to fast, so that segment don't wait
+            # connection from master to be closed.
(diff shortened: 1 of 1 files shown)


paul-guo-:
Normally stopped master should not lead to recv() hang on segments since recv() will return 0 if the tcp connection (libpq) is closed.

edespino:
@liming01 - What is the status of this PR?  It's been idle for over a month.

liming01:
@edespino,  if we can find all hung root cause when hawq stop and fix them, then we don't need this PR. 

However from a long time investigate, these problems still cannot been fixed. So I suggest to apply this PR. This PR can fix all hung problems when hawq stop, (if still hung in some cases, we should check not disable interrupt processing during the hung function called).

Thanks.

radarwave:
When master is down,  it's no matter to stop segment in 'smart' or 'fast' mode. And I see this fix only affect 'hawq stop cluster', which means master is already down before stoping segments. So I think this fix is good enough to me.

+1

wengyanqing:
Agree with Radar's comment. LGTM.

PR #1230, opened 2017-05-03
HAWQ-1453. Fixed relation_close() error at analyzeStmt(): not owned b…

liming01:
…y resource owner TopTransaction (resowner.c:814)

The relation opened at TopResourceOwner, while close at a new transaction resource owner.
So when we close these relations, we need to firstly switch back to the TopResourceOwner.

Verified on reproducible environment. Please help me to review it. Thanks.

### DIFF
--- a/src/backend/commands/analyze.c
+++ b/src/backend/commands/analyze.c
@@ -308,7 +308,7 @@ void analyzeStmt(VacuumStmt *stmt, List *relids, int preferred_seg_num)
 	ListCell				*le1 = NULL;
 	int 					successCount = 0, failCount = 0;
 	StringInfoData 			failNames;
-	ResourceOwner owner, oldOwner;
+	ResourceOwner asOwner, oldOwner1;  /* asOwner for analyzeStmt resource owner */
 
 	/**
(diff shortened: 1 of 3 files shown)


paul-guo-:
Personally I do not like having asOwner and oldOwner1, but it is not a big issue. So +1

src/backend/commands/analyze.c
```diff
@@ -308,7 +308,7 @@ void analyzeStmt(VacuumStmt *stmt, List *relids, int preferred_seg_num)
 	ListCell				*le1 = NULL;
 	int 					successCount = 0, failCount = 0;
 	StringInfoData 			failNames;
-	ResourceOwner owner, oldOwner;
+	ResourceOwner asOwner, oldOwner1;
```
hsyuan: if `as` of `asOwner` means `analyzeStmt`,  we'd better add comments or explanation on the both vars.

liming01: @hsyuan and @paul-guo-,  Add comments for asOwner. Thanks.

radarwave:
+1

PR #1232, opened 2017-05-05
HAWQ-1455. Wrong results on CTAS query over catalog

paul-guo-:
This reverts the previous fix for HAWQ-512, however to HAWQ-512, it looks like
we could modify the lock related code following gpdb to do a real fix. Those code
was probably deleted during early hawq development.

### DIFF
--- a/src/backend/storage/lmgr/lock.c
+++ b/src/backend/storage/lmgr/lock.c
@@ -1019,7 +1019,6 @@ LockCheckConflicts(LockMethod lockMethodTable,
 				   PGPROC *proc)
 {
 	int			numLockModes = lockMethodTable->numLockModes;
-	LOCKMASK	myLocks;
 	LOCKMASK	otherLocks;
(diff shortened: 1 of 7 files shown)


ztao1987:
Seems these codes comes from a bunch of bug fix involved from GPDB commit 6b0e52beadd678c5050af9c978c26d171ba86ae0. For this specific bug, I have no doubt it works. But for lock mechanism, gpdb fix involves more than that. Not sure if there is potential issue in lock.

paul-guo-:
My guess is that we simplified the lock mechanism, i.e. we removed some
code from gpdb. That patch seems to be the whole gpdb patch against pg for
open source.

2017-05-10 17:14 GMT+08:00 zhenglin tao <notifications@github.com>:

ztao1987:
Looks good to me. +1

huor:
+1 for the fix

PR #1235, opened 2017-05-09
HAWQ-1456. Copy RPS configuration files to standby in specific scenarios

interma:
copy RPS configuration files to standby in:
  * hawq init standby
  * enable-ranger-plugin.sh

besides copy files, enhance enable-ranger-plugin.sh to support read items(e.g. hawq_master_address_host) in hawq-site.xml now.

### DIFF
--- a/ranger-plugin/scripts/enable-ranger-plugin.sh
+++ b/ranger-plugin/scripts/enable-ranger-plugin.sh
@@ -70,16 +70,16 @@ function get_ranger_password() {
   done
 }
 
-# get tag value from hawq-site.xml
-function get_value_bytag() {
-    local hawq_site_file="$GPHOME/etc/hawq-site.xml"
-	tag=$1
+# get property value from hawq-site.xml
+function get_hawq_property() {
+  local hawq_site_file="$GPHOME/etc/hawq-site.xml"
+  tag=$1
 
-	if [ -f $hawq_site_file ] ; then
-		value=`cat $hawq_site_file | tr '\n' ' ' | awk -F '<property>' '{ for(i = 1; i <= NF; i++) { print $i; } }' | grep $tag | sed -n 's|.*<value>\(.*\)</value>.*|\1|p'`
-	else
-		value=''
-	fi
+  if [ -f $hawq_site_file ] ; then
+    value=`cat $hawq_site_file | tr '\n' ' ' | awk -F '<property>' '{ for(i = 1; i <= NF; i++) { print $i; } }' | grep $tag | sed -n 's|.*<value>\(.*\)</value>.*|\1|p'`
+  else
+    value=''
+  fi
 }
 
 function get_hawq_url() {
@@ -89,19 +89,19 @@ function get_hawq_url() {
   # 2. read from hawq-site.xml
   if [[ -z "$HAWQ_URL" ]]; then
     value=''
-	get_value_bytag hawq_master_address_host
-	local host=$value
+    get_hawq_property hawq_master_address_host
+    local host=$value
 
-	value=''
-	get_value_bytag hawq_master_address_port
-	local port=$value
+    value=''
+    get_hawq_property hawq_master_address_port
+    local port=$value
 
-	if [[ -z "$host" || -z "$port" ]]; then
-		HAWQ_URL=''
-	else
-		HAWQ_URL="$host:$port"
-	fi
-  fi
+    if [[ -z "$host" || -z "$port" ]]; then
+      HAWQ_URL=''
+    else
+      HAWQ_URL="$host:$port"
+    fi
+fi
 
   # 3. read from user input
   local default=`hostname -f`
@@ -125,14 +125,14 @@ function get_hawq_url() {
   # 2. read from hawq-site.xml
   if [[ -z "$HAWQ_STANDBY_HOST" ]]; then
     value=''
-	get_value_bytag hawq_standby_address_host
-	local host=$value
+    get_hawq_property hawq_standby_address_host
+    local host=$value
 
-	if [[ "$host" == "none" ]]; then
-		HAWQ_STANDBY_HOST=''
-	else
-		HAWQ_STANDBY_HOST=$host
-	fi
+    if [[ "$host" == "none" ]]; then
+      HAWQ_STANDBY_HOST=''
+    else
+      HAWQ_STANDBY_HOST=$host
+    fi
   fi
 
   # 3. read from user input
@@ -158,15 +158,13 @@ function get_hawq_password() {
 }
 
 function sync_rps_configuration() {
-	if [[ -d $GPHOME/ranger/etc/ && ! -z $HAWQ_STANDBY_HOST ]]; then
-		cmd="scp $GPHOME/ranger/etc/* $HAWQ_STANDBY_HOST:$GPHOME/ranger/etc/"
-		#echo $cmd
-		#cmd="scp /usr/local/hawq/ranger/etc/* $HAWQ_STANDBY_HOST:/tmp/"
-		$cmd
-		if [ $? != 0 ]; then
-			fail "Scp RPS configuration files to Standby failed."
-		fi
-	fi
+  if [[ -d $GPHOME/ranger/etc/ && ! -z $HAWQ_STANDBY_HOST ]]; then
+    cmd="scp $GPHOME/ranger/etc/* $HAWQ_STANDBY_HOST:$GPHOME/ranger/etc/"
+    $cmd
+    if [ $? != 0 ]; then
+      fail "Scp RPS configuration files to Standby failed."
+    fi
+  fi
 }
 
 function parse_params() {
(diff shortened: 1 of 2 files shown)


interma:
@linwen @stanlyxiang @zhangh43 help to review, tks!

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -70,14 +70,47 @@ function get_ranger_password() {
   done
 }
 
+# get tag value from hawq-site.xml
+function get_value_bytag() {
```
linwen: suggest use "property" instead of "tag".

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -104,8 +157,20 @@ function get_hawq_password() {
   done
 }
 
+function sync_rps_configuration() {
```
linwen: should follow 2 spaces indent.

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -237,9 +307,14 @@ main() {
...
     usage
   fi
+  if [[ -z "$GPHOME" ]]; then
+	GPHOME="/usr/local/hawq"
+  fi
   SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P)"
   parse_params "$@"
   validate_params
+  sync_rps_configuration
+  #exit 0
```
linwen: useless codes.

interma: all fixed, thanks!

tools/bin/hawq_ctl
```diff
@@ -387,6 +387,10 @@ class HawqInit:
                  (self.GPHOME, self.standby_host_name, self.GPHOME)
         check_return_code(remote_ssh(scpcmd, self.master_host_name, self.user), \
                           logger, "Sync slaves file failed")
+        scpcmd = "scp %s/ranger/etc/* %s:%s/ranger/etc/ > /dev/null" % \
+                 (self.GPHOME, self.standby_host_name, self.GPHOME)
+        check_return_code(remote_ssh(scpcmd, self.master_host_name, self.user), \
+                          logger, "Sync rps configuration files failed")
```
stanlyxiang: This code snippet is used for copy ranger configuration files to standby when do "hawq init standby" when ranger was configured.  But there maybe no ranger.  Standalone case will fail.
e.g. When hawq init cluster(it will init standby), hawq is absolutely not configured with ranger. If there is a cluster with hawq master and segment configured with ranger, then we should scp rps configuration files to standby when "hawq init standby".

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -104,8 +157,18 @@ function get_hawq_password() {
   done
 }
 
+function sync_rps_configuration() {
+  if [[ -d $GPHOME/ranger/etc/ && ! -z $HAWQ_STANDBY_HOST ]]; then
+    cmd="scp $GPHOME/ranger/etc/* $HAWQ_STANDBY_HOST:$GPHOME/ranger/etc/"
+    $cmd
+    if [ $? != 0 ]; then
+      fail "Scp RPS configuration files to Standby failed."
```
stanlyxiang: "scp" is a command, weird to say "Scp".

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -237,9 +305,13 @@ main() {
   if [[ $# -lt 1 ]]; then
     usage
   fi
+  if [[ -z "$GPHOME" ]]; then
+    GPHOME="/usr/local/hawq"
```
stanlyxiang: what if user not set GPHOME and "/usr/local/hawq" is not the right place ?

tools/bin/hawq_ctl
interma: Thanks! I forgot this conditions.

ranger-plugin/scripts/enable-ranger-plugin.sh
interma: when no GPHOME and "/usr/local/hawq", code run here:
https://github.com/interma/interma-hawq/blob/a49f85566fd136cc7940fdff15d50c985a99d2cc/ranger-plugin/scripts/enable-ranger-plugin.sh#L111

It ask user to input, so it's ok.

ranger-plugin/scripts/enable-ranger-plugin.sh
interma: ok, thanks.

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -70,14 +70,47 @@ function get_ranger_password() {
   done
 }
 
+# get property value from hawq-site.xml
+function get_hawq_property() {
+  local hawq_site_file="$GPHOME/etc/hawq-site.xml"
```
denalex: I would use relative path to the script location instead of relying on GPHOME

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -70,14 +70,47 @@ function get_ranger_password() {
...
 function get_hawq_url() {
-  #todo read hawq-site.xml ?
+
+  # get hawq master host and port
+  # 1. read from command parameter -h
+  # 2. read from hawq-site.xml
+  if [[ -z "$HAWQ_URL" ]]; then
+    value=''
+    get_hawq_property hawq_master_address_host
+    local host=$value
```
denalex: if you have function's last line do echo $value, then here you can just say local host = $(func) instead of setting/resetting global variable

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -86,6 +119,27 @@ function get_hawq_url() {
...
+
+    if [[ "$host" == "none" ]]; then
+      HAWQ_STANDBY_HOST=''
+    else
+      HAWQ_STANDBY_HOST=$host
+    fi
+  fi
+
+  # 3. read from user input
+  while [[ -z "$HAWQ_STANDBY_HOST" ]]
```
denalex: what if there is no standby host at all, why would we force the user to enter the value here ?

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -104,8 +157,18 @@ function get_hawq_password() {
   done
 }
 
+function sync_rps_configuration() {
+  if [[ -d $GPHOME/ranger/etc/ && ! -z $HAWQ_STANDBY_HOST ]]; then
```
denalex: I'd print a statement telling user what the script is about to do as well as the status once done

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -104,8 +157,18 @@ function get_hawq_password() {
...
+    cmd="scp $GPHOME/ranger/etc/* $HAWQ_STANDBY_HOST:$GPHOME/ranger/etc/"
```
denalex: why do we need a variable and not run the command directly ?

denalex: are we sure we want to copy the files here as part of this script ? I'd say it would be better to have it working on a single node only and do any cluster-level operations in "hawq" management scripts. Also, in the future config files will be managed by Ambari. I'd say it even could be ok to ask users to copy the files to standby manually, especially since when failover occurs, they will need to provision a new standby and will have to copy config files there manually anyways.

ranger-plugin/scripts/enable-ranger-plugin.sh
denalex: we do not need GPHOME, just use relative path to current script. No reason to ask users to source extra files.

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -237,9 +305,13 @@ main() {
   if [[ $# -lt 1 ]]; then
     usage
   fi
+  if [[ -z "$GPHOME" ]]; then
+    GPHOME="/usr/local/hawq"
+  fi
   SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd -P)"
   parse_params "$@"
   validate_params
+  sync_rps_configuration
```
denalex: if sync fails, Ranger definitions are not created, which is not good. Service can work without standby and without sync, so either not fail (just warn) or move it to the very last step. Better yet, not to sync at all here, as I commented above.

interma: thanks, alex!
all comments are useful, I will fix them.

ranger-plugin/scripts/enable-ranger-plugin.sh
interma: You are right, I saw *SCRIPT_DIR*, I will use it.

ranger-plugin/scripts/enable-ranger-plugin.sh
interma: Reasonable.
But there are some other stuffs(rps ha tests,docs) assumed the script should copy rps configuration files to standby...
I think copy is also ok, and we can finish other stuffs quickly.

In future, we will go back to consider whether remove copy is more easy-to-use for user.

interma:
All fixed, help to review again.
Thanks!

interma:
Removed scp rps files.
Reserved read property from hawq-site.xml (read hawq_master_address_host and hawq_master_address_port).

linwen:
+1

interma:
merged

PR #1241, opened 2017-05-16
HAWQ-1436. Print a message to command line if hawq switches from standby RPS to master RPS

linwen:
This improvement happens in such case: hawq master is talking with standby RPS, when master's RPS has recovered, master switches from standby RPS to local RPS, a reminder message is printed.
Please review.

### DIFF
--- a/src/backend/libpq/rangerrest.c
+++ b/src/backend/libpq/rangerrest.c
@@ -386,6 +386,7 @@ static int call_ranger_rest(CURL_HANDLE curl_handle, const char* request)
 	int retry = 2;
 	CURLcode res;
 	Assert(request != NULL);
+	bool switchToMaster = false;
 
 	/*
 	 * If master is talking with standby RPS, for every predefined interval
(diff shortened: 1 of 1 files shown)


src/backend/libpq/rangerrest.c
```diff
@@ -464,6 +466,11 @@ static int call_ranger_rest(CURL_HANDLE curl_handle, const char* request)
 		}
 		else
 		{
+			if (switchToMaster)
+			{
+				/* master's RPS has recovered, switch from standby's RPS to master's RPS */
+				elog(NOTICE, "switch from standby's RPS to master's RPS");
```
interma: When this message logged, hawq maybe still talking to standby RPS, right?
example:
```
talk to standby before;
time up, set switchToMaster = true;
talk to master but failed.
retry talk to standby rps successfully
if (switchToMaster) elog(...)
```

may be a bit confusing to user: 
saw log "switch back to master", but still talking to standby now.
looks like elog this is better?
```
"tried to switch back, but %s", talkingWithStandby? "failed":"success";
```

And keep it as it's now is also well.  🙂

linwen: Thanks! You are right.
If switch to master and failed again, then talk with standby success, this notice message is also printed, a condition should be added.

interma:
+1

PR #1242, opened 2017-05-17
HAWQ-1469. Don't expose warning messages to command line when switch between master and standby RPS.

linwen:
Please review. Thanks!

### DIFF
--- a/src/backend/libpq/rangerrest.c
+++ b/src/backend/libpq/rangerrest.c
@@ -468,7 +468,7 @@ static int call_ranger_rest(CURL_HANDLE curl_handle, const char* request)
 			}
 			else
 			{
-				elog(ERROR, "permission is unknown due to authorization failure,"
+				elog(ERROR, "permission is unknown due to authorization failure, "
 					"ranger plugin service is unavailable : %s.\n", curl_easy_strerror(res));
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/backend/libpq/rangerrest.c
```diff
@@ -453,23 +453,30 @@ static int call_ranger_rest(CURL_HANDLE curl_handle, const char* request)
...
 			{
-				elog(WARNING, "ranger plugin service from http://%s:%d/rps is unavailable : %s, try another http://%s:%d/rps\n",
+				/* Don't expose this warning message to client, just record in log.
+				 * The value of whereToSendOutput is DestRemote, so set it to DestNone
+				 * and set back after write a warning message in log file.
+				 */
+				CommandDest commandDest = whereToSendOutput;
+				whereToSendOutput = DestNone;
+				elog(WARNING, "ranger plugin service from http://%s:%d/rps is unavailable : %s, "
+						"trying ranger plugin service at http://%s:%d/rps\n",
```
paul-guo-: If so, why not just remove the code and add comments here? Does elog(LOG, ...) possibly expose this to users?

linwen: When master RPS doesn't work due to some reason, and hawq begins to talk with standby RPS, a warning message should be recorded in log file, so that administrators can solve the master RPS problem. elog(LOG, ...) won't expose to console(by default, client_min_sessages is NOTICE), but in this switch case, a warning should be recorded.

paul-guo-:
+1

PR #1244, opened 2017-05-26
HAWQ-1443. Implement Ranger lookup for HAWQ with Kerberos enabled.

zhangh43:
This PR implements the HAWQ client in ranger to lookup HAWQ resource in kerberos environment.
With Kerberos authentication, the jdbc URL looks like:  jdbc:postgresql://host:port/db?kerberosServerName=HawqKerberosServiceName&jaasApplicationName=pgjdbc&user=LookupPricipal
With Simple authentication,  the jdbc URL just looks like:  jdbc:postgresql://host:port/db and with user name and password as parameter.

In Kerberos env, we also do kinit with principal name and principal password. Keytab file is not supported yet.

Signed-off-by: xiang sheng <stanly.sxiang@gmail.com>


linwen:
+1

PR #1245, opened 2017-05-27
HAWQ-1476. Augment enable-ranger-plugin.sh to support kerberos.

zhangh43:
Now ranger can lookup hawq resource in kerberized environment. So we also need to change enable-ranger-plugin.sh to support automatically fill the authentication type and hawq kerberos service name fields

### DIFF
--- a/ranger-plugin/conf/ranger-servicedef-hawq.json
+++ b/ranger-plugin/conf/ranger-servicedef-hawq.json
@@ -244,7 +244,7 @@
       "name": "authentication",
       "type": "enum",
       "subType": "authType",
-      "mandatory": false,
+      "mandatory": true,
       "validationRegEx": "",
       "validationMessage": "",
       "uiHint": "",
--- a/ranger-plugin/scripts/enable-ranger-plugin.sh
+++ b/ranger-plugin/scripts/enable-ranger-plugin.sh
@@ -20,7 +20,7 @@
 #
 
 function usage() {
-  echo "USAGE: enable-ranger-plugin.sh -r ranger_host:ranger_port -u ranger_user -p ranger_password [-h hawq_host:hawq_port] -w hawq_user -q hawq_password"
+  echo "USAGE: enable-ranger-plugin.sh -r ranger_host:ranger_port -u ranger_user -p ranger_password [-h hawq_host:hawq_port -c hawq_kerberos_service_name] -w hawq_user -q hawq_password -t lookup_authentication_type"
   exit 1
 }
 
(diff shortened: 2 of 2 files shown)


ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -20,7 +20,7 @@
 #
 
 function usage() {
-  echo "USAGE: enable-ranger-plugin.sh -r ranger_host:ranger_port -u ranger_user -p ranger_password [-h hawq_host:hawq_port] -w hawq_user -q hawq_password"
+  echo "USAGE: enable-ranger-plugin.sh -r ranger_host:ranger_port -u ranger_user -p ranger_password [-h hawq_host:hawq_port -c hawq_kerberos_service_name] -w hawq_user -q hawq_password -t lookup_authentication_type"
```
stanlyxiang: since lookup_authentication_type will determine whether to have option "-c hawq_kerberos_service_name", I think it would be better to place option "lookup_authentication_type" before "-c hawq_kerberos_service_name".

ranger-plugin/conf/ranger-servicedef-hawq.json
```diff
@@ -244,7 +244,7 @@
       "name": "authentication",
       "type": "enum",
       "subType": "authType",
-      "mandatory": false,
+      "mandatory": true,
```
denalex: are we doing this lookup-side Kerberos support in case where HAWQ authentication is set to Kerberos in pg_hba.conf ? This is not supported by default when we kerberize the cluster, so usually when the whole cluster (including HAWQ) is secured, password based connection to HAWQ will still work and kerberos-based connection will not be setup by default.

ictmalili: Yes, we aim to update the connection to HAWQ using Kerberos when Kerberos authentication is set in pg_hba.conf. 
Why don't Ambari change pg_hba.conf in HAWQ once Kerberos is setup? I think we need protect connection to HAWQ.

ictmalili:
LGTM. +1

ranger-plugin/scripts/enable-ranger-plugin.sh
```diff
@@ -20,7 +20,7 @@
...
   exit 1
 }
```
ictmalili: I think "-t lookup_authentication_type" should be moved in front of "-c hawq_kerberos_service_name", and should also be marked as optional with default value "simple". 
"-c hawq_kerberos_service_name" can be changed to "-s", since it indicates the service name.

ranger-plugin/conf/ranger-servicedef-hawq.json
denalex: We have decided a while ago to not touch pg_hba.conf in Ambari, because users might have their own entries there and this is a very sensitive file. A viable approach might've been to manage all entries in pg_hba.conf via Ambari, but that required extra work.

PR #1249, opened 2017-06-01
HAWQ-1461. Improve partition parameters validation for PXF-JDBC plugin

lavjain:


pxf/build.gradle
```diff
@@ -477,7 +481,7 @@ def buildNumber() {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '2.11'
+    gradleVersion = '2.14.1'
```
sansanichfb: Why do we need to update Gradle version?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/StatsAccessor.java
```diff
@@ -29,11 +29,13 @@
 
     /**
      * Method which reads needed statistics for current split
+     * @throws Exception if retrieving the stats failed
      */
     public void retrieveStats() throws Exception;
 
     /**
      * Returns next tuple based on statistics information without actual reading of data
+     * @return only one row
```
sansanichfb: @return next row without reading it from disk

pxf/build.gradle
lavjain: We are currently using 2 different versions 2.11 and 2.13 (pxf/gradle/wrapper/gradle-wrapper.properties)

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenter.java
```diff
@@ -111,30 +110,40 @@ public JdbcPartitionFragmenter(InputData inConf) throws UserDataException {
...
-        } catch (IllegalArgumentException e1) {
+        } catch (IllegalArgumentException | ArrayIndexOutOfBoundsException e1) {
             throw new UserDataException("The parameter 'PARTITION_BY' invalid, the pattern is 'column_name:date|int|enum'");
         }
 
+        //parse and validate parameter-RANGE
         try {
-            range = inConf.getUserProperty("RANGE").split(":");
+            if (inConf.getUserProperty("RANGE") != null) {
+                range = inConf.getUserProperty("RANGE").split(":");
```
denalex: calling same function twice, would be better to call it once saving result and then using the result for further logic

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenter.java
```diff
@@ -111,30 +110,40 @@ public JdbcPartitionFragmenter(InputData inConf) throws UserDataException {
...
             partitionType = PartitionType.getType(partitionBy[1]);
-        } catch (IllegalArgumentException e1) {
+        } catch (IllegalArgumentException | ArrayIndexOutOfBoundsException e1) {
             throw new UserDataException("The parameter 'PARTITION_BY' invalid, the pattern is 'column_name:date|int|enum'");
         }
 
+        //parse and validate parameter-RANGE
         try {
-            range = inConf.getUserProperty("RANGE").split(":");
+            if (inConf.getUserProperty("RANGE") != null) {
```
denalex: "RANGE" should be an enum value or at least a constant somewhere

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenter.java
```diff
@@ -111,30 +110,40 @@ public JdbcPartitionFragmenter(InputData inConf) throws UserDataException {
...
+                range = inConf.getUserProperty("RANGE").split(":");
+                if (range.length == 1 && partitionType != PartitionType.ENUM)
+                    throw new UserDataException("The parameter 'RANGE' does not specify '[:end_value]'");
```
denalex: why did we drop support for open-ended range ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenter.java
```diff
@@ -144,7 +153,7 @@ public JdbcPartitionFragmenter(InputData inConf) throws UserDataException {
                 rangeEnd.setTime(df.parse(range[1]));
             }
         } catch (ParseException e) {
-            throw new UserDataException("The parameter 'RANGE' include invalid date format.");
+            throw new UserDataException("The parameter 'RANGE' has invalid date format.");
```
denalex: should we also say what expected format is, such as : "yyyy-MM-dd" ?

pxf/pxf-jdbc/src/test/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenterTest.java
```diff
@@ -164,8 +163,8 @@ public void inValidParameterFormat() throws Exception {
         when(inputData.getUserProperty("RANGE")).thenReturn("100:200");
         try {
             new JdbcPartitionFragmenter(inputData);
-            fail("Expected an ArrayIndexOutOfBoundsException");
-        } catch (ArrayIndexOutOfBoundsException ex) {
+            fail("Expected an UserDataException");
```
denalex: use @Test(expected = UserDataException.class)  so that there is no need for try / catch / fail lines : https://github.com/junit-team/junit4/wiki/exception-testing

pxf/pxf-jdbc/src/test/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenterTest.java
```diff
@@ -211,6 +210,90 @@ public void inValidIntervaltype() throws Exception {
...
+    public void testIntervaltypeMissingValid() throws Exception {
+        prepareConstruction();
+        when(inputData.getDataSource()).thenReturn("sales");
+        when(inputData.getUserProperty("PARTITION_BY")).thenReturn("year:int");
+        when(inputData.getUserProperty("RANGE")).thenReturn("2001:2012");
+        when(inputData.getUserProperty("INTERVAL")).thenReturn("1");
+
+        JdbcPartitionFragmenter fragment = new JdbcPartitionFragmenter(inputData);
+        List<Fragment> fragments = fragment.getFragments();
+        assertEquals(fragments.size(), 11);
```
denalex: the order of params for assertEquals usually goes (expected, actual) : http://junit.sourceforge.net/javadoc/org/junit/Assert.html#assertEquals(java.lang.Object, java.lang.Object) as it will report errors as such, so it might be confusing when error happens what was actually expected and what was returned.

pxf/build.gradle
```diff
@@ -111,6 +111,10 @@ subprojects { subProject ->
             force 'org.codehaus.jackson:jackson-mapper-asl:1.9.13'
             force 'junit:junit:4.11'
         }
+        configurations {
+            exclude group: 'org.json', module: 'json'
```
denalex: are we sure it will run OK without it ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPartitionFragmenter.java
sansanichfb: If we want to support open-ended range on fragmenter phase - we might have the last fragment having more data than rest of fragments, which is acceptable and very user-friendly so +1 for having open-ended range.

denalex: in the similar fashion, we might support open ended lower range as well -- in cases where min value is not known or changes dynamically.

lavjain: Supporting open-ended range would require changes to the JDBC code. The current implementation does not support it.

PR #1250, opened 2017-06-05
Hawq 1477. Ranger-plugin connect to Ranger admin under kerberos security.

interma:


interma:
@linwen @stanlyxiang @zhangh43 help to review, tks!

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
```diff
@@ -45,8 +47,43 @@
     /**
      * Constructor. Creates a new instance of the resource that uses <code>RangerHawqAuthorizer</code>.
      */
-    public RangerHawqPluginResource() {
+    public RangerHawqPluginResource()
+    {
+        // set UserGroupInformation under kerberos authentication
+        if (Utils.getAuth().equals("kerberos"))
```
stanlyxiang: using variable instead of using "kerberos" string everywhere.

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/Utils.java
```diff
@@ -68,6 +77,30 @@ public static String getVersion() {
     }
 
     /**
+     * Retrieves the authentication
+     * @return kerberos or simple[default]
+     */
+    public static String getAuth() {
+        return System.getProperty(AUTH_KEY_ENV, properties.getProperty(AUTH_KEY_FILE, "simple"));
```
stanlyxiang: public static final String SIMPLE="simple"; and use SIMPLE here. So does EMPTY_STRING for ""

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
```diff
@@ -45,8 +47,43 @@
...
+                try
+                {
+                    UserGroupInformation.loginUserFromKeytab(prin, keytab);
+                }
+                catch (Exception e)
+                {
+                    LOG.warn(String.format("loginUserFromKeytab failed, user[%s], keytab[%s]", prin, keytab));
+                }
+            }
+        }
```
stanlyxiang: To my understanding for this logic, principal and keytab are needed for loginUserFromKeytab.But the default value of the 2 configurations are empty, empty values are not work for kerberos, if user set RPS_AUTH to kerberos and leave them empty, he will only get a warning of "get login user failed" until the program tries to getLoginUser(). So why don't we check them earlier ?

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
```diff
@@ -45,8 +47,43 @@
...
+
+        try
+        {
+            UserGroupInformation user = UserGroupInformation.getLoginUser();
+            LOG.info(String.format("login user: %s", user));
+        }
+        catch (Exception e)
+        {
+            LOG.warn("get login user failed");
```
stanlyxiang: If get login user failed, this constructor can run well anyway ?

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
```diff
@@ -45,8 +47,43 @@
...
+                {
+                    LOG.warn(String.format("loginUserFromKeytab failed, user[%s], keytab[%s]", prin, keytab));
+                }
+            }
+        }
+
+        try
+        {
+            UserGroupInformation user = UserGroupInformation.getLoginUser();
+            LOG.info(String.format("login user: %s", user));
```
denalex: this should be debug level, we do not want to log into log file every time request is processed normally

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
denalex: always log more details, like exception type and message and context information, current message doesn't help to troubleshoot.

interma: Thanks, I will fix them!

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
interma: It uses the kinit credential cache to when auth method set to authentication and two of them are empty. 
So it's ok here.

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
interma: Yes, if login failed here, it just uses the simple auth method(e.g. gpadmin), but it usually fail in the below code (try use simple auth user to login a kerberized sevice).

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
interma: Yes, but I think use `org.apache.hadoop.security.UserGroupInformation.AuthenticationMethod` is better:
```
  public static enum AuthenticationMethod {
    // currently we support only one auth per method, but eventually a 
    // subtype is needed to differentiate, ex. if digest is token or ldap
    SIMPLE(AuthMethod.SIMPLE,
        HadoopConfiguration.SIMPLE_CONFIG_NAME),
    KERBEROS(AuthMethod.KERBEROS,
        HadoopConfiguration.USER_KERBEROS_CONFIG_NAME),
    TOKEN(AuthMethod.TOKEN),
    CERTIFICATE(null),
    KERBEROS_SSL(null),
    PROXY(null);
```
Let's refactor it in future, keep it as it is now.

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
stanlyxiang: So if it usually fail in the below code due to try use simple auth user to login a kerberized service, why do we let it go failure and why don't we abort here ?

interma: Two reasons:
1. Abort immediately may be not appropriate, sometimes it can run happily in below code (e.g. set simple auth and get login_user() failed by some reasons, but will call a no-secure webservice later)
2. Another small reason: this code locates in a constructor, I didn't find a easy way to deal with exception... 

So I just leave it as it is, it will fail in the first place which need auth in the below code (I tested, it usually prints a warning message contains "HTTP 401/403" in the log, looks well).

ranger-plugin/service/src/main/java/org/apache/hawq/ranger/authorization/RangerHawqPluginResource.java
interma: But I forgot to deal with lower/upper case, I will fix it and add a enum for auth method.
Thanks xiang and wen~

interma:
fixed all comments issues.

zhangh43:
+1

interma:
Merged.

PR #1253, opened 2017-06-13
HAWQ-1485. Use user/password instead of credentials cache in Ranger lookup for HAWQ with Kerberos enabled.

interma:


ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/RangerServiceHawq.java
```diff
@@ -25,41 +25,48 @@
...
+            // try normal password (user input in webform)
+            result = check_connection(configs);
+            if (!(boolean)(result.get("connectivityStatus"))) {
+                retry = true;
+            }
+
+            if (retry) {
+                // try decrypt password
+                decrypt_password(configs);
+                result = check_connection(configs);
```
interma: Explain why try check_connection() twice in this validateConfig():

validateConfig() is called in ranger UI "test connection" buttion, and the password stored in ranger db is encrypted (I have tested).

There are two situations when user clicks "test connection" button:
1. User already inputed a new password, and the password is plain(not encrypted). 
2. User didn't change anything, just click button, but the password is encrypted (fetch directly from ranger db).

Since there is no way to identify these two situations, we just try twice: 
plain password and decrypted password, test connection is success if one of them is passed.

ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/HawqClient.java
```diff
@@ -91,125 +88,42 @@
...
+
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("<== HawqClient.checkConnection configuration" + connectionProperties );
+        }
+
+        if (connectionProperties.containsKey(AUTHENTICATION) && connectionProperties.get(AUTHENTICATION).equals(KERBEROS)) {
+            //kerberos mode
+            props.setProperty("kerberosServerName", connectionProperties.get("principal"));
+            props.setProperty("jaasApplicationName", "pgjdbc");
+        }
```
interma: just need add the two properties in kerberos mode in jdbc.

interma:
@linwen @stanlyxiang @zhangh43 help to review, thanks!

zhangh43:
+1

PR #1256, opened 2017-06-15
HAWQ-1485. Fix exception of decryptPassword twice in lookupResource()

interma:
In my earlier PR,  *decryptPassword()* is called in the beginning of *lookupResource()*, but *BaseClient* (in ranger lib) also need decrypt password in some situation, decrypt twice will cause a exception.

This PR fixed this issue.

### DIFF
--- a/ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/RangerServiceHawq.java
+++ b/ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/RangerServiceHawq.java
@@ -102,6 +102,13 @@ public class RangerServiceHawq extends RangerBaseService {
         return result;
     }
 
+    /**
+     * decrypt password field of configs
+     * Note:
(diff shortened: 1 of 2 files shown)


ranger-plugin/admin-plugin/src/main/java/org/apache/hawq/ranger/service/RangerServiceHawq.java
```diff
@@ -112,7 +119,7 @@ private void decryptPassword(Map<String, String> configs) throws Exception {
                 // when decrypt failed do nothing
                 LOG.warn("decrypt_password failed: " + e);
             }
-            configs.put("password", normal_password);
+            configs.put("password_jdbc", normal_password);
```
interma: put the decrypted password into a new field, so it doesn't influence the BaseClient decrypt it again.

interma:
@linwen @zhangh43 @stanlyxiang help to review again, thanks!

linwen:
LGTM

PR #1261, opened 2017-06-29
HAWQ-1490. Added new dummy plugin for offline pxf testing.

Quikling:
This plugin allows for pxf testing without any underlying file system or filestore.


pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
```diff
@@ -0,0 +1,46 @@
+package org.apache.hawq.pxf.api.examples;
```
sansanichfb: This will fail rat check, please add Apache license header.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+public class DummyAccessor extends Plugin implements ReadAccessor {
+    private static final Log LOG = LogFactory.getLog(DummyAccessor.class);
+    private int rowNumber;
+    private int fragmentNumber;
+    public DummyAccessor(InputData metaData) {
+        super(metaData);
+    }
+    @Override
+    public boolean openForRead() throws Exception {
+        /* fopen or similar */
```
sansanichfb: What does this comment mean?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
```diff
@@ -0,0 +1,29 @@
...
+import org.apache.hawq.pxf.api.ReadResolver;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.apache.hawq.pxf.api.utilities.Plugin;
+import java.util.LinkedList;
+import java.util.List;
+import static org.apache.hawq.pxf.api.io.DataType.INTEGER;
+import static org.apache.hawq.pxf.api.io.DataType.VARCHAR;
+
+public class DummyResolver extends Plugin implements ReadResolver {
+    private int rowNumber;
```
sansanichfb: Why is this member needed?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
```diff
@@ -0,0 +1,29 @@
...
+    public DummyResolver(InputData metaData) {
+        super(metaData);
+        rowNumber = 0;
+    }
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
+        /* break up the row into fields */
+        List<OneField> output = new LinkedList<OneField>();
+        String[] fields = ((String) row.getData()).split(",");
```
sansanichfb: I would add validations here, whether a row is not null, whether it has data, and if it has two tokens separated by a comma.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
```diff
@@ -0,0 +1,39 @@
+package org.apache.hawq.pxf.api.examples;
+
+import org.apache.hawq.pxf.api.Fragmenter;
+import org.apache.hawq.pxf.api.Fragment;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import java.util.List;
+
+/**
+ * Fragmenter class for text data on local filesystem.
```
sansanichfb: According to logic, it's not really a local filesystem fragmenter, so please update comments.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
```diff
@@ -0,0 +1,39 @@
...
+ * Given an text file on local filesystem(a file, directory, or wild card pattern),divide
+ * the data into fragments and return a list of them along with a list of
+ * host:port locations for each.
+ */
+public class DummyFragmenter extends Fragmenter{
+    public DummyFragmenter(InputData metaData) {
+        super(metaData);
+    }
+    /*
+     * path is a data source URI that can appear as a file name, a directory name or a wildcard
```
sansanichfb: Please follow Javadoc format of comments.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
```diff
@@ -0,0 +1,39 @@
...
+     * returns the data fragments - identifiers of data and a list of available hosts
+     */
+    @Override
+    public List<Fragment> getFragments() throws Exception {
+        String localhostname = java.net.InetAddress.getLocalHost().getHostName();
+        String[] localHosts = new String[]{localhostname, localhostname};
+        fragments.add(new Fragment(inputData.getDataSource() + ".1" /* source name */,
+                localHosts /* available hosts list */,
+                "fragment1".getBytes()));
```
sansanichfb: /* fragment's metadata */

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
```diff
@@ -0,0 +1,29 @@
...
+import static org.apache.hawq.pxf.api.io.DataType.VARCHAR;
+
+public class DummyResolver extends Plugin implements ReadResolver {
+    private int rowNumber;
+    public DummyResolver(InputData metaData) {
+        super(metaData);
+        rowNumber = 0;
+    }
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
```
sansanichfb: This resolver returns static schema with two columns, let's return dynamic schema, which comes as a BridgeApi parameters. Please refer to InputData#getColumn(int index) and InputData#getColumns() method to get requested column's metadata.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
Quikling: Thanks, will add.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
Quikling: It's an example of what function might go there. Will update the comment to reflect what the function currently does.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
Quikling: will remove

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
Quikling: row can never be null here as the ReadBridge specifically check for null. None of the other resolvers have this check as well. Will add validation for other things you mentioned

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
denalex: I personally don't find the style of including comments inside function call signature all that useful.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+        return true;
+    }
+    @Override
+    public OneRow readNextObject() throws Exception {
+        /* return next row , <key=fragmentNo.rowNo, val=rowNo,text,fragmentNo>*/
```
denalex: not sure whether this is relevant here, but can we produce "X,Hello World X" string where X is a segment id ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
Quikling: Thanks, will update.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
```diff
@@ -0,0 +1,46 @@
+package org.apache.hawq.pxf.api.examples;
+
+import org.apache.hawq.pxf.api.OneRow;
+import org.apache.hawq.pxf.api.ReadAccessor;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.apache.hawq.pxf.api.utilities.Plugin;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+public class DummyAccessor extends Plugin implements ReadAccessor {
```
Quikling: Thanks! changed to "Static"

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
Quikling: Thanks, removed.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyAccessor.java
Quikling: As long as you define the schema with text as the data type for each field, you will still see text data. So shouldn't matter

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyFragmenter.java
Quikling: Will update.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DummyResolver.java
Quikling: At this stage, we're just trying to test simple demo functionality. We can introduce dynamic types later.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DemoResolver.java
```diff
@@ -0,0 +1,59 @@
...
+        super(metaData);
+    }
+
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
+        List<OneField> output = new LinkedList<OneField>();
+        Object data = row.getData();
+
+        /* break up the row into fields */
+        String[] fields = ((String) data).split(",");
```
sansanichfb: Why are we not handling a case when data doesn't have 3 tokens? It would be useful for a demo profile to show how to gracefully handle errors for real profile.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DemoTextResolver.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param metaData
+     */
+    public DemoTextResolver(InputData metaData) {
+        super(metaData);
+    }
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
+        List<OneField> output = new LinkedList<OneField>();
+        Object data = row.getData();
+        output.add(new OneField(VARCHAR.getOID(), data));
```
sansanichfb: Instead of adding one field we can just call inputData.getColumns() and add that many fields to output how many was requested. In this way, it will be dynamic and work with any table definition.

denalex:
Do we also need to add pxf-api.jar to the private classpath so that the Demo* use case is available to users without modifications ?

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DemoTextResolver.java
shivzone: Text resolver only sends data as one field.

pxf/pxf-api/src/main/java/org/apache/hawq/pxf/api/examples/DemoResolver.java
shivzone: Yes, it has now changed to make it flexible

shivzone:
@denalex i've added pxf-api jar to the pxf-private.classpath file. I haven't modified the hdp or the bigtop versions of the classpath file

pxf/pxf-api/src/test/java/org/apache/hawq/pxf/api/DemoAccessorTest.java
```diff
@@ -0,0 +1,79 @@
...
+import org.apache.hawq.pxf.api.examples.DemoAccessor;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import static org.junit.Assert.*;
```
kdunn926: IMO it would be better to explicitly identify each import, rather than a glob import.

kdunn926:
LGTM 👍

PR #1265, opened 2017-07-11
HAWQ-1500. HAWQ-1501. HAWQ-1502. Support TDE write function.

amyrazz44:


amyrazz44:
@linwen @interma  Please feel free to review this pr. Thank you.

depends/libhdfs3/src/client/HttpClient.cpp
```diff
@@ -0,0 +1,337 @@
...
+        	        "Cannot initialize headers in HttpClient: %s: %s");
+	} else {
+		LOG(DEBUG1, "HttpClient : Http Header is NULL");
+	}
+
+	if (curl != NULL) {
+		CURL_SETOPT_ERROR2(curl, CURLOPT_URL, url.c_str(),
+    	        "Cannot initialize url in HttpClient: %s: %s");
+	} else {
+		LOG(DEBUG1, "HttpClient : Http URL is NULL");
```
wengyanqing: It should be error if URL is NULL

depends/libhdfs3/src/client/HttpClient.cpp
```diff
@@ -0,0 +1,337 @@
...
+	}
+	LOG(DEBUG1, "HttpClient : The http method is %d. The http url is %s. The http response is %s.", method, url.c_str(), response.c_str());
+	return response;
+}
+
+/**
+ * Http GET method.
+ */
+std::string HttpClient::get() {
+	httpMethod method = HTTP_GET;
```
wengyanqing: temp variable "method" is redundant ？

linwen:
Please unify the indent. We should avoid use both "space" and "tab" for indent in one source file.

depends/libhdfs3/CMake/FindSSL.cmake
```diff
@@ -0,0 +1,26 @@
...
+
+FIND_LIBRARY(SSL_LIBRARIES crypto)
+
+INCLUDE(FindPackageHandleStandardArgs)
+
+# handle the QUIETLY and REQUIRED arguments and set SSL_FOUND to TRUE if
+# all listed variables are TRUE
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(SSL DEFAULT_MSG SSL_LIBRARIES SSL_INCLUDE_DIR)
+
+MARK_AS_ADVANCED(SSL_INCLUDE_DIR SSL_LIBRARIES)
```
paul-guo-: Can we leverage the code for ssl and curl in configure?
By the way, now we should auto-enable with-openssl if with-libhdfs3 is enabled.

depends/libhdfs3/mock/MockHttpClient.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "gmock/gmock.h"
+
+#include "client/HttpClient.h"
+#include "client/KmsClientProvider.h"
+#include <boost/property_tree/ptree.hpp>
+
+using boost::property_tree::ptree;
+
+class MockHttpClient: public Hdfs::HttpClient {
```
interma: I think we should add a unit test for HttpClient, because libcurl has many tricky behaviors, maybe we forgot some exceptional path. 
Just a reminder, let's do it in future.

depends/libhdfs3/src/client/HttpClient.h
```diff
@@ -0,0 +1,155 @@
...
+    HTTP_GET = 0,
+	HTTP_POST = 1,
+	HTTP_DELETE = 2,
+	HTTP_PUT = 3
+} httpMethod;
+
+namespace Hdfs {
+
+class HttpClient {
+public:
```
interma: I think we should add a unit test for HttpClient, because libcurl have many tricky behaviors, maybe we forgot some exceptional path.
Just a reminder, let's do it in future.

depends/libhdfs3/src/client/KmsClientProvider.cpp
```diff
@@ -0,0 +1,318 @@
...
+std::string	KmsClientProvider::base64Encode(const std::string &data)
+{
+	char * buffer = NULL;
+	size_t len = 0;
+	int rc = 0;
+	std::string result;
+
+	LOG(DEBUG1, "KmsClientProvider : Encode data is %s", data.c_str());
+
+	if (GSASL_OK != (rc = gsasl_base64_to(data.c_str(), data.size(), &buffer, &len))) {
```
interma: Must use `data.data()` instead of `data.c_str()`, please check whether the similar issues existed in other place.

depends/libhdfs3/src/client/CryptoCodec.cpp
```diff
@@ -0,0 +1,163 @@
...
+		EVP_CIPHER_CTX_free(encryptCtx);
+}
+
+/**
+ * Get decrypted key from kms.
+ */
+std::string CryptoCodec::getDecryptedKeyFromKms()
+{
+	ptree map = kcp->decryptEncryptedKey(*encryptionInfo);
+	std::string key = map.get<std::string>("material");
```
interma: need a try catch here? when map doesn't has a material field.

interma: need a try catch here? when map doesn't has a material field.

depends/libhdfs3/src/client/CryptoCodec.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	int rem = key.length() % 4;
+	if (rem) {
+		rem = 4 - rem;
+		while (rem != 0) {
+			key = key + "=";
+			rem--;
+		}
+	}
```
interma: maybe we should refer the hadoop-kms code to find the answer...

depends/libhdfs3/src/client/HttpClient.h
stanlyxiang: If we have something to do in the future, I think a jira should be filed.

depends/libhdfs3/src/client/CryptoCodec.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+	std::replace(key.begin(), key.end(), '-', '+');
+	std::replace(key.begin(), key.end(), '_', '/');
+
+	LOG(INFO, "material is :%s", key.c_str());
```
linwen: Suggest provide more clear log message, and if this function is called frequently, use DEBUG3 instead of INFO.

depends/libhdfs3/src/client/CryptoCodec.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+namespace Hdfs {
+
+/**
+ * Construct a CryptoCodec instance.
+ * @param encryptionInfo the encryption info of file.
+ * @param kcp a KmsClientProvider instance to get key from kms server.
+ * @param bufSize crypto buffer size.
+ */
+CryptoCodec::CryptoCodec(FileEncryptionInfo *encryptionInfo, std::shared_ptr<KmsClientProvider> kcp, int32_t bufSize) : encryptionInfo(encryptionInfo), kcp(kcp), bufSize(bufSize)
```
interma: Should use `shared_ptr` instead of `std::shared_ptr`, because libhdfs have two implements of shared_ptr. 
See: depends/libhdfs3/src/common/Memory.h

amyrazz44:
Only support TDE write function on single data node. Will support multiple data nodes in the next pull request.

depends/libhdfs3/src/client/OutputStreamImpl.cpp
```diff
@@ -278,7 +316,10 @@ void OutputStreamImpl::append(const char * buf, int64_t size) {
 
 void OutputStreamImpl::appendInternal(const char * buf, int64_t size) {
     int64_t todo = size;
+    if (fileStatus.isFileEncrypted()) {
+        buf = cryptoCodec->encode(buf, size).c_str();
```
interma: encode() reture a **temporary** string object, so buf point a uncertain memory.
MUST fix it.

wengyanqing:
Great job. LGTM.

PR #1269, opened 2017-07-25
HAWQ-1506. Support multi-append a file within encryption zone

interma:
This bug is happened on: 
1, append a existed file. 
2, call multi `hdfsWrite()`

Test case please see `TEST(TestCInterfaceTDE, TestAppendMultiTimes)`.

### DIFF
--- a/depends/libhdfs3/src/client/CryptoCodec.cpp
+++ b/depends/libhdfs3/src/client/CryptoCodec.cpp
@@ -106,14 +106,7 @@ namespace Hdfs {
 			return 0;
 
 		// Get decrypted key from KMS
-		std::string key;
-		try {
-			key = getDecryptedKeyFromKms();
-		}
-		catch (...) {
-			LOG(WARNING, "getDecryptedKeyFromKms failed");
-			return -1;
-		}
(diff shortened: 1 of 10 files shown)


interma:
@linwen @wengyanqing @amyrazz44 @wcl14 help to review, thanks!

depends/libhdfs3/src/client/CryptoCodec.cpp
```diff
@@ -25,154 +25,188 @@
...
+		if (is_init)
+			return 0;
+
+		// Get decrypted key from KMS
+		std::string key;
+		try {
+			key = getDecryptedKeyFromKms();
+		}
+		catch (...) {
+			LOG(WARNING, "getDecryptedKeyFromKms failed");
```
wengyanqing: It's better to throw the exception to upper function then return error.

interma: You are right, the exception should throw to upper caller to judge how to deal with it.
Thanks!

wengyanqing:
LGTM.

radarwave:
+1

interma:
merged

PR #1273, opened 2017-08-02
HAWQ-1502. Add verification to support TDE write function.

amyrazz44:


depends/libhdfs3/test/function/TestCInterface.cpp
```diff
@@ -369,29 +448,154 @@ TEST(TestCInterfaceTDE, TestAppendWithTDELargeFiles_Success) {
...
+    char bufGets[128];
+    while (fgets(bufGets, sizeof(bufGets), file)) {
+    }
+    pclose(file);
+    //Check the buffer's md5 value is eaqual to the tdefile's md5 value.
     system("rm -rf ./testfile");
-    system("hadoop fs -get /TDE/testfile ./");
-    diff_file2buffer("testfile", &buffer[0]);
+    system("hadoop fs -get /TDEAppend3/testfile ./");
+    char resultFile[33] = { 0 };
```
linwen: Since 33 is the size of MD5(32 characters) + 1, so I suggest define a const variable or macro for it. So that we can avoid use such strange number in this source file.

linwen:
+1

PR #1278, opened 2017-08-10
HAWQ-1498. Segments keep open file descriptors for deleted files

jiny2:
The idea is to hold a list of all connections ever used for dropping objects in HDFS, then when the transaction ends, no matter commit or rollback, those connections are forced to close. Those connections cached but never used for dropping, are not touched.

### DIFF
--- a/src/backend/storage/file/fd.c
+++ b/src/backend/storage/file/fd.c
@@ -228,6 +228,7 @@ typedef struct
  * hash table of hdfs file systems, key = hdfs:/<host>:<port>, value = hdfsFS
  */
 static HTAB * HdfsFsTable = NULL;
+static HTAB * HdfsFsTable4Drop = NULL;
 static MemoryContext HdfsGlobalContext = NULL;
 #define EXPECTED_MAX_HDFS_CONNECTIONS 10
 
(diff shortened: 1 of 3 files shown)


jiny2:
I manually tested this fix in my environment, please review and comment.

@linwen @radarwave

src/backend/storage/file/fd.c
```diff
@@ -2403,8 +2479,22 @@ HdfsGetConnection(const char * path)
...
+		 * check the table for dropping. */
+		if (!isForDrop) {
+			entry = (struct FsEntry *) hash_search(HdfsFsTable,
+												   location,
+												   HASH_ENTER,
+												   &found);
+		}
+		else
+		{
+			elog(LOG, "search 4 drop 1");
```
linwen: Is it a necessary log or better to be changed more meaningful?

src/backend/storage/file/fd.c
```diff
@@ -2432,7 +2525,12 @@ HdfsGetConnection(const char * path)
 				{
 					if (!login())
 					{
-						hash_search(HdfsFsTable, location, HASH_REMOVE, &found);
+						if (!isForDrop)
+							hash_search(HdfsFsTable, location, HASH_REMOVE, &found);
+						else {
+							elog(LOG, "search 4 drop 2");
```
linwen: Same comment as last one, this log message is more like for debugging.

linwen:
Please see the comments.  
The change LGTM. Thanks!

PR #1280, opened 2017-08-14
HAWQ-1510. Add TDE-related functionality into hawq init command

interma:


interma:
@linwen @wengyanqing @amyrazz44 help to review, thanks.

Note:
Only include necessary code of **hawq-init**, other command modification will be added in another PR.

src/bin/gpcheckhdfs/gpcheckhdfs.c
```diff
@@ -110,6 +114,12 @@ int main(int argc, char * argv[]) {
         krb_keytabfile = argv[5];
     }
 
+    //get tde key name param
+    for (int i = 1; i < argc; i++) {
+        if (strcmp(argv[i], "--with-tde") == 0 && i+1 < argc)
+            tde_keyname = argv[i+1];
```
interma: why not use `argv[6]` as the `tde_keyname`?
because the params before it may be empty...

amyrazz44: Add some comments in code about this will be better.

amyrazz44:
LGTM +1

src/bin/gpcheckhdfs/gpcheckhdfs.c
interma: thanks, I will fix it.

wengyanqing:
LGTM

src/bin/gpcheckhdfs/gpcheckhdfs.c
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the
  * "License"); you may not use this file except in compliance
  * with the License.  You may obtain a copy of the License at
- * 
+ *
  *   http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
```
stanlyxiang: just a reminder, we should not update license header unless there is a rat-check failure or necessary changes.

interma: thanks!

PR #1282, opened 2017-08-29
HAWQ-1520. gpcheckhdfs should skip hdfs trash directory

interma:
When enable **hdfs trash** feature, there is a *Trash* directory under the encryption zone. 
So gpcheckhdfs should still consider the hawq directory contains *Trash* directory as an empty directory.

more details in JIRA:
https://issues.apache.org/jira/browse/HAWQ-1520

### DIFF
--- a/src/bin/gpcheckhdfs/gpcheckhdfs.c
+++ b/src/bin/gpcheckhdfs/gpcheckhdfs.c
@@ -57,6 +57,8 @@
 #define KRBLOGIN_ERR 107
 #define DIRECTORY_ERR 108
 
+#define TRASH_DIRECTORY_NAME ".Trash"
+
 char * conncat(const char * dfs_name, const char * dfs_url);
 void getHostAndPort(const char * dfs_url, char * host, char * port);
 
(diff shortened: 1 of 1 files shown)


interma:
@radarwave @linwen @wcl14 help to review, thanks.

src/bin/gpcheckhdfs/gpcheckhdfs.c
```diff
@@ -271,6 +273,21 @@ int testHdfsConnect(hdfsFS * fsptr, const char * host, int iPort,
     return 0;
 }
 
+/*
+ * check path is a trash directory
+ * path, e.g: /hawq_default/.Trash
+ */
+static int is_trash_directory(const char *path) {
```
linwen: I suggest the name of this function uses "isTrashDirectory" instead of current one, since the other functions name follow this style: testFooBar, testXxxYyy, etc.

interma: thanks, I will fix it later.

radarwave:
+1

interma:
merged

PR #1283, opened 2017-08-29
HAWQ-1523. Provide ability to bundle PXF artifacts into a single tarball

lavjain:


sansanichfb:
@lavjain will layout be the same as for install target?

shivzone:
@sansanichfb the layout would be exactly the same as make install target

PR #1286, opened 2017-09-12
HAWQ-1525. Segmentation fault occurs if reindex database when loading…

wcl14:
… data from Hive to HAWQ using hcatalog


wengyanqing:
LGTM

PR #1289, opened 2017-09-21
HAWQ-1526. Added support of functions in WHERE clause for a PPD.

sansanichfb:


src/backend/access/external/pxffilters.c
```diff
@@ -851,12 +852,51 @@ append_attr_from_var(Var* var, List* attrs)
...
+/*
+ * append_attr_from_func_args
+ *
+ * extracts all columns from FuncExpr into attrs
+ * assigns false to expressionIsSupported if at least one of items is not supported
+ */
+static List*
+append_attr_from_func_args(FuncExpr *expr, List* attrs, bool* expressionIsSupported) {
+	ListCell *lc = NULL;
+	if (!expressionIsSupported) {
```
denalex: this can be the first check, so that you don't have to allocate lc variable if expression is not supported

src/backend/access/external/pxffilters.c
```diff
@@ -851,12 +852,51 @@ append_attr_from_var(Var* var, List* attrs)
...
+
+/*
+ * get_attrs_from_expr
+ *
+ * extracts and returns list of all columns from Expr
+ * assigns false to expressionIsSupported if at least one of items is not supported
+ */
 static List*
-get_attrs_from_expr(Expr *expr)
+get_attrs_from_expr(Expr *expr, bool* expressionIsSupported)
```
denalex: an alternative way is to have all these functions return a boolean, indicating whether selection can be made or not  and take a list as an OUT function argument. This feels more intuitive, IMO.

src/backend/access/external/pxfheaders.c
```diff
@@ -65,14 +65,18 @@ void build_http_header(PxfInputData *input)
 	
 	if (proj_info != NULL && proj_info->pi_isVarList)
 	{
-		List* qualsAttributes = extractPxfAttributes(input->quals);
+		bool qualsAreSupported = true;
+		List* qualsAttributes = extractPxfAttributes(input->quals, &qualsAreSupported);
```
shivzone: qualsAreSupported seems redundant. Is there any scneario when qualsAttributes is not null when we don't want to support projection info ?

sansanichfb: qualsAttributes could be null when WHERE clause is empty, and in this case, we still want to use projection info. But if we don't support some operation in WHERE clause - we just report qualsAreSupported flag and projection info will not be used.

src/backend/access/external/pxffilters.c
shivzone: I agree

sansanichfb: Right, might be more intuitive to reverse signatures, but I am not sure if that refactoring will give many benefits other than the more intuitive perception of code, which also makes sense, but I would stick to the current signature at this point.

src/backend/access/external/pxffilters.c
sansanichfb: Sure, updated

PR #1290, opened 2017-09-25
HAWQ-1529. Fix segment resource manager hang when postmaster died

kuien:
If PostmasterIsAlive() is under implicit declaration, %eax (32-bits) will be used for comparison rather than %al (8-bits), BUT PostmasterIsAlive() only set the lower 8-bits (because 'bool' is really a 'char'). Then segment resource manager will never exit after postmaster died.


linwen:
LGTM, +1

jianlirong:
LGTM, +1

PR #1291, opened 2017-09-28
HAWQ-1531. Templetized PXF user and log directories.

sansanichfb:
PXF now could be build for different profiles(flavors) - HAWQ(by default) or GPDB


pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -247,13 +255,43 @@ function commandWebapp()
...
+    chown -R $instance_owner $PXF_RUNDIR
+    chmod 700 $PXF_RUNDIR
+    return 0
+}
+
+function generatePrivateClasspath()
+{
+    sed -e "s|PXF_HOME|$PXF_HOME|g" $PXF_HOME/conf-templates/pxf-private.classpath.template > $PXF_HOME/conf/pxf-private.classpath
+    sed -i -e "s|HADOOP_HOME|$hadoop_home|g" $PXF_HOME/conf/pxf-private.classpath
+    sed -i -e "s|HBASE_HOME|$hbase_home|g" $PXF_HOME/conf/pxf-private.classpath
```
denalex: what if user did not provide HBASE_HOME and / or HIVE_HOME ? Should we check the variables are set before substituting ?

pxf/pxf-service/src/main/resources/pxf-log4j.properties
```diff
@@ -23,7 +23,7 @@ log4j.rootLogger=INFO, ROLLINGFILE
 
 log4j.appender.ROLLINGFILE=org.apache.log4j.RollingFileAppender
 # use the same value directory as set for $PXF_LOGDIR
-log4j.appender.ROLLINGFILE.File=/var/log/pxf/pxf-service.log
+log4j.appender.ROLLINGFILE.File=$${env:$PXF_LOGDIR}/pxf-service.log
```
lavjain: Are two dollar symbols above intentional?

sansanichfb: Log4j 2.x documentation suggests this way of reading environment var - $${env:USER}. Still testing though.

denalex: I would assume these properties would be retrieved from Java in which case they would need to be set via -D option when starting Java ?

sansanichfb: According to docs, Log4J supports both environment and system vars, $${env:USER} and ${sys:logPath} respectively.

shivzone:
Please make sure you update the PXF Build instructions as well since the init command has changed.
https://cwiki.apache.org/confluence/display/HAWQ/PXF+Build+and+Install

PR #1294, opened 2017-10-03
HAWQ-1534. Change user to PXF_USER only for HAWQ database.

sansanichfb:


pxf/pxf-service/src/scripts/pxf-env.sh
```diff
@@ -34,7 +34,11 @@ export PXF_LOGDIR=@pxfLogDir@
 export PXF_RUNDIR=@pxfRunDir@
 
 # Configured user
-export PXF_USER=${PXF_USER:-@pxfDefaultUser@}
+if [ ! -z "@pxfDefaultUser@" ]; then
```
denalex: these conditions will look really weird in the generated script, would it be better to assign this to variable, which gets filtered value and then check the value of that variable in all the if statements.

sansanichfb:
@denalex introduced variable allow_custom_pxf_user

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -341,6 +358,14 @@ function validateParameters()
         exit 1
     fi
 
+    # make sure current user is not root
+    if [ $EUID -eq 0 ]; then
+        echo "ERROR: Cannot run as root user"
+        exit 1
+    fi
+
```
lavjain: Minor - extra spaces

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -72,9 +77,11 @@ function createInstance()
...
 	fi
 
-	chown -R $instance_owner $instance_root
-	chmod 700 $instance_root/$instance_name
-	return 0
+    if $allow_custom_pxf_user; then
+        chown -R $instance_owner $instance_root
+    fi
+    chmod 700 $instance_root/$instance_name
+    return 0
```
lavjain: Minor - indentation

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -72,9 +77,11 @@ function createInstance()
 		return 1
 	fi
 
-	chown -R $instance_owner $instance_root
-	chmod 700 $instance_root/$instance_name
-	return 0
+    if $allow_custom_pxf_user; then
+        chown -R $instance_owner $instance_root
```
lavjain: Should we use chown -R $instance_owner:$instance_owner instead (to include group)?

sansanichfb: There is no guaranty that that group exists, so we don't have to set it here.

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -35,9 +35,14 @@ else
...
+pxf_user=${PXF_USER}
 instance_port=${PXF_PORT:-51200}
 instance_name=pxf-service
+if [ ! -z '@pxfDefaultUser@' ]; then
+    allow_custom_pxf_user=true
+else
+    allow_custom_pxf_user=false
+fi
 
 if [ -z $PXF_HOME ]; then
```
shivzone: We don't need to set instance_owner  when PXF_HOME is set

pxf/pxf-service/src/scripts/pxf-env.sh
```diff
@@ -33,8 +33,19 @@ export PXF_LOGDIR=@pxfLogDir@
...
+else
+    allow_custom_pxf_user=false
+fi
+
 # Configured user
-export PXF_USER=${PXF_USER:-@pxfDefaultUser@}
+if [ $allow_custom_pxf_user ]; then
+    export PXF_USER=${PXF_USER:-@pxfDefaultUser@}
+else
+    export PXF_USER=${USER}
```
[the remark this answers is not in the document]
sansanichfb: Thanks, updated

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -267,15 +280,19 @@ function commandWebapp()
 function createLogsDir()
 {
     mkdir -p $PXF_LOGDIR
-    chown -R $instance_owner $PXF_LOGDIR
+    if $allow_custom_pxf_user; then
+        chown -R $instance_owner $PXF_LOGDIR
+    fi
     chmod 700 $PXF_LOGDIR
```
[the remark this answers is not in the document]
sansanichfb: I would prefer to keep it restrictive and change to less restrictive if needed later.

pxf/pxf-service/src/scripts/pxf-service
sansanichfb: Thanks, updated

PR #1295, opened 2017-10-04
HAWQ-1535. PXF make install should also generate non versioned jars

shivzone:


pxf/pxf-service/src/configs/templates/pxf-private.classpath.template
```diff
@@ -23,16 +23,15 @@
...
+PXF_HOME/lib/pxf-hdfs.jar
+PXF_HOME/lib/pxf-hive.jar
+PXF_HOME/lib/pxf-json.jar
+PXF_HOME/lib/pxf-jdbc.jar
 
-# Hadoop/Hive/HBase configurations
+# Hadoop/Hive configurations
 HADOOP_HOME/etc/hadoop
 HIVE_HOME/conf
-HBASE_HOME/conf
```
sansanichfb: Why are we removing HBase entries?

PR #1298, opened 2017-10-09
HAWQ-1536. Fix PXF Json no seek unit test

shivzone:
Fixed PartitionedJsonParserNoSeekTest to not have any expectation in expected response order.


shivzone:
This solution will not be required due to https://github.com/apache/incubator-hawq/pull/1299

PR #1301, opened 2017-10-12
Re-added pxf profile default to rpm and tar tasks

shivzone:
The tar and the rpm targets were missing pxf-profiles-default.xml.

### DIFF
--- a/pxf/build.gradle
+++ b/pxf/build.gradle
@@ -245,6 +245,11 @@ project('pxf-service') {
             into "/etc/pxf-${project.version}/conf"
         }
 
+        from('src/main/resources/pxf-profiles-default.xml') {
+            fileType CONFIG | NOREPLACE
+            into "/etc/pxf-${project.version}/conf"
+        }
+
         from('src/configs/pxf-site.xml') {
             fileType CONFIG | NOREPLACE
             into "/etc/pxf-${project.version}/conf"
@@ -310,6 +315,7 @@ project('pxf-service') {
 
     project.distTar {
         from('src/main/resources/pxf-profiles.xml') { into 'conf' }
+        from('src/main/resources/pxf-profiles-default.xml') { into 'conf' }
         from("src/main/resources") { into 'conf' include '**/pxf-private*.classpath'}
         from("src/main/resources/pxf-private${hddist}.classpath") { into 'conf' rename {'pxf-private.classpath'} }
         from('src/main/resources/pxf-public.classpath') { into 'conf' }


pxf/build.gradle
```diff
@@ -310,6 +315,7 @@ project('pxf-service') {
 
     project.distTar {
         from('src/main/resources/pxf-profiles.xml') { into 'conf' }
+        from('src/main/resources/pxf-profiles-default.xml') { into 'conf' }
```
lavjain: This may not work because pxf-profiles-default.xml has to be copied to pxf-profiles.xml

shivzone: For the product's perspective why is that needed ?

lavjain: So, people would have to manually copy it then?

shivzone: Not really. The purpose of pxf-profile-default.xml is to have all the default profiles. The users need to only add custom profiels to pxf-profile.xml. The PXF application loads both the xml files during startup.

lavjain: Then why did it fail for pxf-automation on HAWQ?

shivzone: Automation test failure is a different issue and that needs to be addressed separately out of this PR. This would not be an end user issue.
The automation test assumes pxf-profile.xml has all the profiles in it which is not really the case anymore. The tests need to addressed separately.

lavjain: I still do not understand why pxf-profile.xml needs to be empty but that is a product question for @kongyew

lavjain: @sansanichfb just verified in the code that we are loading both profiles.

shivzone:
@lavjain pxf-profiles.xml is better to be empty to prevent users from incorrectly updating the default entires in it. This is similar to the separation we have between pxf-private.classpath from pxf-public.classpath.
@lavjain @sansanichfb  does that mean we don't need pxf-profiles.default ?

pxf/build.gradle
shivzone: @lavjain pxf-profiles.xml is better to be empty to prevent users from incorrectly updating the default entires in it. This is similar to the separation we have between pxf-private.classpath from pxf-public.classpath.
@lavjain @sansanichfb  does that mean we don't need pxf-profiles.default ?

lavjain:
If pxf-private-default is not needed, then we should remove it. The less number of conf files we have, the better.

shivzone:
After some investigation, the webapp does have pxf-profiles-default.xml file in pxf-service/webapps/pxf/WEB-INF/classes/ along with other configuration files. @lavjain is right, we dont' need pxf-profiles-default.xml in conf/.

PR #1306, opened 2017-11-02
HAWQ-1543. Make pxf configurable upon restart through pxf-env

shivzone:


pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -216,17 +174,16 @@ function doInit()
...
 }
 
 #
-# patchWebapp patches the webapp config files
-# patch applied only if PXF_HOME is defined
+# configureWebapp patches the webapp with pxf and user overriden configs
+# applied only if PXF_HOME is defined
 #
-function patchWebapp()
+function configureWebapp()
```
denalex: since this function is now called on every start, we should not be unpacking WAR file and edit web.xml -- only catalina files that use dynamic properties should be regenerated. Ideally, they should be regenerated from templates that do not get destroyed, so that customers have a chance to change the template, if needed.

pxf/pxf-service/src/scripts/pxf-service
```diff
@@ -249,6 +206,34 @@ function patchWebapp()
...
     mv web.xml.tmp $web_file
+
+    # set port
+    catalinaProperties=$instance_root/$instance_name/conf/catalina.properties
+    cat $catalinaProperties | \
+    sed "s|^[[:blank:]]*connector.http.port=.*$|connector.http.port=$instance_port|g" \
+    > ${catalinaProperties}.tmp
+
+    rm $catalinaProperties
+    mv ${catalinaProperties}.tmp $catalinaProperties
```
denalex: can we just copy from a template and "sed" in place instead of all these file manipulations ?

pxf/pxf-service/src/scripts/pxf-env.sh
```diff
@@ -41,6 +41,9 @@ fi
 # Port
 export PXF_PORT=${PXF_PORT:-51200}
 
+# Memory
+export PXF_HEAP_OPTS="-Xmx2g -Xms1g"
```
denalex: I would call these PXF_JVM_OPTS since there can be other options here (like GC setting, etc)

shivzone: will change

pxf/pxf-service/src/scripts/pxf-service
shivzone: The purpose of patching the war is so that an update done to PXF as part of upgrades wouldn't require the user to run pxf init again. They can simply use pxf restart.
Can evaluate that change outside of this PR

pxf/pxf-service/src/scripts/pxf-service
lavjain: sed -i would be preferred as Alex suggested.

PR #1307, opened 2017-11-06
HAWQ-1544. prompt file count doesn't match hash bucket number when re…

interma:
…organize table

### DIFF
--- a/src/backend/cdb/cdbcat.c
+++ b/src/backend/cdb/cdbcat.c
@@ -296,10 +296,10 @@ GpPolicyStore(Oid tbloid, const GpPolicy *policy)
 /*
  * Sets the policy of a table into the gp_distribution_policy table
  * from a GpPolicy structure.
- *
+ * @param update_bucketnum, whether update the bucketnum field
  */
-void
-GpPolicyReplace(Oid tbloid, const GpPolicy *policy)
+static void
+GpPolicyReplace_inner(Oid tbloid, const GpPolicy *policy, bool update_bucketnum)
 {
 	Relation	gp_policy_rel;
 	HeapTuple	gp_policy_tuple = NULL;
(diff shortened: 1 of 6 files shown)


interma:
How to reproduct this issue, see jira: https://issues.apache.org/jira/browse/HAWQ-1544

Root cause:
When create a table, it inserts a row into `gp_distribution_policy ( localoid | bucketnum | attrnums )`, and it will update this row on the follow reorganize table. 
But it only updates attrnums field, the correct logic also needs updating bucketnum field to the current default_hash_table_bucket_number.

interma:
have already added feature-test.

@linwen @amyrazz44 @wengyanqing @radarwave help to review, thanks.

src/backend/cdb/cdbcat.c
```diff
@@ -534,3 +536,15 @@ checkPolicyForUniqueIndex(Relation rel, AttrNumber *indattr, int nidxatts,
...
 	}
 }
+
+void 
+GpPolicyReplace(Oid tbloid, const GpPolicy *policy)
+{
+	GpPolicyReplace_inner(tbloid, policy, false);
+}
+void 
+GpPolicyReplaceWithBucketNum(Oid tbloid, const GpPolicy *policy)
```
interma: The older `GpPolicyReplace` function doesn't update `bucketnum` field though I have already set this field in `policy` structure, because `repl[1] = false;`
So I add a new GpPolicyReplaceWithBucketNum function to do it.

src/backend/cdb/cdbcat.c
```diff
@@ -296,7 +296,7 @@ GpPolicyStore(Oid tbloid, const GpPolicy *policy)
 /*
  * Sets the policy of a table into the gp_distribution_policy table
  * from a GpPolicy structure.
- *
+ * @param update_bucketnum, whether update the bucketnum field
```
linwen: A line of redundant comment. Should be removed.

interma:
thanks, all fixed.

radarwave:
+1

interma:
merged

PR #1308, opened 2017-11-07
HAWQ-1530. Illegally killing a JDBC select query causes locking problems

jiny2:
This bug root cause maybe twice FATAL error throwing, I cannot exactly reproduce it according to this issue's comment, just based on my understanding and previous bug fix experiences. We cannot allow a FATAL throwed again when anohter FATAL causes process into exit phase.

### DIFF
--- a/src/backend/cdb/dispatcher.c
+++ b/src/backend/cdb/dispatcher.c
@@ -59,6 +59,7 @@
 #include "tcop/pquery.h"			/* PortalGetResource */
 
 #include "resourcemanager/communication/rmcomm_QD2RM.h"
+#include "storage/ipc.h"
 
 /* Define and structure */
 typedef struct DispatchTask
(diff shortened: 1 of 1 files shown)


jiny2:
@kuien please check this fix

jiny2:
@radarwave @linwen @changleicn please review and comment

radarwave:
+1

amyrazz44:
LGTM +1

src/backend/cdb/dispatcher.c
```diff
@@ -1414,9 +1415,12 @@ dispatch_cleanup(DispatchData *data)
 	if (dispatcher_is_state_error(data))
```
kuien: May we check the status of 'proc_exit_inprogress' earlier and simpler? That is:
```
if (proc_exit_inprogress)
    return;
```

jiny2: The same logic, no special performance improvement in my opinion

PR #1309, opened 2017-11-08
HAWQ-1545. Fix pxf-private.classpath file for Cloudera tarball distro

lavjain:


shivzone:
I would definitely add this specific version dependancy on jackson library to release notes.

PR #1316, opened 2017-11-29
HAWQ-1550. Constrain max log file storage/duration

kuien:
A) add guc 'log_max_size' to constrain max log storage
B) add guc 'log_max_age' to constraint log size

### DIFF
--- a/src/backend/postmaster/syslogger.c
+++ b/src/backend/postmaster/syslogger.c
@@ -24,12 +24,15 @@
  */
 #include "postgres.h"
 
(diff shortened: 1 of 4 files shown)


interma:
Thanks @kuien 
I think we should delegate log rotation works to some outer tools, and the behavior is the same as GPDB/Posgres.

kuien:
make sense

interma:
Thanks @kuien, welcome to propose more PRs.

PR #1319, opened 2017-12-04
HAWQ-1563. Adapt to run Makefile in /bin/sh

larham:
- Since each line of Makefile runs in its own shell, pushd and popd are not necessary
- Also, pushd is not supported in /bin/sh, default shell in Ubuntu


denalex:
Committed: https://github.com/apache/incubator-hawq/commit/9578ab04c80d0d0cc1d93e8a8cde98f9e14b6e7d, closing PR

larham:
Alex pushed, thanks!

PR #1320, opened 2017-12-05
HAWQ-1564. Add Pluggable Storage Dependent Information

chiyang10000:
In this commit, I have add some dependent information needed by the following pluggable storage. 
The info added are mainly about external URI, block location, file splits and formatter action. 

The following two commits on including the pluggable storage in the executor, which rely on these information have already been prepared.

### DIFF
--- a/src/backend/access/external/url.c
+++ b/src/backend/access/external/url.c
@@ -24,6 +24,7 @@
 #include <sys/wait.h>
 
 #include "access/fileam.h"
+#include "access/filesplit.h"
 #include "access/heapam.h"
 #include "access/valid.h"
 #include "catalog/pg_extprotocol.h"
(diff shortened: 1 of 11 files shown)


src/backend/access/external/url.c
```diff
@@ -2318,7 +2321,24 @@ InvokeExtProtocol(void 	   		*ptr,
 	extprotocol->prot_scanquals = file->u.custom.scanquals;
 	extprotocol->prot_last_call = last_call;
 	extprotocol->desc = desc;
+	extprotocol->splits = NULL;
 	
+	if (*psplits != NULL) {
```
huor: check psplits itself to make sure it is not null here

src/include/access/relscan.h
```diff
@@ -127,6 +132,16 @@ typedef struct FileScanDescData
 	/* custom data formatter */
 	FormatterData *fs_formatter;
 	
+	/* formatter type and name */
```
huor: the indent of declaration need to be consistent with existing code

chiyang10000: I have updated my code due to your comments. thx!

PR #1321, opened 2017-12-08
HAWQ-1565. Include Pluggable Storage Format Framework in External Table Scan

chiyang10000:
Rewrite the tuple construct and consume work flow to improve the read and write performance in pluggable storage framework. To be specific, it uses virtual tuple in tuple table slot instead of heap tuple between storage and executor to avoid tuple data copy during query processing.

@huor @jiny2

### DIFF
--- a/src/backend/executor/nodeExternalscan.c
+++ b/src/backend/executor/nodeExternalscan.c
@@ -34,9 +34,12 @@
  *		ExecExternalReScan				rescans the relation
  */
 #include "postgres.h"
+#include "fmgr.h"
 
 #include "access/fileam.h"
+#include "access/filesplit.h"
 #include "access/heapam.h"
+#include "access/plugstorage.h"
 #include "cdb/cdbvars.h"
 #include "executor/execdebug.h"
 #include "executor/nodeExternalscan.h"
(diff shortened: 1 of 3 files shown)


huor:
Looks great to have performance gain in pluggable storage framework. +1

Is there any evaluation regarding the performance boost comparing with existing external table framework?

chiyang10000:
There will be a contrast performance test in the near days.

src/backend/executor/nodeExternalscan.c
```diff
@@ -80,8 +83,61 @@ ExternalNext(ExternalScanState *node)
...
+		FmgrInfo *getnextInitFunc = scandesc->fs_ps_scan_funcs.getnext_init;
+
+		if (getnextInitFunc)
+		{
+			/*
+			 * pg_strncasecmp(scandesc->fs_formatter_name, "orc", strlen("orc"))
+			 * Performance improvement for string comparison.
+			 */
+			const char *formatter_name = "orc";
+			if (*(int *)(scandesc->fs_formatter_name) != *(int *)formatter_name)
```
interma: I think pg_strncasecmp() or other some string compare function is easier to understand.
Just a advise, do not change is also ok.

chiyang10000: Thanks for your advise. I will update this PR with a simple demo in the near days.

PR #1322, opened 2017-12-12
Support Hive OpenCSVSerde

shivzone:


pxf/pxf-hive/src/main/java/org/apache/hawq/pxf/plugins/hive/HiveLineBreakAccessor.java
```diff
@@ -43,7 +43,7 @@
     public HiveLineBreakAccessor(InputData input) throws Exception {
         super(input, new TextInputFormat());
         ((TextInputFormat) inputFormat).configure(jobConf);
-        HiveUserData hiveUserData = HiveUtilities.parseHiveUserData(input, PXF_HIVE_SERDES.LAZY_SIMPLE_SERDE);
+        HiveUserData hiveUserData = HiveUtilities.parseHiveUserData(input, PXF_HIVE_SERDES.LAZY_SIMPLE_SERDE, PXF_HIVE_SERDES.CSV_SERDE);
```
denalex: why do we need to keep a list of supported Serde ? If Serde is transparent to us, let support all of them and not restrict the users to use specific ones.

lavjain: I second that. We should support all serdes (Hive is responsible for them)

shivzone:
@sansanichfb based on team's suggestion, I've removed the explicit serde check from HiveLineBreakAccessor as it supports multiple custom serdes.
We can eventually do away with the strict serde class check upon further investigation with the other Accessors.

PR #1326, opened 2017-12-29
HAWQ-1575. Implemented readable Parquet profile for PXF.

sansanichfb:


pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -184,4 +184,13 @@ under the License.
             <resolver>org.apache.hawq.pxf.plugins.jdbc.JdbcReadResolver</resolver>
         </plugins>
     </profile>
+    <profile>
+        <name>Parquet</name>
```
lisakowen: if someone could create another connector/profile that would access parquet files from a different external data store, we should consider qualifying the profile name, i.e. HdfsParquet.  (along the same lines, perhaps we should qualify the profiles named Avro and Json.)

sansanichfb: The storage layer is determined in fragmenter and accessor, resolver should remain the same. I would leave it as Parquet and later reconsider storage vs format naming conventions for all profiles.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/utilities/HdfsUtilities.java
```diff
@@ -151,18 +153,42 @@ public static boolean isThreadSafe(String dataDir, String compCodec) {
...
+    private static ByteArrayOutputStream writeBaseFragmentInfo(long start, long length, String[] locations) throws IOException {
+        ByteArrayOutputStream byteArrayStream = new ByteArrayOutputStream();
+        ObjectOutputStream objectStream = new ObjectOutputStream(byteArrayStream);
+        objectStream.writeLong(start);
+        objectStream.writeLong(length);
+        objectStream.writeObject(locations);
+        return byteArrayStream;
+    }
+
+    public static byte[] prepareFragmentMetadata(long start,
```
shivzone: I do'nt see this function used anywhere ?

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetDataFragmenter.java
```diff
@@ -0,0 +1,103 @@
...
+            String filepath = fsp.getPath().toUri().getPath();
+            String[] hosts = fsp.getLocations();
+
+            Path file = new Path(filepath);
+
+            ParquetMetadata metadata = ParquetFileReader.readFooter(
+                    job.getConfiguration(), file, ParquetMetadataConverter.NO_FILTER);
+            MessageType schema = metadata.getFileMetaData().getSchema();
+
+            byte[] fragmentMetadata = HdfsUtilities.prepareFragmentMetadata(fsp.getStart(), fsp.getLength(), fsp.getLocations());
```
shivzone: Can we simply use prepareFragmentMetadata(fsp)

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/utilities/HdfsUtilities.java
```diff
@@ -151,18 +153,42 @@ public static boolean isThreadSafe(String dataDir, String compCodec) {
...
-                byteArrayStream);
-        objectStream.writeLong(fsp.getStart());
-        objectStream.writeLong(fsp.getLength());
-        objectStream.writeObject(fsp.getLocations());
+
+        return prepareFragmentMetadata(fsp.getStart(), fsp.getLength(), fsp.getLocations());
+
+    }
+
+    public static byte[] prepareFragmentMetadata(long start, long length, String[] locations)
```
shivzone: This can be made private

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetDataFragmenter.java
```diff
@@ -0,0 +1,103 @@
...
+        JobConf jobConf = new JobConf(new Configuration(), ParquetDataFragmenter.class);
+        try {
+            job = Job.getInstance(jobConf);
+        } catch (IOException e) {
+            throw new RuntimeException("Unable to instantiate a job for reading fragments", e);
+        }
+    }
+
+
+    @Override
```
shivzone: Comments would be useful describing the components of the Fragment data here

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.parquet.io.ColumnIOFactory;
+import org.apache.parquet.io.MessageColumnIO;
+import org.apache.parquet.io.RecordReader;
+import org.apache.parquet.schema.MessageType;
+
+import java.io.IOException;
+import java.util.Iterator;
+
+/**
+ * Parquet file accessor.
```
shivzone: Also mention what exactly the accessor call returns (record or chunk etc)

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetFileAccessor.java
```diff
@@ -0,0 +1,168 @@
...
+        private Group readNextGroup() {
+            Group g = null;
+            if (rowsRemainedInRowGroup == 0) {
+                readNextRowGroup();
+                if (currentRowGroup != null) {
+                    g = recordReader.read();
+                }
+            } else {
+                g = recordReader.read();
+                if (g == null) {
```
shivzone: Instead of this. Why don't we simply invoke recordReader.read() currentRowGroup.getRowCount() # of times instead of decrementing rowsRemainedInRowGroup which is making this code look complicated

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetDataFragmenter.java
```diff
@@ -0,0 +1,103 @@
...
+    public List<Fragment> getFragments() throws Exception {
+        String absoluteDataPath = HdfsUtilities.absoluteDataPath(inputData.getDataSource());
+        ArrayList<InputSplit> splits = getSplits(new Path(absoluteDataPath));
```
denalex: usually best to declare type as List, especially since there is no direct access calls that would justify narrowing this to ArrayList

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/utilities/HdfsUtilities.java
denalex: or better to incorporate 2 lines from this function into the parent function, if it only is used once.

pxf/pxf-service/src/scripts/pxf-env.sh
```diff
@@ -54,3 +54,5 @@ export HADOOP_DISTRO=${HADOOP_DISTRO}
 # Parent directory of Hadoop client installation (optional)
 # used in case of tarball-based installation when all clients are under a common parent directory
 export HADOOP_ROOT=${HADOOP_ROOT}
+
+export CATALINA_OPTS="-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005"
```
denalex: this is for debugging, should not be committed !

sansanichfb: Sure, deleted.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetDataFragmenter.java
sansanichfb: Thanks, updated

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/utilities/HdfsUtilities.java
sansanichfb: Thanks, deleted as unnecessary

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetDataFragmenter.java
sansanichfb: This method is needed to support `org.apache.hadoop.mapreduce.lib.input.FileSplit` type.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/utilities/HdfsUtilities.java
sansanichfb: Both functions are used so I would rather keep them both for sake of compatibility.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetFileAccessor.java
sansanichfb: Even though could looks slightly more complex - we are saving on invoking that method for every single record.

pxf/pxf-hdfs/src/main/java/org/apache/hawq/pxf/plugins/hdfs/ParquetFileAccessor.java
sansanichfb: Added

PR #1328, opened 2018-01-05
HAWQ-1578. Regression Test (Feature->Ranger)Failed

chiyang10000:
Due to some unknown issue (`ERROR: remote component error (0): (libchurl.c:897)`), I could not set up the test environment to verify my fix. 
 
Please help to apply the patch into your test environment to verify it. Thanks!
@interma @radarwave

### DIFF
--- a/src/backend/access/external/plugstorage.c
+++ b/src/backend/access/external/plugstorage.c
@@ -59,6 +59,11 @@ void getExternalTableTypeInList(const char formatType,
 	}
 	else if (fmttype_is_custom(formatType))
 	{
+		*formatterType = ExternalTableType_CUSTOM;
+		*formatterName = NULL;
+	}
+	else if (fmttype_is_custom(formatType))
+	{
 		Assert(formatOptions);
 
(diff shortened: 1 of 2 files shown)


src/backend/access/external/plugstorage.c
```diff
@@ -89,6 +94,11 @@ void getExternalTableTypeInStr(const char formatType,
 		*formatterName = NULL;
 	}
 	else if (fmttype_is_custom(formatType))
+	{
+		*formatterType = ExternalTableType_CUSTOM;
+		*formatterName = NULL;
+	}
+	else if (fmttype_is_custom(formatType))
```
interma: There is a same if condition above (`fmttype_is_custom(formatType)`).
Is one of them useless?

chiyang10000: Yes. The latter branch is disabled temporarily, serving as a workaround to fix the PXF issue. Once the pluggable storage framework is well tested, this workaround will be dropped.

interma: Thanks!

radarwave:
+1

PR #1331, opened 2018-01-22
HAWQ-1557. Concurrent drop should not report error for drop IF EXISTS.

kuien:
Concurrent drop should not report error for drop IF EXISTS. The reason is: relation can potential be dropped or altered while waiting to acquire lock, it may happen that the name is no longer valid by the time DROP acquires the lock. In such case, DROP TABLE should look up the name again; if it fails again, we need to unlock the relation as well.


linwen:
LGTM.

PR #1332, opened 2018-01-24
HAWQ-1581. Separate PXF system parameters from user parameters

shivzone:


lavjain:
pxf/pxf-service/src/scripts/pxf-service should not be part of this commit.
pxf/pxf-jdbc/README.md needs to be updated

pxf/pxf-api/src/test/java/org/apache/hawq/pxf/api/utilities/ProfilesConfTest.java
```diff
@@ -96,16 +96,18 @@ public void definedProfile() throws Exception {
                 optionalFile.toURI().toURL());
 
         Map<String, String> hbaseProfile = ProfilesConf.getProfilePluginsMap("HBase");
+        System.out.println(hbaseProfile);
+        System.out.println(hbaseProfile.get("X-GP-OPTIONS-PLUGIN1"));
```
denalex: did you intend to have System.out here ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPlugin.java
```diff
@@ -58,8 +58,8 @@ public JdbcPlugin(InputData input) throws UserDataException {
         super(input);
         jdbcDriver = input.getUserProperty("JDBC_DRIVER");
         dbUrl = input.getUserProperty("DB_URL");
-        user = input.getUserProperty("USER");
-        pass = input.getUserProperty("PASS");
+        user = input.getUserProperty("DB_USER");
+        pass = input.getUserProperty("DB_PASS");
```
denalex: no longer necessary to change these property names

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/ProtocolData.java
```diff
@@ -82,19 +83,26 @@ public ProtocolData(Map<String, String> paramsMap) {
...
         /*
-         * accessor - will throw exception from getPropery() if outputFormat is
+         * accessor - will throw exception if outputFormat is
          * BINARY and the user did not supply accessor=... or profile=...
-         * resolver - will throw exception from getPropery() if outputFormat is
+         * resolver - will throw exception if outputFormat is
          * BINARY and the user did not supply resolver=... or profile=...
          */
-        profile = getOptionalProperty("PROFILE");
+        profile = getUserProperty("PROFILE");
```
denalex: should we preserve semantic of getUserProperty() that throws exception if the property is not present and getUserOptionalProperty() that returns null ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPlugin.java
shivzone: yes ... we don't need this

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/ProtocolData.java
shivzone: I used getUserProperty since I thought User property implicitly indicates that these are custom properties

shivzone:
@lavjain the pxf-service updated has been reverted.
With this patch, we no longer need to modify the pxf-jdbc user credential parameters. No readme updates required.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/ProtocolData.java
denalex: but you repeat the same code multiple times here -- throwing error if the property is not present, which makes it a mandatory "user property".

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/ProtocolData.java
```diff
@@ -82,19 +83,26 @@ public ProtocolData(Map<String, String> paramsMap) {
...
         if (profile != null) {
             setProfilePlugins();
         }
-        accessor = getProperty("ACCESSOR");
-        resolver = getProperty("RESOLVER");
-        fragmenter = getOptionalProperty("FRAGMENTER");
-        metadata = getOptionalProperty("METADATA");
+        accessor = getUserProperty("ACCESSOR");
+        if(accessor == null) {
+            protocolViolation(accessor);
```
denalex: what's the point of sending null to protocolViolation function ? would you want to send string instead ?

PR #1334, opened 2018-01-30
HAWQ-1584. Don't ignore exceptions during bridge.endIteration when writing to HDFS

lavjain:


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
```diff
@@ -159,20 +159,16 @@ private static Response writeResponse(Bridge bridge,
                 ++totalWritten;
             }
         } catch (ClientAbortException e) {
-            LOG.debug("Remote connection closed by HAWQ", e);
+            LOG.error("Remote connection closed by HAWQ", e);
```
denalex: this was not considered an error, for some reason, we are not throwing exception in this case, so maybe warning level ?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
```diff
@@ -159,20 +159,16 @@ private static Response writeResponse(Bridge bridge,
...
-            LOG.debug("totalWritten so far " + totalWritten + " to " + path);
+            LOG.error("totalWritten so far " + totalWritten + " to " + path);
             throw ex;
         } finally {
-            try {
-                bridge.endIteration();
-            } catch (Exception e) {
-                // ignore ... any significant errors should already have been handled
-            }
+            bridge.endIteration();
```
denalex: if Exception happened in try and is rethrown by catch and also endIteration throws an exception, then the original one will be forgotten and the latest one form endIteration will be thrown. If this is ok, we should at least log at error level the original one in catch block. Alternatively, have an error flag set by catch and then ignore exception here only if the flag is set to have the original one propagate to the client.

lavjain: Both (original catch exception as well endIteration) are logging at error level. Any exception would be propagated to the client.

denalex: I do not see exception being logged: `LOG.error("totalWritten so far " + totalWritten + " to " + path);` and only the latest one will be returned to client, not the original which potentially caused the issue

radarwave:
Any Jira number for this PR and commit? Thanks.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
```diff
@@ -143,36 +142,38 @@ private static synchronized Response synchronizedWriteResponse(Bridge bridge,
...
+        } catch (Exception e) {
+            LOG.error("Exception: totalWritten so far " + totalWritten + " to " + path, e);
+            ex = e;
         } finally {
             try {
                 bridge.endIteration();
             } catch (Exception e) {
-                // ignore ... any significant errors should already have been handled
+                if (ex == null)
+                    ex = e;
```
denalex: Another way would be to preserve throwing the exception inside original catch block (line 162), then here you would say
```
if (ex == null) 
     throw e;
 else 
     throw ex;
``` 
and you would not need the block below (lines 170-172) as the original will still be thrown if endIterations() completes without an error.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/rest/WritableResource.java
```diff
@@ -143,36 +142,38 @@ private static synchronized Response synchronizedWriteResponse(Bridge bridge,
...
             }
+            // propagate any exceptions
+            if (ex != null)
+                throw ex;
         }
 
         String censuredPath = Utilities.maskNonPrintables(path);
-        returnMsg = "wrote " + totalWritten + " bulks to " + censuredPath;
+        String returnMsg = "wrote " + totalWritten + " bulks to " + censuredPath;
```
denalex: move string concat inside LOG.isDebugEnabled()

lavjain: returnMsg is also being used in the response.

PR #1335, opened 2018-01-31
Hawq 1536

yk-st:
As Cloudera has mentioned below
https://www.cloudera.com/documentation/enterprise/5-12-x/topics/cm_ig_parcels.html

Hadoop jar was installed in /opt/cloudera/parcels/CDH/ from the Cloudera Manager(parcel), from yum or somthing(package) , installed in /usr/lib/hadoop.
To deal with both parcel and package, Replace CDH_HADOOP_JAR with specific value.


lavjain:
@yk-st Your change looks good. However, it might be cleaner to create a separate template for CDH manager (eg `pxf-private-cdh-manager.classpath.template`) rather than substituting the token values. This will also ensure that PXF pipelines for GPDB are not affected.

shivzone:
@yk-st any update on this ?

PR #1341, opened 2018-02-11
HAWQ-1583 Add vectorized executor extension and GUC

weinan003:
1. Create vexecutor plugin folder in contrb, all logic for vectorization
execution will implement here

2. Add hook point in funciton ExecInitNode, ExecProcNode and ExecEndNode
respectively

3. create new GUC boolean type key "vectorized_executor_enable" as
framework switch

### DIFF
--- a/src/backend/executor/execProcnode.c
+++ b/src/backend/executor/execProcnode.c
@@ -147,6 +147,7 @@
 #include "utils/debugbreak.h"
 #include "pg_trace.h"
 
+VectorExecMthd vmthd = {};
 #ifdef CDB_TRACE_EXECUTOR
 #include "nodes/print.h"
 static void ExecCdbTraceNode(PlanState *node, bool entry, TupleTableSlot *result);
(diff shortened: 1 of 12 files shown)


wengyanqing:
It's better to add a README for describing what this module is and how it works.

src/include/executor/executor.h
```diff
@@ -41,6 +41,17 @@
 
 #include "cdb/cdbdef.h"                 /* CdbVisitOpt */
 
+#define ISVECTORIZED(note) vectorized_executor_enable
```
wengyanqing: note -> node ?

src/backend/executor/execProcnode.c
```diff
@@ -797,7 +801,7 @@ ExecSliceDependencyNode(PlanState *node)
 	ExecSliceDependencyNode(outerPlanState(node));
 	ExecSliceDependencyNode(innerPlanState(node));
 }
-    
+
```
wengyanqing: This diff should be removed.

wengyanqing:
All new files should add ASF license header.

PR #1344, opened 2018-02-26
HAWQ-1599. PXF Ignite plugin

leskin-in:
Add a new PXF plugin to provide read and wrtie access to [Ignite database](https://ignite.apache.org/) via its native REST API.

Key features:
* Supports partitioning;
* A dedicated Ignite client with REST API enabled is used to access Ignite database;
* Every SELECT or INSERT query splits tuples of data into groups, each of which is sent from (to) Ignite in one response (query). The size of a group is customizable;
* The implemented REST API can be modified (with relatively small changes) and used to connect to any other database (or service) which supports SQL queries via REST, and responds using JSON objects.

Plugin documentation is available at `pxf/pxf-ignite/README.md`.

### Dependencies
* [GSON](https://github.com/google/gson) is required to serialize and deserialize JSON;
* [Apache commons compress ByteUtils](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/utils/ByteUtils.html) is required to properly encode fragment metadata.


pxf/pxf-ignite/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+# Partitioning
+## Introduction
+
+PXF Ignite plugin supports simultaneous access to Ignite database from multiple PXF segments. *Partitioning* should be used in order to perform such operation.
+
+Partitioning in PXF Ignite plug-in works just like in PXF JDBC plug-in.
+
+This feature is optional. However, a bug in the `pxf-service` which makes partitioning necessary for any query was fixed only on 17th Jan 2018 in [this commit](https://github.com/apache/incubator-hawq/commit/0d620e431026834dd70c9e0d63edf8bb28b38227), so the older versions of PXF may return an exception if a query does not contain a meaningful `PARTITION_BY` parameter.
```
sansanichfb: What's default partitioning strategy if `PARTITION_BY` was omitted?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/ByteUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+import org.apache.commons.lang.ArrayUtils;
+
+/**
+ * A tool class, used to deal with byte array merging, split and other methods.
+ * IgnitePartitionFragmenter requires this class in order to work.
+ */
+public class ByteUtil {
+
+    public static byte[] mergeBytes(byte[] b1, byte[] b2) {
```
sansanichfb: Why not use ArrayUtils.addAll directly?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java
```diff
@@ -0,0 +1,502 @@
...
+        }
+
+        // Insert partition constaints
+        sb = new IgnitePartitionFragmenter(inputData).buildFragmenterSql(sb);
+
+        // Format URL
+        urlReadStart = buildQueryFldexe(sb.toString(), filterConstants);
+
+        // Send the first REST request that opens the connection
+        JsonElement response = sendRestRequest(urlReadStart);
```
sansanichfb: Maybe check if a request was successful and return `false` if not.

pxf/pxf-ignite/README.md
leskin-in: All the data is retrieved by a single PXF segment.

Added this to README.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/ByteUtil.java
leskin-in: This piece of code comes from PXF JDBC plugin; for sure, `ArrayUtils.addAll()` can be used directly; this also removes the only dependence of `ByteUtil.java`. As this package is used to process byte arrays in `IgnitePartitionFragmenter` anyway, an extra dependence may be imported there without any harm, I suppose.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java
leskin-in: This check is performed by `sendRestRequest()` function. If the request was not successfull, `sendRestRequest` will throw a `java.net.MalformedURLException`, `java.io.IOException` or `java.net.ProtocolException`. All these errors will include the URL that caused them and (in case of error reported by Ignite in it's response) additional data. As these exceptions should be passed "up", they are not handled here. See https://github.com/apache/incubator-hawq/pull/1344/files#diff-f5894bad29c44342eee2760e259a91d6R480 for the details.

When the exception is thrown by `readNextObject()`, the procedure `closeForRead()` is called, as it would be in case `false` was returned. So both variants (`return false` and `throw ...`) are safe. However, (AFAIK) if the exception is not thrown by `readNextObject()`, the user will not get the message that the read went wrong. So throwing exceptions from `public` `IgniteAccessor` functions is necessary.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java
```diff
@@ -0,0 +1,502 @@
...
+        sb.append("pageSize=0");
+        sb.append("&");
+        if (cacheName != null) {
+            sb.append("cacheName=");
+            // Note that Ignite supports only "good" cache names (those that should be left unchanged by the URLEncoder.encode())
+            sb.append(URLEncoder.encode(cacheName, "UTF-8"));
+            sb.append("&");
+        }
+        int counter = 1;
+        if (filterConstants != null) {
```
hornn: could you please the filter construction? it looks like it creates something like
`arg1=A&arg2=B&arg3=C`
how do we know what field argX stands for? and if we want to filter by the last field, how can it be done? (I would think in that case we will have to pass the filterConstants with empty values as placeholders for the other fields?)
If I understand correctly, this code is never executed because we always send filterConstants as null, so I just want to make sure it will work as expected.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePlugin.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hawq.pxf.api.UserDataException;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.apache.hawq.pxf.api.utilities.Plugin;
+
+/**
+ * This class resolves the jdbc connection parameter and manages the opening and closing of the jdbc connection.
```
denalex: copy & paste from JDBC ?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePlugin.java
```diff
@@ -0,0 +1,90 @@
...
+            LOG.debug("Constructor started");
+        }
+        
+        igniteHost = inputData.getUserProperty("IGNITE_HOST");
+        if (igniteHost == null) {
+            igniteHost = igniteHostDefault;
+        }
+
+        cacheName = inputData.getUserProperty("IGNITE_CACHE");
+        if (cacheName == null) {
```
denalex: remove this ?

pxf/pxf-service/src/configs/templates/pxf-private-hdp.classpath.template
```diff
@@ -62,6 +63,8 @@ PXF_HOME/lib/pxf-jdbc.jar
 /usr/hdp/current/hadoop-client/client/snappy-java.jar
 /usr/hdp/current/hadoop-client/lib/asm-*[0-9].jar
 /usr/hdp/current/hadoop-client/lib/jersey-server-*[0-9].jar
+/usr/hdp/current/hadoop-client/client/gson.jar
+
```
denalex: extra line ?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -193,4 +193,13 @@ under the License.
             <resolver>org.apache.hawq.pxf.plugins.hdfs.ParquetResolver</resolver>
         </plugins>
     </profile>
+    <profile>
+        <name>Ignite</name>
+        <description>A profile to read and write data from/to Ignite</description>
```
denalex: should refer to it as Apache Ignite ?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePlugin.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hawq.pxf.api.UserDataException;
+import org.apache.hawq.pxf.api.utilities.InputData;
+import org.apache.hawq.pxf.api.utilities.Plugin;
+
+/**
+ * This class resolves the jdbc connection parameter and manages the opening and closing of the jdbc connection.
+ * Implemented subclasses: {@link IgniteReadAccessor}.
```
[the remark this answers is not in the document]
hornn: Also please add IgniteResolver to the list.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+
+        columns = input.getTupleDescription();
+        
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Constructor successful");
+        }
+    }
+
+    /**
+     * Transform a JsonArray object stored in {@link OneRow} into a List {@link OneField}
```
hornn: typo `a list of OneField`

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+     */
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
+        JsonArray result = (JsonArray)row.getData();
```
hornn: just wondering - does this line generate a warning? if so, we should make a safer cast or suppress the warning.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+        LinkedList<OneField> fields = new LinkedList<OneField>();
+
+        for (int i = 0; i < columns.size(); i++) {
+            Object value = null;
+            OneField oneField = new OneField(columns.get(i).columnTypeCode(), null);
+
+            // Handle null values
+            if (result.get(i).isJsonNull()) {
+                oneField.val = null;
```
hornn: is this line necessary? don't we initialize the oneField with null value in line 76?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+
+    /**
+     * Transform a JsonArray object stored in {@link OneRow} into a List {@link OneField}
+     */
+    @Override
+    public List<OneField> getFields(OneRow row) throws Exception {
+        JsonArray result = (JsonArray)row.getData();
+        LinkedList<OneField> fields = new LinkedList<OneField>();
+
+        for (int i = 0; i < columns.size(); i++) {
```
hornn: what happens if the json array has more/less fields than the table? should we add a check that the sizes match?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+                case REAL:
+                case BIGINT:
+                case SMALLINT:
+                case BOOLEAN:
+                    sb.append(String.valueOf(oneField.val));
+                    break;
+                case VARCHAR:
+                case BPCHAR:
+                case TEXT:
+                case NUMERIC:
```
hornn: why does NUMERIC type need quotes?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+                    sb.append("'" + String.valueOf(oneField.val) + "'");
+                    break;
+                case BYTEA:
+                    sb.append("'" + Hex.encodeHexString((byte[])(oneField.val)) + "'");
```
hornn: usually getFields and setFields use the same logic (the idea is that if you write and then read the same data, you will get the same data back). Just wondering why use here Hex and not Base64 serialization?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/WhereSQLBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+import org.apache.hawq.pxf.api.io.DataType;
+import org.apache.hawq.pxf.api.utilities.ColumnDescriptor;
+import org.apache.hawq.pxf.api.utilities.InputData;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Parse filter object generated by parent class  {@link org.apache.hawq.pxf.plugins.ignite.IgniteFilterBuilder},
```
hornn: minor typos - please remove extra spaces: after `parent class`, after `Multiple filters`, after `HDOP_AND`, before and after `LogicalOperation`.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/WhereSQLBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+ */
+public class WhereSQLBuilder extends IgniteFilterBuilder {
+    private InputData inputData;
+
+    public WhereSQLBuilder(InputData input) {
+        inputData = input;
+    }
+
+    /**
+     * 1.check for LogicalOperator, Jdbc currently only support HDOP_AND.
```
hornn: typo jdbc?

pxf/pxf-ignite/src/test/java/org/apache/hawq/pxf/plugins/ignite/SqlBuilderTest.java
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Validate SQL string generated by the {@link IgnitePartitionFragmenter#buildFragmenterSql} method
+ * and the {@link WhereSQLBuilder#buildWhereSQL} method.
+ */
+public class SqlBuilderTest {
+    private static final Log LOG = LogFactory.getLog(SqlBuilderTest.class);
+    static final String ORIGINAL_SQL = "SELECT * FROM sales";
+    InputData inputData;
+
+    @Before
```
hornn: are the before/after necessary?

pxf/pxf-ignite/src/test/java/org/apache/hawq/pxf/plugins/ignite/IgniteFilterBuilderTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void parseNestedExpressionWithLogicalOperation() throws Exception {
+        WhereSQLBuilder builder = new WhereSQLBuilder(null);
+        LogicalFilter filter = (LogicalFilter) builder.getFilterObject("a1c25s5dfirsto5a2c20s1d2o2l0a1c20s1d1o1l1");
+        assertEquals(LogicalOperation.HDOP_OR, filter.getOperator());
+        assertEquals(LogicalOperation.HDOP_AND, ((LogicalFilter) filter.getFilterList().get(0)).getOperator());
+        assertEquals(HDOP_LT, ((BasicFilter) filter.getFilterList().get(1)).getOperation());
+    }
+
+    @Test
+    public void parseISNULLExpression() throws Exception {
```
hornn: maybe change test names to `parseIsNullExpression` and `parseIsNotNullExpression`

pxf/pxf-ignite/src/test/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+
+        acc.openForRead();
+        acc.readNextObject();
+        acc.closeForRead();
+
+        PowerMockito.verifyPrivate(acc, Mockito.times(3)).invoke(method(IgniteAccessor.class, "sendRestRequest", String.class)).withArguments(captor.capture());
+
+        List<String> allParams = captor.getAllValues();
+
+        assertEquals(allParams.get(0), "http://127.0.0.1:8080/ignite?cmd=qryfldexe&pageSize=0&qry=SELECT+id%2Cname%2Cbirthday%2Ckey+FROM+TableTest");
```
hornn: just wondering - when do the spaces in the query get replaced by `+`?

pxf/pxf-ignite/src/test/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessorTest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.mockito.Mockito;
+import static org.mockito.Matchers.anyString;
+
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+import static org.powermock.api.support.membermodification.MemberMatcher.method;
+
+
+@RunWith(PowerMockRunner.class)
```
hornn: Really good tests!

pxf/pxf-ignite/src/test/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessorTest.java
leskin-in: When `java.net.URLEncoder.encode()` is called [here in IgniteAccessor](https://github.com/arenadata/incubator-hawq/blob/06a287693f6bdeef72c43d606aeebf44a18ced98/pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java#L342)

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
leskin-in: Fixed: https://github.com/arenadata/incubator-hawq/blob/a23df5964ecee25fefd8509da0169026c2f7ac0d/pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java#L154

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java
```diff
@@ -0,0 +1,502 @@
...
+        sb.append(tableName);
+
+        // Insert the column names
+        sb.append("(");
+        ArrayList<ColumnDescriptor> columns = inputData.getTupleDescription();
+        if (columns == null) {
+            throw new UserDataException("Tuple description must be present.");
+        }
+        String fieldDivisor = "";
+        for (int i = 0; i < columns.size(); i++) {
```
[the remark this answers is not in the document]
leskin-in: I don't see the need for it here. Although it is used in both `openForRead()` and `openForWrite()`, in `openForWrite` it is surrounded by `(` and `)`, so the code there would look odd, I suppose. Besides, the function would encapsulate just six lines.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteAccessor.java
leskin-in: This behaviour is correct.

Apache Ignite [supports](https://apacheignite.readme.io/docs/rest-api#section-sql-fields-query-execute) passing the constraints' constants separately from the constraint expression itself. I am not sure why there is such an option; maybe it is there for safety (to protect Ignite from SQL injections, if the request is formed by, say, javascript on some webpage).

This feature may be implemented in the future, so I left the code as it is. To implement the feature, relatively serious changes must be done both in `IgnitePartitionFragmenter` and `WhereSqlBuilder`; however, the effect of those changes is not evident.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
```diff
@@ -0,0 +1,173 @@
...
+import java.text.SimpleDateFormat;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.codec.binary.Hex;
+
+import com.google.gson.JsonArray;
+
+
```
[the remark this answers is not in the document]
leskin-in: Is it correct to put two new lines after all the imports?

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgniteResolver.java
leskin-in: That is how Ignite handles such fields :smile:

I have not found the Ignite documentation for this case. However, when Base64 field is sent to Ignite, it handles it some wrong way (it does not return an error, too).

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePartitionFragmenter.java
```diff
@@ -0,0 +1,411 @@
...
+    }
+
+    /**
+     * Insert partition constraints into the prepared SQL query.
+     * 
+     * @param inputData pre-validated PXF InputData
+     * @param sb the SQL query that is prepared for appending WHERE constraints.
+     * Other SQL statements may be present, but they must be complete. Note that no check is performed to check their "completeness"
+     */
+    public static StringBuilder buildFragmenterSql(InputData inputData, StringBuilder sb) {
```
sansanichfb: Do we really want to mutate input parameter and then return it?

leskin-in: The `StringBuilder sb` may include other WHERE constraints (e.g., partition constraints), thus the result of `buildFragmenterSql()` strongly depends on the contents of `sb`. 

Consider the following example. Let `buildFragmenterSql` return a `String`. In some cases the return value will be like `WHERE id=10 AND ...`, while in other cases ` AND id=10 AND ...`. In the second case, the output has no sense without the initial query, so both the initial query and the result of the function must be present.

Besides, passing a mutable parameter here points to the fact that the existing `WHERE` constraints (present in `StringBuilder sb`) must be complete and that the query will be changed accordingly.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/ByteUtil.java
```diff
@@ -0,0 +1,77 @@
...
+        byte[][] newBytes = new byte[len][];
+        int j = 0;
+        for (int i = 0; i < len; i++) {
+            newBytes[i] = new byte[n];
+            for (int k = 0; k < n; k++) newBytes[i][k] = bytes[j++];
+        }
+        return newBytes;
+    }
+
+    public static byte[] getBytes(long value) {
```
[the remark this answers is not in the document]
leskin-in: Done. Thank you for the comment.

This also eliminates the dependency on Apache Commons ArrayUtils.

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePartitionFragmenter.java
sansanichfb: Since you are mutating initial query - I think it would be cleaner to do it in place and return `void`.

lisakowen:
@leskin-in - i am writing pxf developer docs.  i have a few questions if you are willing to share:

- on what OS did you develop the plug-in?
- what build tool/IDE did you use?
- where did you find reference material for your work?  HAWQ PXF wiki?  HAWQ docs?  code? - which was the most relevant/helpful?
- what plug-in (fragmenter, accessor, resolver) did you implement first?
- do you have any design/specification documents for the ignite work that you would be willing to share?

thank you!

pxf/pxf-ignite/src/main/java/org/apache/hawq/pxf/plugins/ignite/IgnitePartitionFragmenter.java
leskin-in: Thank you! This is a better solution. Done.

leskin-in:
@lisakowen, 
* Ubuntu 16.04; CentOS 7.4
* Gradle 2.10, Visual Studio Code 1.20.1, IntelliJ Idea 2017.3
* Key sources (the most useful first): http://hawq.incubator.apache.org/docs/userguide/2.1.0.0-incubating/pxf/PXFExternalTableandAPIReference.html (this article is far more useful than anything else I found and used); PXF JDBC plug-in code; PXF API code
* Accessor. The key details on how to implement other classes are already present in other plugins
* I am going to expand the README to add information on `INSERT` queries and perfomance test results. Other documentation is mostly in the code itself. If adding some extra docs would be really helpful, please let me know.

kapustor:
@denalex @hornn @sansanichfb Is there anything we need to do/fix for this PR to be merged?

radarwave:
@kapustor To get this PR merged, we need this PR have two '+1' from the reviewer, and you need to create corresponding jira and include the jira number in the title of the PR. You can refer to other PRs. Thanks.

kapustor:
Hi guys,

Any news on this PR?

denalex:
@kapustor -- we are ready to commit it if you do not plan any further changes. Need to put it through a regression run on our side, hope to be done with it next week. Thanks.

PR #1348, opened 2018-03-19
HAWQ-1593. Vectorized execution condition check in plan tree

zhangjackey:
1. Replace the vectorized data types and expressions immediately after check.
 2.Move the VExecInitNode to the end of the ExecInitNode.
 3.Process the Parent PlanState and the vectorized flag in the VExecInitNode.

### DIFF
--- a/contrib/vexecutor/vcheck.c
+++ b/contrib/vexecutor/vcheck.c
@@ -57,8 +57,8 @@ static HTAB *hashMapVFunc = NULL;
 
 typedef struct VecFuncHashEntry
 {
-    Oid src;
-    vFuncMap *vFunc;
+	Oid src;
+	vFuncMap *vFunc;
 } VecFuncHashEntry;
 
 typedef struct VecTypeHashEntry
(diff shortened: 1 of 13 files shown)


contrib/vexecutor/vcheck.c
```diff
@@ -54,7 +61,274 @@ typedef struct VecFuncHashEntry
...
+
+    ReplacePlanVectorzied(root, plan->lefttree);
+    ReplacePlanVectorzied(root, plan->righttree);
+    ReplacePlanNodeWalker(root, plan);
+
+    return plan;
+}
+
+Plan*
+CheckAndReplacePlanVectorized(PlannerInfo *root, Plan *plan)
```
wengyanqing: How to check initPlans of main Plan node?

contrib/vexecutor/vcheck.c
```diff
@@ -54,7 +61,274 @@ typedef struct VecFuncHashEntry
...
+
+        if(ctx->replace)
+        {
+        	    op->opresulttype = rettype;
+        	    op->opfuncid = voper->oprcode;
+        }
+
+        ctx->retType = rettype;
+        return false;
+    }
```
wengyanqing: Maybe need to check more nodes, such as FuncExpr.

contrib/vexecutor/vcheck.c
zhangjackey: So far it is not supported, but it does not seem to affect the normal running.

contrib/vexecutor/vcheck.c
zhangjackey: Yes, we will support more expression nodes step by step.

wengyanqing:
LGTM.

PR #1349, opened 2018-03-27
HAWQ-1598. Vectorized Scan Node Framework initialization

weinan003:


src/test/feature/vexecutor/test_vexecutor.cpp
```diff
@@ -60,11 +59,47 @@ TEST_F(TestVexecutor, vadt)
...
+				 "	fivethous	int4,"
+				 "	tenthous	int4,"
+				 "	odd			int4,"
+				 "	even		int4,"
+				 "	stringu1	name,"
+				 "	stringu2	name,"
+				 "	string4		name) WITH (appendonly = true, orientation = PARQUET, pagesize = 1048576, rowgroupsize = 8388608, compresstype = SNAPPY) DISTRIBUTED RANDOMLY;");
+
+  std::string pwd = util.getTestRootPath();
+  std::string cmd = "COPY test1 FROM '" + pwd + "/vexecutor/data/tenk.data'";
```
interma: why need so many data in tenk.data?

contrib/vexecutor/vadt.c
```diff
@@ -39,10 +39,11 @@ size_t v##type##Size(vheader *vh) \
 vheader* buildv##type(int n) \
 { \
 	vheader* result; \
-	result = (vheader*) palloc0(offsetof(v##type,values) + n * sizeof(type)); \
+	result = (vheader*) malloc(offsetof(v##type,values) + n * sizeof(type)); \
+	memset(result,0,offsetof(v##type,values) + n * sizeof(type)); \
 	result->dim = n; \
 	result->elemtype = typeoid; \
-	result->isnull = palloc(sizeof(bool) * n); \
+	result->isnull = malloc(sizeof(bool) * n); \
```
interma: why use malloc here? and the below code is still palloc()

src/test/feature/vexecutor/test_vexecutor.cpp
weinan003: I believe that it can relieve future test workload.

contrib/vexecutor/vadt.c
weinan003: It does make sense, I will attach another commit change it

contrib/vexecutor/vexecutor.c
```diff
@@ -21,15 +21,23 @@
...
+#include "miscadmin.h"
+#include "tuplebatch.h"
+#include "executor/nodeTableScan.h"
+#include "catalog/catquery.h"
+#include "cdb/cdbvars.h"
+#include "execVScan.h"
 
 PG_MODULE_MAGIC;
-
+#define BATCHSIZE 1024
```
wengyanqing: It's better to make BATCHSIZE as GUC so that we could tune this value easily.

contrib/vexecutor/vexecutor.c
```diff
@@ -96,20 +140,70 @@ static PlanState* VExecInitNode(PlanState *node,EState *eState,int eflags)
...
+	bool ret = false;
+	switch (nodeTag(node))
+	{
+		case T_AppendOnlyScanState:
+		case T_ParquetScanState:
+            tbDestroy((TupleBatch *) (&node->ps_ResultTupleSlot->PRIVATE_tb));
+            tbDestroy((TupleBatch *) (&((TableScanState *) node)->ss.ss_ScanTupleSlot->PRIVATE_tb));
+			ret = true;
+			break;
+		case T_TableScanState:
```
wengyanqing: The logic of case T_TableScanState in VExecEndNode is not the same as VExecInitNode, what's the difference ?

contrib/vexecutor/execVScan.c
```diff
@@ -0,0 +1,167 @@
...
+getVScanMethod(int tableType)
+{
+    static const ScanMethod scanMethods[] =
+            {
+                    //HEAPSCAN
+                    {
+                    },
+                    //APPENDONLYSCAN
+                    {
+                            &AppendOnlyScanNext, &BeginScanAppendOnlyRelation, &EndScanAppendOnlyRelation,
```
wengyanqing: Are these scan functions placeholders?

contrib/vexecutor/vexecutor.c
```diff
@@ -66,12 +75,47 @@ _PG_fini(void)
 	vmthd.ExecEndNode_Hook = NULL;
 }
 
-static PlanState* VExecInitNode(PlanState *node,EState *eState,int eflags)
+static void backportTupleDescriptor(PlanState* ps,TupleDesc td)
```
wengyanqing: What's the purpose of this function ? Would you give the comment ?

weinan003: Due to Vectorized type unrecognizable in normal execution node, the TupleDesc structure should backport metadata to normal type before we do V->N result pop up. At present, we only support scan node vectorized execution, so we force this backport function invoked. In the future, we will check if its parent node is "N" type. I will comment the function

contrib/vexecutor/execVScan.c
weinan003: Well, this PR only provide a vectorized scan framework, AO and Parquet vectorized scan function will implement later. For easy to do feature test for this PR. I use normal AO and Parquet function as the tuple scanner under the framework

contrib/vexecutor/vexecutor.c
weinan003: In Init process, we prefer to replace some metadata after normal init process done, because both normal and vectorized init process most part are same. In the End step, in the current, I think there is no more we need to do. Tuplebatch buffer which is reusable buffer can be recycled by memory context, others data destruct in the normal source release function.

amyrazz44:
LGTM +1

PR #1350, opened 2018-03-30
HAWQ-1600. Parquet table data vectorized scan

weinan003:


contrib/vexecutor/execVQual.c
```diff
@@ -0,0 +1,125 @@
...
+        /* simple Var list: this always succeeds with one result row */
+        if (isDone)
+            *isDone = ExprSingleResult;
+
+        ExecVecVariableList(projInfo,slot->PRIVATE_tb);
+        ExecStoreVirtualTuple(slot);
+    }
+    else
+    {
+        elog(FATAL,"does not support expression in projection stmt");
```
wengyanqing: Any unsupport operation in vectorized execution, it should be check before execution and fallback to original executor. Must make sure it could get right query results.

contrib/vexecutor/execVScan.c
```diff
@@ -38,8 +40,8 @@ getVScanMethod(int tableType)
                     },
                     //PARQUETSCAN
                     {
-                            &ParquetScanNext, &BeginScanParquetRelation, &EndScanParquetRelation,
-                            &ReScanParquetRelation, &MarkRestrNotAllowed, &MarkRestrNotAllowed
+                            &ParquetVScanNext, &BeginScanParquetRelation, &EndScanParquetRelation,
+                            NULL,NULL,NULL
```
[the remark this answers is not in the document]
weinan003: At present, we do not know what kinds of query can cover these function.

contrib/vexecutor/execVQual.c
weinan003: The expression op will be supported soon in next

contrib/vexecutor/ao_reader.c
```diff
@@ -0,0 +1,78 @@
...
+
+        slot = scanState->ss_ScanTupleSlot;
+        if(TupIsNull(slot))
+            break;
+
+        for(int i = 0;i < tb->ncols ; i ++)
+        {
+
+           if(vs->proj[i])
+            {
```
linwen: redundant space here

contrib/vexecutor/ao_reader.c
```diff
@@ -0,0 +1,78 @@
...
+                /* if attribute is a reference, deep copy the data out to prevent ao table buffer free before vectorized scan batch done */
+                if(!slot->tts_mt_bind->tupdesc->attrs[i]->attbyval)
+                    *ptr = datumCopy(*ptr,slot->tts_mt_bind->tupdesc->attrs[i]->attbyval,slot->tts_mt_bind->tupdesc->attrs[i]->attlen);
+            }
+        }
+
+        AppendOnlyScanDesc scanDesc = ((AppendOnlyScanState*)scanState)->aos_ScanDesc;
+        VarBlockHeader *header = scanDesc->executorReadBlock.varBlockReader.header;
+
+        //if(row + 1 == VarBlockGet_itemCount(header))
```
linwen: Please remove useless codes.

contrib/vexecutor/parquet_reader.c
```diff
@@ -0,0 +1,194 @@
...
+	{
+		ParquetRowGroupReader_FinishedScanRowGroup(rowGroupReader);
+		return false;
+	}
+
+	/*
+	 * get the next item (tuple) from the row group
+	 */
+	int ncol = slot->tts_tupleDescriptor->natts;
+    TupleBatch tb = (TupleBatch )slot->PRIVATE_tb;
```
linwen: Please fix indent here.

contrib/vexecutor/parquet_reader.c
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	int colReaderIndex = 0;
+	for(int i = 0; i < tb->ncols ; i++)
+	{
+		if(projs[i] == false)
+			continue;
+
+		Oid hawqTypeID = tupDesc->attrs[i]->atttypid;
+        Oid hawqVTypeID = GetVtype(hawqTypeID);
```
linwen: Please fix indent here.

contrib/vexecutor/parquet_reader.c
```diff
@@ -0,0 +1,194 @@
...
+		if(!tb->datagroup[i])
+			tbCreateColumn(tb,i,hawqVTypeID);
+
+		vheader* header = tb->datagroup[i];
+        header->dim = tb->nrows;
```
linwen: Please fix indent here.

contrib/vexecutor/vcheck.h
```diff
@@ -27,6 +27,8 @@ typedef struct vFuncMap
     Oid ntype;
     vheader* (* vtbuild)(int n);
     void (* vtfree)(vheader **vh);
+	Datum (* gettypeptr)(vheader *vh,int n);
```
linwen: Please fix indent here.

contrib/vexecutor/vcheck.h
```diff
@@ -37,6 +39,7 @@ typedef struct VectorizedState
 {
 	bool vectorized;
 	PlanState *parent;
+    bool* proj;
```
linwen: Please fix indent here.

contrib/vexecutor/parquet_reader.c
```diff
@@ -0,0 +1,194 @@
...
+
+		/* no more items in the row group, get new buffer */
+		scan->bufferDone = true;
+	}
+}
+
+/*
+ * Get next tuple batch from current row group into slot.
+ *
+ * Return false if current row group has no tuple left, true otherwise.
```
linwen: According the comments, this function returns true or false, but at last it returns a number of rows. If the function returns a number, it should not return false when it finish scan row group, use "0" instead, since there is no rows. If this function return a bool, it should not return a number.

contrib/vexecutor/execVScan.c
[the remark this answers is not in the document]
wengyanqing: So we need to understand what's the different between MarkRestrNotAllowed and NULL, is it possible to cause crash if setting NULL ?

contrib/vexecutor/Makefile
```diff
@@ -17,7 +17,7 @@
 
 
 MODULE_big = vexecutor
-OBJS    = vexecutor.o vadt.o vcheck.o tuplebatch.o execVScan.o
+OBJS    = vexecutor.o vadt.o vcheck.o tuplebatch.o execVScan.o execVQual.o parquet_reader.o ao_reader.o
```
interma: Should add some feature tests for ao & parquet scan.

contrib/vexecutor/vcheck.h
```diff
@@ -39,6 +39,7 @@ typedef struct VectorizedState
...
```
wengyanqing: Because this variable is only related to ao table, or has different content based on different table format. It's better to rename it to a more readable name, aoprojs or just use (void * opaque).

PR #1351, opened 2018-04-04
HAWQ-1603.add new hook API

zhangjackey:
add two hook API.


contrib/vexecutor/execVQual.c
```diff
@@ -123,3 +123,431 @@ VirtualNodeProc(ScanState* state,TupleTableSlot *slot){
...
+
+	/* isNull is a single value, it can not be used when data is vectorized */
+	*isNull = false;
+
+	/* Fetch the value from the slot */
+	tb = (TupleBatch )slot->PRIVATE_tb;
+
+	Assert(NULL != tb);
+
+	return PointerGetDatum(tb->datagroup[attnum]);
```
wengyanqing: attnum's value starts with 1, it should be attnum -1 as the index of tb->datagroup ?

contrib/vexecutor/execVQual.c
```diff
@@ -123,3 +123,431 @@ VirtualNodeProc(ScanState* state,TupleTableSlot *slot){
...
+				bstate->args = (List *)
+					ExecInitExpr((Expr *) boolexpr->args, parent);
+				state = (ExprState *) bstate;
+			}
+			break;
+		/*TODO: More and more expressions should be vectorized */
+		default:
+			break;
+	}
+
```
wengyanqing: Does it need "state->expr = node;" here?

contrib/vexecutor/execVQual.c
```diff
@@ -123,3 +123,431 @@ VirtualNodeProc(ScanState* state,TupleTableSlot *slot){
...
+						break;
+				}
+				bstate->args = (List *)
+					ExecInitExpr((Expr *) boolexpr->args, parent);
+				state = (ExprState *) bstate;
+			}
+			break;
+		/*TODO: More and more expressions should be vectorized */
+		default:
+			break;
```
wengyanqing: It's better to think about how to fallback to original executor when we find any unsupported behavior to make sure the query could have right result. There are so many places that just error out or assert(false) which is not friendly.

contrib/vexecutor/vexecutor.c
```diff
@@ -163,24 +165,35 @@ static PlanState* VExecInitNode(PlanState *node,EState *eState,int eflags,Memory
 			case T_AppendOnlyScan:
 			case T_ParquetScan:
 			case T_TableScanState:
-				START_MEMORY_ACCOUNT(curMemoryAccount);
+				//START_MEMORY_ACCOUNT(curMemoryAccount);
```
wengyanqing: Any reason that comment this code line ?

contrib/vexecutor/execVQual.c
```diff
@@ -123,3 +123,431 @@ VirtualNodeProc(ScanState* state,TupleTableSlot *slot){
...
+
+		default:				/* get the tuple from the relation being
+								 * scanned */
+			slot = econtext->ecxt_scantuple;
+			break;
+	}
+
+	/* isNull is a single value, it can not be used when data is vectorized */
+	*isNull = false;
+
```
wengyanqing: It's better to check or assert to make sure slot is not null here.

contrib/vexecutor/execVQual.c
zhangjackey: yes, add in next commit.

contrib/vexecutor/execVQual.c
zhangjackey: add Assert in next commit.

contrib/vexecutor/execVQual.c
zhangjackey: yes, change it to attnum - 1.

PR #1352, opened 2018-04-08
HAWQ-1604. Add A New GUC hawq_hashjoin_bloomfilter 

linwen:
HAWQ-1604. Add A New GUC hawq_hashjoin_bloomfilter to indicate if use Bloom filter for hash join.
Remove gp_hashjoin_bloomfilter and bloom filter in hash join table, this part of legacy codes has been verified that it won't improve hash join performance.

Please review, thanks!


wengyanqing:
LGTM.

PR #1353, opened 2018-04-09
HAWQ-1605. Support INSERT in PXF JDBC plugin

leskin-in:
Add support of INSERT queries in PXF JDBC plugin:
* Implement interfaces `WriteAccessor` (by `JdbcAccessor` class) and `WriteResolver` (by `JdbcResolver` class);
* Support query batching in `JdbcAccessor` when processing INSERT query. The size of a batch is defined by user and may be "infinite";
* In `JdbcAccessor`, use `java.sql.PreparedStatement` and JDBC standard functions to process queries (and to support batching);
* In `setFields()` method of `JdbcResolver`, perform type conversions of the data tuples received from PXF;
* Support both transactional and non-transactional databases when performing INSERT queries.

Refactor the code in PXF JDBC plugin, make some microoptimizations and fixes:
* Fix the handling of TIMESTAMP when performing SELECT requests;
* Make functions for building WHERE statements static where possible to reduce the number of `InputData` checks. This is proposed by @sansanichfb;
* Prettify some SQL statements generated by the PXF JDBC plugin and change the tests respectively;
* Organize imports;
* Expand some intricate `if ... else` constructions.

Improve documentation:
* Clarify, expand or rewrite Javadoc strings;
* Rewrite README.md.


mebelousov:
This is great feature.
We really need to upload data from Greenplum to other databases.

deem0n:
Yep, our customers have plenty different data sources, mostly JDBC compliant. This feature makes GPDB a very flexible and friendly player in corporate IT landscape.

kongyew:
**This JDBC profile is not officially supported** by Greenplum since it is not tested at all.  Use it at your own risks.  The interface maybe changed in the future since the current profile is not securely storing credentials.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    /**
+     * openForRead() implementation
+     * Create query, open JDBC connection, execute query and store the result into resultSet
+     *
+     * @throws SQLException if a database access error occurs
+     * @throws SQLTimeoutException if a problem with the connection occurs
+     * @throws ParseException if th SQL statement provided in PXF InputData is incorrect
+     * @throws ClassNotFoundException if the superclass implementation disappeared
```
sansanichfb: I am wondering how likely it might happen?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,469 @@
...
+     * Create query template and open JDBC connection
+     *
+     * @throws SQLException if a database access error occurs
+     * @throws SQLTimeoutException if a problem with the connection occurs
+     * @throws ParseException if the SQL statement provided in PXF InputData is incorrect
+     * @throws ClassNotFoundException if the superclass implementation has disappeared
+     */
+    @Override
+    public boolean openForWrite() throws SQLException, SQLTimeoutException, ParseException, ClassNotFoundException {
+        if (statementWrite != null && !statementWrite.isClosed()) {
```
sansanichfb: Are we having the intention to allow multiple calls of this method? Shouldn't we throw an exception in this case?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,469 @@
...
+        super.openConnection();
+        if (dbMeta.supportsTransactions()) {
+            dbConn.setAutoCommit(false);
+        }
+
+        queryWrite = buildInsertQuery();
+        statementWrite = dbConn.prepareStatement(queryWrite);
+
+        if ((batchSize != 0) && (!dbMeta.supportsBatchUpdates())) {
+            LOG.info(
```
sansanichfb: IMO warning is more appropriate here since the user had an expectation and explicitly set batch set to non-zero.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,469 @@
...
+                case DATE:
+                    if (field.val == null) {
+                        statementWrite.setNull(i, Types.DATE);
+                    }
+                    else {
+                        statementWrite.setDate(i, (Date)field.val);
+                    }
+                    break;
+                default:
+                    throw new IOException("The data tuple from JdbcResolver is corrupted");
```
sansanichfb: This message is not very actionable. I'd be more specific and mention that certain data type is not supported.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,469 @@
...
+            // log exception as error, but do not throw it (the actual cause of rollback will be thrown instead)
+            LOG.error("An exception happened during the transaction rollback: '" + e.toString() + "'");
+        }
+        return rollbackException;
+    }
+
+
+    private static final Log LOG = LogFactory.getLog(JdbcAccessor.class);
+
+    private static final String failedInsertMessage = "Insert failed due to an SQLException. The target database does not support transactions and SOME DATA MAY HAVE BEEN INSERTED. ";
```
sansanichfb: This should be named as constant, something like `FAILED_INSERT_MESSAGE`.

sansanichfb:
Good job leskin-in! I left some comments. Also, I am wondering, which target database patch was tested against? The transactional approach doesn't seem clear to me - if DB doesn't support transactions - you are still sending data in batches, which might leave target db in the non-consistent state since you can't rollback the last transaction. Should we reconsider this approach and send only one batch(not ideal though, maybe ask users to tweak partitioning) in that case?

kapustor:
Hi @sansanichfb!

> Also, I am wondering, which target database patch was tested against? The transactional approach doesn't seem clear to me - if DB doesn't support transactions - you are still sending data in batches, which might leave target db in the non-consistent state since you can't rollback the last transaction. Should we reconsider this approach and send only one batch(not ideal though, maybe ask users to tweak partitioning) in that case?

We have tested the patch with Oracle, Postgres, MSSQL and Ignite cluster targets - they all work fine, the perfomance is good enough (especially for Ignite - about 200 Mbyte/s for two-node Greenplum and Ignite clusters).
If you know any other target that should be tested - pls let me know, I will do it.
As for transactional approach - even if the target do support transactions, there still may be inconsistense in data, for example if one of the GP segements will fail to insert it's batch (other segements may succeed and only part of the data will be inserted). As for sending all the data in one batch - as I understand, such approach will limit the max data size by memory available for PXF service, because PXF have to aggregate the batch in its memory.

pxf/pxf-jdbc/README.md
```diff
@@ -1,139 +1,235 @@
...
+
+Example partitions:
+* `&PARTITION_BY=id:int&RANGE=42:142&INTERVAL=2`
+* `&PARTITION_BY=createdate:date&RANGE=2008-01-01:2010-01-01&INTERVAL=1:month`
+* `&PARTITION_BY=grade:enum&RANGE=excellent:good:general:bad`
+
+
+### Mechanism
+
+Note that **by default PXF does not support more than 100 fragments**.
```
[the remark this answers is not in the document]
leskin-in: This was incorrect. I have tested this case; the mentioned limitation does not exist.

Many thanks for this comment!

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
leskin-in: This message is correct. If a certain data type is not supported, the exception must be thrown [in `JdbcResolver`](https://github.com/arenadata/incubator-hawq/blob/pxf_jdbc_writeAndFix/pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcResolver.java#L223).

sansanichfb:
@kapustor we can't fail safely when something wrong happens with at least one fragment, without dealing with the complexity of distributed transactions.
So transaction handling doesn't look necessary in this PR.
I would recommend to clearly state in the documents, that JDBC plugin doesn't guarantee a consistent outcome.
Furthermore, we should set clear expectations for users and ask them to use staging tables for inserting data from Greenplum. It would allow us to simplify code and get rid of cumbersome rollback logic.

WRT:
>  As for sending all the data in one batch - as I understand, such approach will limit the max data size by memory available for PXF service, because PXF have to aggregate the batch in its memory.
- this is not exactly accurate since PXF is designed to be a lightweight layer between external storage and consumer thus no aggregation in memory is needed(except some reasonable buffering, depending on I/O and network sockets).

kapustor:
Hi @sansanichfb 

We've removed the notification about the consistency and transactions from the error text.

Thank you!

kapustor:
Hi @sansanichfb,

Any news on this PR?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
leskin-in: Fixed: https://github.com/apache/incubator-hawq/pull/1353/commits/7fe4cac858a64ed5260d0e86342980f355fe7eb1#diff-6829beb5b62de0a4880b5d5dbebca91aR139

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
leskin-in: Fixed: https://github.com/apache/incubator-hawq/pull/1353/commits/7fe4cac858a64ed5260d0e86342980f355fe7eb1#diff-6829beb5b62de0a4880b5d5dbebca91aR70

leskin-in:
Hi @sansanichfb,
You have mentioned some questions that were not replied. I am a bit confused; could you clarify what exactly is unclear or should be corrected?

I have also left notes on each of the comments you've made in the code.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -0,0 +1,364 @@
...
+                default:
+                    throw new IOException("The data tuple from JdbcResolver is corrupted");
+            }
+        }
+    }
+
+    private static final Log LOG = LogFactory.getLog(JdbcResolver.class);
+
+    // SimpleDateFormat to parse TEXT into DATE
+    private static ThreadLocal<SimpleDateFormat> dateSDF = new ThreadLocal<SimpleDateFormat>() {
```
denalex: how are these threadlocals are getting cleaned ? if this logic executes on Tomcat thread pool, those threads are created and discarded as load changes, so when a thread is discarded, the thread local needs to be cleaned, which is the usual problem with threadlocals. Maybe use some thread-safe library for formatting, like joda-time ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        rows.add(row);
+    }
+
+    @Override
+    public boolean isCallRequired() {
+        if ((maxRowsCount > 0) && (rows.size() >= maxRowsCount)) {
+            return true;
+        }
+        return false;
```
denalex: would be better to call the method isBatchFull() and provide a 1-liner implementation:
```
return (maxRowsCount > 0) && (rows.size() >= maxRowsCount)
```

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    BatchWriterCallable(JdbcPlugin plugin, String query, PreparedStatement statement, int maxRowsCount) throws IllegalArgumentException {
+        if ((plugin == null) || (query == null)) {
+            throw new IllegalArgumentException("The provided JdbcPlugin or SQL query is null");
+        }
+        this.plugin = plugin;
+        this.query = query;
+        this.statement = statement;
+        if (maxRowsCount < 0) {
+            maxRowsCount = 0;
```
denalex: I am not sure if we should allow unlimited batch size. Would it be better to default it to some value, say 1000 ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/WriterCallableFactory.java
```diff
@@ -0,0 +1,108 @@
...
+        statement = null;
+    }
+
+    /**
+     * Get an instance of WriterCallable
+     *
+     * @return an implementation of WriterCallable, chosen based on parameters that were set for this factory
+     * @throws IllegalArgumentException if a WriterCallable cannot be created with requested parameters
+     */
+    public WriterCallable get() throws IllegalArgumentException {
```
denalex: do we need to explicitly declare throwing of runtime exceptions, such as IllegalArgumentException ? It's not like the caller can recover from them in these use cases.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
```diff
@@ -0,0 +1,437 @@
...
+                    throw e;
+                }
+            }
+        }
+
+        if (rollbackException != null) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Rollback is now required");
+            }
+            rollbackException = tryRollback(statementWrite.getConnection(), rollbackException);
```
denalex: we cannot guarantee transactional semantics for this profile anyways, as a single INSERT will be executed by multiple segments and multiple PXF threads, and there is no way for one of the threads to know the status of the others. It was suggested before to remove rollback logic altogether as it serves no practical purpose and just complicates the code and the processing logic. @leskin-in @kapustor -- What do you think ?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
```diff
@@ -177,11 +177,11 @@ under the License.
     </profile>
     <profile>
         <name>Jdbc</name>
-        <description>A profile for reading data into HAWQ via JDBC</description>
+        <description>A profile to access (read & write) data via JDBC</description>
```
denalex: the "&" character here breaks XML and the PXF server does not even startup, please remove it.

deem0n: or may be change `&` to `&amp;` ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
kapustor: 1000 for default is too small, for example regular value for Ignite DB is 25000. Maybe 10000?
Btw, what are disadvantages of unlimited batch?

pxf/pxf-service/src/main/resources/pxf-profiles-default.xml
denalex: this is supposed to be user readable, rather than machine readable, just substitute with word "and" to avoid issues.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
denalex: unlimited batch size might result in PXF running out of memory if the user forgot to set batch size in the table definition. For example, on a small cluster with 5 nodes and 5 PXF instances, exporting 50G dataset from the cluster via PXF might result in 10G+ data per PXF instance and by default PXF is only given 2G of memory.

going with smaller default batch size will ensure batching still works on systems with default memory setting (2G) . Users will redefine the batch size to achieve best performance, if needed.

Oracle recommends setting batch size to 50-100 (https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754), so maybe actually 100 is a better safe default number ?

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcAccessor.java
leskin-in: @denalex , you are right. 
The only scenario when `tryRollback()` works properly is one PXF segment, one thread, and an external database that supports transactions.
There are more than one PXF segment available in all cases, thus there is no guarantee that rollback will work properly, and therefore, rollback is of no use.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
leskin-in: Done. The default number is set by `WriterCallableFactory`: https://github.com/apache/incubator-hawq/pull/1353/commits/52e9f7f2050b6e200e8e9cdd1dba2a7f22c31923#diff-b21204dffcdf144d4d2c5203f2b9a8b5R84

denalex: thanks you for the changes. Should we still remove the case of infinite batching and error out if user sets BATCH_SIZE < 0 ? This is dangerous logic, and if a given use case supports huge batches, let users define tables with appropriate batch sizes.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/WriterCallableFactory.java
```diff
@@ -75,16 +75,15 @@ public void setQuery(String query) {
...
      */
     public void setBatchSize(int batchSize) {
-        if (batchSize < 0) {
-            batchSize = 0;
-        }
-        else if (batchSize == 0) {
-            batchSize = 1;
+        if (batchSize == 0) {
+            // Set the recommended value: https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754
+            batchSize = 100;
```
denalex: this is counter-intuitive, a good guideline is to never change the meaning of the user-set parameter. I would propose the following -- user provides value of BATCH_SIZE:
1. nothing (not provided) -- use default of 100
2. 0 -- error out -- batch size of 0 does not make sense
3. 1 -- error out -- batch size of 1 does not make sense
4.  < 1 -- error out -- batch size of <1 does not make sense
5. > 1 and < 65k -- use the number provided by the user
6. > 65k -- error out -- too large for practical purposes ?

leskin-in: This makes sense.
Do you propose to use batches by default? Currently, `BATCH_SIZE` controls not only the size of a batch, but also whether it is used or not.
Also, if we allow user not to use batches, we should make `1` (or some other value) a valid `BATCH_SIZE`.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcResolver.java
leskin-in: @denalex, I would be glad if you could tell about the details of a `ThreadLocal`'s lifecycle  in Tomcat. As far as I know (see https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html), every `ThreadLocal` object is a subject to garbage collection when a thread dies.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/WriterCallableFactory.java
denalex: oh, thanks, got it, I thought batch on/off was a separate property. If we want to stick with a single property (and keep it simple for users), then I think using batching by default (with a default value) is a good idea (if the backend db supports it). This way, though, we will have to let users to turn it off for a query, which they will achieve by setting the value to 0 (batch feature is off) or 1 (batch with size 1 practically means no batching, implementation can take care of that). I think this is closer to your original design, except I would take out negative values to prohibit unlimited size, as well as maybe put an upper bound on batch size to prevent crashing PXF with OOM.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcResolver.java
denalex: thanks, right, looking more into details of ThreadLocal implementation, it seems there should be no issue with GC when Thread finishes. We also don't store request-specific state there, so there is no concern of cleaning it when thread is returned to the pool. So we can consider this question closed, and we can move to thread-safe formatters as further improvement later on.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/WriterCallableFactory.java
kapustor: As I understand, now optimal parameter values look like this:
```
1. nothing (not provided) -- use default of 100
2. 0 -- no batching
3. 1 -- no batching
4. >1 -- use the number provided by the user
```
In my opinion, we shouldn't limit possible `BATCH_SIZE` max value (or we should set it very high, to 500k for example) as we dont know the usecases and external DBs that users will use it with. It is possible that there are some DBs or other systems that use high batch size for optimal inserts.
@denalex what do you think about `BATCH_SIZE` max value?

denalex: @kapustor -- agree, limiting max is superficial, as with a given memory there might be a value below max that would cause OOM and there is no other reasons to limit the max. So, let's not limit the max and just use a value provided by the user.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/writercallable/BatchWriterCallable.java
```diff
@@ -88,24 +88,41 @@ public SQLException call() throws IOException, SQLException, ClassNotFoundExcept
...
+    BatchWriterCallable(JdbcPlugin plugin, String query, PreparedStatement statement, int batchSize) {
+        if (plugin == null || query == null) {
             throw new IllegalArgumentException("The provided JdbcPlugin or SQL query is null");
         }
+
+        try {
+            if (!plugin.getConnection().getMetaData().supportsBatchUpdates()) {
+                throw new IllegalArgumentException("The external database does not support batch updates");
+            }
+        }
```
divyabhargov: Since batching is default, it is risky to have this failure happen. We should make a decision to use SimpleWriterCallable much before this. Basically, the decision to use batching should depend upon both the batch size specified (or implied by default) and if the external database supports batching or not.

pxf/pxf-jdbc/README.md
```diff
@@ -114,14 +114,13 @@ The `PARTITION_BY`, `RANGE` and `INTERVAL` parameters in such tables are ignored
 INSERT queries can be batched. This may significantly increase perfomance if batching is supported by an external database.
 
 To enable batching, create an external table with the parameter `BATCH_SIZE` set to:
-* `0`. Batch will be of [recommended](https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754) size (`100`);
-* `1`. Batching will not be used;
-* `integer > 1`. Batches of the given size will be used;
-* `integer < 0`. A batch of infinite size will be used (all tuples will be sent in one huge JDBC query from each segment). Choosing this value **may cause errors**, as PXF will consume lots of memory; also, every database has a limit on the maximum size of JDBC queries.
+* `integer < 1`. Batch will be of [recommended](https://docs.oracle.com/cd/E11882_01/java.112/e16548/oraperf.htm#JJDBC28754) size (`100`);
```
divyabhargov: Changing the batch size value after it is being explicitly specified by the user seems wrong. I suggest we should throw an exception if the specified batch size is a negative integer. We should use the default of 100 only if the batch size was never specified. Both 0, and 1 should result in no batching. Right now, batch size 0 will result as batch size 100, which is not what we want. Here is the desirable chart:

If external database supports batching:
```
* Not specified: Batch will be of recommended size(100)
* 0 or 1: Do not use batching
* > 1 : Use the specified size
* Anything else: Error out that the batch size is invalid
```

If the external database does not support batching:
```
* Not specified: Do not use batching
* 0 or 1: Do not use batching
* > 1 : Error out stating the external database does not support batching
* Anything else: Error out that the batch size is invalid
```

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcResolver.java
```diff
@@ -0,0 +1,364 @@
...
+            }
+
+            if (LOG.isDebugEnabled()) {
+                if (DataType.get(oneField.type) == DataType.BYTEA) {
+                    LOG.debug("OneField content (conversion from BYTEA): '" + new String((byte[])oneField.val) + "'");
+                }
+            }
+
+            // Convert TEXT columns into native data types
+            if ((oneField.val != null) && (DataType.get(oneField.type) == DataType.TEXT) && (DataType.get(column.columnTypeCode()) != DataType.TEXT)) {
```
divyabhargov: This bit failed for us when we had a timestamp or date field that had a NULL value. The type reported by GPDBWritable was not being converted from TEXT to timestamp, because conversion of types would only apply if the value were not NULL. Even if the value is null, we want oneField's type to be the column type code and not string. 

Here's the snippet that worked for us:
```
            if (LOG.isDebugEnabled()) {
                if (DataType.get(oneField.type) == DataType.BYTEA) {
                    LOG.debug("OneField content (conversion from BYTEA): '" + new String((byte[])oneField.val) + "'");
                }
            }
            // convert values of non TEXT types represented as text by the GPDBWritable serializer into native types
            if ((DataType.get(oneField.type) == DataType.TEXT) && (DataType.get(column.columnTypeCode()) != DataType.TEXT)) {
                // convert type information
                oneField.type = column.columnTypeCode();

                // Convert values if they are present (not null)
                if ((oneField.val != null)) {
                    String rawVal = (String)oneField.val;
                    if (LOG.isDebugEnabled()) {
                        LOG.debug("OneField content (conversion from TEXT): '" + rawVal + "'");
                    }
        ...
```
and remove
```
oneField.type = column.columnTypeCode();
```
from the bottom of the `if` statement.

pxf/pxf-jdbc/src/main/java/org/apache/hawq/pxf/plugins/jdbc/JdbcPlugin.java
```diff
@@ -19,95 +19,207 @@
...
+        user = input.getUserProperty("USER");
+        if (user != null) {
+            pass = input.getUserProperty("PASS");
+        }
+
+        // This parameter is not required. The default value is 0
+        String batchSizeRaw = input.getUserProperty("BATCH_SIZE");
+        if (batchSizeRaw != null) {
+            try {
+                batchSize = Integer.parseUnsignedInt(batchSizeRaw);
```
denalex: this breaks compilation on JDK 1.7, since the API is only available in 1.8. I will make a trivial change to fix this before merging.

PR #1356, opened 2018-04-26
HAWQ-1611. refactor the vtype in order to advance the performance, Most of the c…

zhangjackey:
…odes are written by Wang Weinan


contrib/vexecutor/tuplebatch.c
```diff
@@ -93,91 +85,119 @@ tbSerializationSize(TupleBatch tb)
...
 
     //deserial member value -- datagroup
     if(tb->ncols != 0)
     {
-        tb->datagroup = palloc0(sizeof(vheader*) * tb->ncols);
         int colid;
-        while (len < buflen)
+        tmplen = sizeof(vtype*) * tb->ncols;
+        tb->datagroup = palloc0(tmplen);
+        while (((len + 0x8) & (~0x7)) < buflen)
```
wengyanqing: What's the meaning of these magic number ? It's better to have a comment or use a literal const.

linwen:
+1

contrib/vexecutor/tuplebatch.c
zhangjackey: add comment for the buffer length alignment.

PR #1358, opened 2018-04-27
HAWQ-1609. Implement Vectorized Motion Node

weinan003:
1. add vectorized motion node
2. Tuplebatch structure "create on used" in deserialization function, so that TB used memory reduce the allocation times.
3. change ExecVProcNode function return value type as boolean to indicate if the hook function takes responsibility for the execution process.
4. export some static function for vectorized library

New test code has not been created as previous feature test can cover this patch

### DIFF
--- a/contrib/vexecutor/nodeVMotion.c
+++ b/contrib/vexecutor/nodeVMotion.c
@@ -1,0 +1,410 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+#include "nodeVMotion.h"
+#include "tuplebatch.h"
+#include "execVQual.h"
+/*=========================================================================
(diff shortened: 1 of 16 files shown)


contrib/vexecutor/nodeVMotion.c
```diff
@@ -0,0 +1,410 @@
...
+
+        Assert(!motion->sendSorted);
+
+        tuple = execVMotionUnsortedReceiver(node);
+
+        if (tuple == NULL)
+            node->ps.state->active_recv_id = -1;
+        else
+        {
+            Gpmon_M_Incr(GpmonPktFromMotionState(node), GPMON_QEXEC_M_ROWSIN);
```
wengyanqing: Is gpmon info different between tuple and tuplebatch in motion operation ?

contrib/vexecutor/nodeVMotion.c
```diff
@@ -0,0 +1,410 @@
...
+             * targetRoute setup that we could possibly send to.  So we
+             * can cheat and just fix the targetRoute to 0 (the 1st
+             * route).
+             */
+            targetRoute = 0;
+        }
+    }
+    else if (motion->motionType == MOTIONTYPE_HASH) /* Redistribute */
+    {
+        //TODO:: Implement later
```
wengyanqing: It's better to error out here than just ignore this condition.

contrib/vexecutor/tuplebatch.c
```diff
@@ -150,19 +150,26 @@ tbSerialization(TupleBatch tb )
     return ret;
 }
 
-TupleBatch tbDeserialization(unsigned char *buffer)
+void tbDeserialization(unsigned char *buffer,TupleBatch* pTB )
```
wengyanqing: Is tuple batch deserialization always successful ? If not, please use bool for return value type.

contrib/vexecutor/nodeVMotion.c
weinan003: VE conditional check in init hook close "vmotion" when it is redistribution type. I literally leave a room for next patch

contrib/vexecutor/nodeVMotion.c
weinan003: Gpmon in VE model records tuplebatch instead of per tuple

PR #1359, opened 2018-05-04
HAWQ-1609. Update regression test configuration

weinan003:
This PR's propose for adapting regression test for our new feature including VE and runtime filter
**Test:**
fix regression test failure, do not involve new test


radarwave:
LGTM.

PR #1360, opened 2018-05-06
HAWQ-1607. This commit implements applying Bloom filter during Scan outer table

linwen:
1. Pash down Bloom filter structure to outer table scan(only support parquet);
    2. Check if the tuple from outer table is found in Bloom filter structure.
    3. Add a GUC hawq_hashjoin_bloomfilter_sampling_number. This guc value controls the Bloom filter sampling number, while scanning outer table, for first N tuples of the outer table, if the ratio is larger than hawq_hashjoin_bloomfilter_ratio, the remain tuples will not be checked by Bloom filter.
    4. If there is any expression on outer join keys except T_Var(projection), such as, fact.c1 + 1 = dim.c1. 2, if there are multiple join keys, e.g. fact.c1 = dim.c1 and fact.c2 = dim.c2, Bloomfilter won't be created. Since these cases invloves pushing down expression and project information to scan, which will be implemented later.

Please review, thanks!

### DIFF
--- a/src/backend/executor/nodeHashjoin.c
+++ b/src/backend/executor/nodeHashjoin.c
@@ -867,8 +867,9 @@ ExecHashJoinOuterGetTuple(PlanState *outerNode,
 	 * Adaptive runtime filter check.
 	 */
 	HashJoin *hj = (HashJoin*)hjstate->js.ps.plan;
-	if (rf != NULL && rf->hasRuntimeFilter && !rf->stopRuntimeFilter && !rf->checkedSamples &&
-			rf->bloomfilter->nTested >= hawq_hashjoin_bloomfilter_sampling_number)
+	if (outerNode->type == T_TableScanState && rf != NULL && rf->hasRuntimeFilter
+			&& !rf->stopRuntimeFilter && !rf->checkedSamples
+			&& rf->bloomfilter->nTested >= hawq_hashjoin_bloomfilter_sampling_number)
 	{
 		double real_ratio = (rf->bloomfilter->nMatched) / (rf->bloomfilter->nTested);
 		if(real_ratio > hawq_hashjoin_bloomfilter_ratio)
--- a/src/backend/executor/nodeTableScan.c
+++ b/src/backend/executor/nodeTableScan.c
@@ -100,6 +100,7 @@ ExecEndTableScan(TableScanState *node)
 	if (node->ss.runtimeFilter != NULL)
 	{
 		FreeScanRuntimefilterState(node->ss.runtimeFilter);
+		node->ss.runtimeFilter = NULL;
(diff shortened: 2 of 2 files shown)


linwen:
This commit doesn't contain test cases, test cases will be added with HAWQ-1608. After finish HAWQ-1608, users can use "explain analyze" statement to know if the Bloom filter is used for hash join.

src/include/nodes/execnodes.h
```diff
@@ -1522,6 +1540,9 @@ typedef struct ScanState
 	/* The type of the table that is being scanned */
 	TableType tableType;
 
+	/* Runtime filter */
+	struct RuntimeFilterState runtimeFilter;
```
wengyanqing: Since ScanState need to be copied some times, it's better to use a point of RuntimeFilterState in the struct and allocate memory dynamically.

src/backend/executor/nodeHashjoin.c
```diff
@@ -776,6 +802,33 @@ ExecEndHashJoin(HashJoinState *node)
...
+		ExprState  *keyexpr = (ExprState *) lfirst(hk);
+		Var *variable = (Var *) keyexpr->expr;
+		rf->joinkeys = lappend_int(rf->joinkeys, variable->varattno);
+		i++;
+	}
+	rf->hashfunctions = (FmgrInfo *) palloc(i * sizeof(FmgrInfo));
+	memcpy(rf->hashfunctions, hjstate->hj_HashTable->hashfunctions, i*sizeof(FmgrInfo));
+	size_t size = offsetof(BloomFilterData, data) + hjstate->hj_HashTable->bloomfilter->data_size;
+	rf->bloomfilter = palloc0(size);
+	memcpy(rf->bloomfilter, hjstate->hj_HashTable->bloomfilter, size);
```
wengyanqing: Why not just assign hjstate->hj_HashTable->bloomfilter to rf->bloomfilter ?

src/backend/cdb/cdbparquetrowgroup.c
```diff
@@ -206,54 +207,54 @@ ParquetRowGroupReader_ScanNextTuple(
...
 
-		ParquetColumnReader *nextReader =
-			&rowGroupReader->columnReaders[colReaderIndex];
-		int hawqTypeID = tupDesc->attrs[i]->atttypid;
+		Datum *values = slot_get_values(slot);
+		bool *nulls = slot_get_isnull(slot);
 
-		if(hawqAttrToParquetColNum[i] == 1)
+		int colReaderIndex = 0;
+		int16 proj[128];
```
wengyanqing: It's better to use natts instead of 128.

PR #1361, opened 2018-05-08
HAWQ-1612. Implement vectorized aggregate functions, such as sum/count/avg.

zhangjackey:


contrib/vexecutor/vagg.h
```diff
@@ -0,0 +1,94 @@
+#ifndef VAGG_H
+#define VAGG_H
+
+#include "executor/execHHashagg.h"
+
+/* batch hashagg group linklist header */
+typedef struct GroupData {
+	HashAggEntry *entry;	// pointer to agg_hash_entry
+	int idx; 				// pointer to idx_list
```
interma: Refine my poor comments:
1. use /**/ instead //.
2. unify the ident.

contrib/vexecutor/vexecutor.c
```diff
@@ -117,7 +122,7 @@ static void backportTupleDescriptor(PlanState* ps,TupleDesc td)
 		tuple = caql_getnext(pcqCtx);
 
 		if (!HeapTupleIsValid(tuple))
-			elog(ERROR, "cache lookup failed for type %u", oidtypeid);
+			continue;
```
interma: Why use `continue` instead of `elog ERROR` here?

contrib/vexecutor/vagg.h
```diff
@@ -0,0 +1,94 @@
...
+} GroupData;
+
+/* batch hashagg group data */
+typedef struct BatchAggGroupData{
+	GroupData 	group_header[1024];	//group linklist header
```
interma: 1024, use a macro is better.

weinan003: variable **BATCHSIZE** in *vexecutor.c* indicate the number of slots the TupleBatch hold.

contrib/vexecutor/vexecutor.c
zhangjackey: For the grouping aggregation, there may generate the tuple descriptor like this: {vtype,  vtype,  non-vtype, non-vtype}, we have to convert the vtype to non-vtype and skip the non-vtype which already exists.

PR #1368, opened 2018-05-25
HAWQ-1616. Fix the wrong result of hash join when enable Bloom filter.

linwen:
The projection information of join keys hasn't been pushed down to parquet scan correctly. It causes computing a wrong hash value during parquet scan.


linwen:
merged.

PR #1369, opened 2018-05-25
HAWQ-1617. Incorrect processing of boolean operators in pushdown

leskin-in:
When accessing external tables, the pushdown feature sometimes works incorrect. Consider the following query:
```
SELECT * FROM table_ex WHERE bool1=false AND id1=60003;
```
When this query is executed, an error "stack is not empty ..." happens.

Turns out that such query is transformed into a list of three items that represents the constraints: `BoolExpr`, `Var` and `OpExpr` (this is equal to a query `WHERE NOT(bool1 = true) AND (id1 = 60003)`). Note that the list does not contain (implicit) AND operators.

Then, the list is processed by a [piece of code in pxffilters.c](https://github.com/apache/incubator-hawq/blob/master/src/backend/access/external/pxffilters.c#L1259), where there is a check of presence of `BoolExpr`. If expression items of this kind are detected, no implicit AND operators are added.

In the case described, one 'BoolExpr' element is present, and no implicit AND operators are added. This leads to the error mentioned above.

This commit changes the signatures of `enrich_trivial_expression()` and `pxf_free_expression_items_list()` from [pxffilters.c](https://github.com/apache/incubator-hawq/blob/master/src/backend/access/external/pxffilters.c) in order to fix the bug:
* A number of impicit AND operators to be added is now passed to `add_extra_and_expression_items()`
* `add_extra_and_expression_items()` now requires the pointer to pointer to `Node` object to store the pointer to an expression item for the implicit AND that it creates (before, it was stored in the expression item itself; however, due to the presence of other logical operators, we cannot rely on this mechanism)
* `pxf_free_expression_items_list()` simplifies, due to the changes in the way how the pointer to `Node` object is stored

The current mechanism of implicit AND expressions addition works well only while the
OR operators are not supported (or not present in a query). When the OR operators appear, the implicit ANDs must be added in different parts of a query, not only to the end, as done now.

### DIFF
--- a/src/backend/access/external/pxffilters.c
+++ b/src/backend/access/external/pxffilters.c
@@ -31,7 +31,7 @@
 #include "utils/guc.h"
 #include "utils/lsyscache.h"
 
-static List* pxf_make_expression_items_list(List *quals, Node *parent, int *logicalOpsNum);
+static List* pxf_make_expression_items_list(List *quals, Node *parent);
 static void pxf_free_filter(PxfFilterDesc* filter);
 static char* pxf_serialize_filter_list(List *filters);
 static bool opexpr_to_pxffilter(OpExpr *expr, PxfFilterDesc *filter);
@@ -43,7 +43,7 @@ static bool supported_operator_type_scalar_array_op_expr(Oid type, PxfFilterDesc
 static void scalar_const_to_str(Const *constval, StringInfo buf);
 static void list_const_to_str(Const *constval, StringInfo buf);
 static List* append_attr_from_var(Var* var, List* attrs);
-static void add_extra_and_expression_items(List *expressionItems, int extraAndOperatorsNum, BoolExpr **extraNodePointer);
+static void add_extra_and_expression_items(List *expressionItems, int extraAndOperatorsNum);
 static List* get_attrs_from_expr(Expr *expr, bool* expressionIsSupported);
 
 /*
@@ -298,14 +298,15 @@ pxf_free_expression_items_list(List *expressionItems)
  *
  */
 static List *
-pxf_make_expression_items_list(List *quals, Node *parent, int *logicalOpsNum)
+pxf_make_expression_items_list(List *quals, Node *parent)
 {
 	ExpressionItem *expressionItem = NULL;
 	List			*result = NIL;
 	ListCell		*lc = NULL;
 	ListCell		*ilc = NULL;
+	int				quals_size = list_length(quals);
 
-	if (list_length(quals) == 0)
+	if (quals_size == 0)
 		return NIL;
 
 	foreach (lc, quals)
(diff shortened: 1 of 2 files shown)


src/backend/access/external/pxffilters.c
```diff
@@ -1238,64 +1236,77 @@ list_const_to_str(Const *constval, StringInfo buf)
...
- * Takes list of expression items which supposed to be just a list of OpExpr
- * and adds needed number of AND items
- *
+ * Adds a given number of AND expression items to an existing list of expression items
+ * Note that this will not work if OR operators are introduced
  */
-void enrich_trivial_expression(List *expressionItems) {
-
+void
+add_extra_and_expression_items(List *expressionItems, int extraAndOperatorsNum, BoolExpr **extraNodePointer)
```
benchristel: It's probably worth documenting the meaning and intended usage of these parameters in a comment. E.g. it would be nice to clarify that `extraNodePointer` is an out param.

shivzone:
@leskin-in based on our discussion in the greenplum-pxf thread. Please apply the same patch https://github.com/greenplum-db/gpdb/pull/5470 to fix the issue for HAWQ as well.
If you don't have the bandwidth, we can follow up on this HAWQ PR ourselves.

src/backend/access/external/pxffilters.c
```diff
@@ -1238,64 +1236,77 @@ list_const_to_str(Const *constval, StringInfo buf)
...
+{
+	if ((!expressionItems) || (extraAndOperatorsNum < 1))
```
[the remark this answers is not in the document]
leskin-in: Done

src/backend/access/external/pxffilters.c
leskin-in: The last parameter was removed; that is a solution from https://github.com/greenplum-db/gpdb/pull/5470. The other parameters are quite straightforward.

leskin-in:
Update according to https://github.com/greenplum-db/gpdb/commit/3ac93fb4259436c87b6c1705bdb05c003ca93423.

shivzone:
Thanks for the quick turnaround !

shivzone:
Merged to master. @leskin-in close the PR

radarwave:
Please check the unittest failures of those commits:

`pxffilters_test.c: In function ‘test__pxf_serialize_filter_list__oneFilter’:
pxffilters_test.c:817:2: error: too many arguments to function ‘pxf_free_expression_items_list’
  pxf_free_expression_items_list(expressionItems, true);
  ^
In file included from pxffilters_test.c:26:0:
../pxffilters.c:267:1: note: declared here
 pxf_free_expression_items_list(List *expressionItems)
 ^
pxffilters_test.c: In function ‘test__pxf_serialize_fillter_list__nullFilter’:
pxffilters_test.c:835:2: error: too many arguments to function ‘pxf_free_expression_items_list’
  pxf_free_expression_items_list(expressionItems, true);
  ^
In file included from pxffilters_test.c:26:0:
../pxffilters.c:267:1: note: declared here
 pxf_free_expression_items_list(List *expressionItems)
 ^
pxffilters_test.c: In function ‘test__pxf_serialize_filter_list__manyFilters’:
pxffilters_test.c:873:2: error: too many arguments to function ‘pxf_free_expression_items_list’
  pxf_free_expression_items_list(expressionItems, true);
  ^
In file included from pxffilters_test.c:26:0:
../pxffilters.c:267:1: note: declared here
 pxf_free_expression_items_list(List *expressionItems)
 ^
make[3]: *** [pxffilters_test.o] Error 1
make[3]: *** Waiting for unfinished jobs....`

leskin-in:
@radarwave thank you! The necessary corrections have been made.

radarwave:
Still seeing one unittest failure, please make sure unittests are passed on centos7 platform.

> [ RUN         ] test__distribute_work_to_gp_segments__big_cluster_many_active_nodes
pxffilters_test.o: In function `test__pxf_serialize_filter_list__manyFilters':
/tmp/build/90177b2f/hdb_apache/src/backend/access/external/test/pxffilters_test.c:869: undefined reference to `enrich_trivial_expression'
collect2: error: ld returned 1 exit status
make[3]: *** [pxffilters.t] Error 1
make[3]: *** Waiting for unfinished jobs....

radarwave:
Now looks good, merged.

PR #1373, opened 2018-05-31
HAWQ-1618. Segment panic at workfile_mgr_close_file() when transaction

interma:
ROLLBACK

**Fix what?**
Segment panic sometimes when transaction rollback.
Core at at `workfile_mgr_close_file() `

**How to fix?**
Workfile memory is alloced in `ExecWorkFile_Create()/ExecWorkFile_Open()`'s current memory context, and this context will be released early in some cases, then segment panic in `ntuplestore_cleanup()`.
So should use a appropriate memory context to hold them: its life is the same as `NTupleStore`.

This fix borrow from gpdb:
https://github.com/greenplum-db/gpdb/commit/f0a0a593bde8cf0c9b9bbc79061a09f7164b54f7#diff-b37908413e016b54a45d549cf0539121

**Test?**
No test, reason:
It's a memory bug, need a appropriate time to cancel query and accessing a recycled memory context doesn't always trigger panic, hard to test them.


linwen:
+1

radarwave:
+1

interma:
merged

PR #1376, opened 2018-06-12
HAWQ-1624. Change libhdfs3 to be ABI compatible with libhdfs

superbobry:
I have been working on this in the context of tensorflow/tensorflow#16919, and the three discrepancies I've identified are:

* libhdfs3 does not define `hdfsPread` which nonetheless can be implemented in terms of `hdfsTell`, `hdfsSeek` and `hdfsRead`;
* `hdfsSync` is called `hdfsHSync` in libhdfs3;
* In libhdfs `hdfsFlush` flushes the internal buffer of the `BufferedInputStream`; in libhdfs3 it always flushes to HDFS, i.e. it does what `hdfsHFlush` should do. One way to address this is to make hdfsFlush a noop, but I guess it's already too late to change the implementation, wdyt?

Do you want me to add a test for `hdfsPread`? It does not do much on its own, but I can add one nonetheless.

### DIFF
--- a/depends/libhdfs3/src/client/Hdfs.cpp
+++ b/depends/libhdfs3/src/client/Hdfs.cpp
@@ -35,7 +35,6 @@
 #include "Thread.h"
 #include "XmlConfig.h"
 
-#include <boost/scope_exit.hpp>
 #include <vector>
 #include <string>
 #include <libxml/uri.h>
@@ -792,19 +791,20 @@ tSize hdfsRead(hdfsFS fs, hdfsFile file, void * buffer, tSize length) {
     return -1;
 }
 
-tSize hdfsPread(hdfsFS fs, hdfsFile file, tOffset position,
+tSize hdfsPread(hdfsFS fs, hdfsFile file, tOffset offset,
                 void * buffer, tSize length) {
-    PARAMETER_ASSERT(fs && file && position >= 0 && buffer && length > 0, -1, EINVAL);
+    PARAMETER_ASSERT(fs && file && offset >= 0 && buffer && length > 0, -1, EINVAL);
     PARAMETER_ASSERT(file->isInput(), -1, EINVAL);
 
+    tOffset oldPosition = -1;
+    tSize result = -1;
     InputStream & is = file->getInputStream();
     try {
-        tOffset oldPosition = is.tell();
-        is.seek(position);
-        BOOST_SCOPE_EXIT(is, oldPosition) {
-            is.seek(oldPosition);
-        } BOOST_SCOPE_EXIT_END
-        return is.read(static_cast<char *>(buffer), length);
+        oldPosition = is.tell();
+        is.seek(offset);
+        result = is.read(static_cast<char *>(buffer), length);
+    } catch (const Hdfs::HdfsEndOfStream & e) {
+        result = 0;
     } catch (const std::bad_alloc & e) {
         SetErrorMessage("Out of memory");
         errno = ENOMEM;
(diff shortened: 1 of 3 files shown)


amyrazz44:
Please create [Apache HAWQ JIRA](https://issues.apache.org/jira/projects/HAWQ/summary)  for your commit and add the corresponding test for your code changes. Thanks.

depends/libhdfs3/src/client/Hdfs.cpp
```diff
@@ -35,6 +35,7 @@
 #include "Thread.h"
 #include "XmlConfig.h"
 
+#include <boost/scope_exit.hpp>
```
wangzw: Do not make boost as hard requirement. libhdfs3 only needs boost if you are using an old version compiler such as GCC 4.4.

If the function provided by boost is not in the standard c++ library, do not use such function.

superbobry: Removed. Which C++ version do you target for libhdfs3? 

Also, just out of curiosity where does the compiler version constraint come from? CentOS 6?

superbobry:
@amyrazz44, @wangzw thanks for the feedback! Could you kindly comment on my concern regarding the behaviour of `hdfsSync`?

depends/libhdfs3/src/client/Hdfs.cpp
wangzw: I did not means the compiler version constrain, just use c++ standard library first, and use boost as an alternative.

interma:
@superbobry Thanks for your PR and sorry for response late.

* hdfsPread, LGTM, thanks for your tests.
* hdfsSync, LGTM
Looks you just added a new `hdfsHSync()` function and didn't change original `hdfsSync()` behavior, so it works with HAWQ.
* hdfsFlush
Do you mean you want to modify the `hdfsFlush()` behavior? I greped code, HAWQ just uses `hdfsHFlush()` by now. So you can feel free to modify `hdfsFlush()`, but make sure pass all tests (run all tests is a hard thing for new committer, if your modification is simple, I can run this test for you).

superbobry:
Hi @interma, thank you for the review. 

Regarding hdfsFlush: what I meant was how to proceed with making its behaviour compatible with libhdfs, i.e. removing the H part from the implementation. Even though HAWQ does not rely on hdfsFlush, other users of the library could, in theory, assume that hdfsFlush is hdfsHFlush. What is the usual policy for dealing with such potentially breaking changes?

interma:
@superbobry 
As I known, no/not many person declare that they depend on libhdfs3. In the past, they just make a copy into their repo if need it (not a good things, but understandable, maybe making all hawq tests passed is a big burden for them). 
So I think breaking changes are not a big deal in libhdfs3, if guys found something broken, hope them to back to HAWQ community.

superbobry:
@interma this is OK with me. Do you think the PR is good to merge?

interma:
@superbobry 
Code review is ok, I need to do more tests to verify your modification. 

I will @ you if all tests pass (or provide failure massages so that you can fix them), I plan to finish all tests and reply to you in the first half of next week. 

Wait some moments, Thanks.

interma:
@superbobry all tests passed, cool!

@amyrazz44 @wengyanqing for another +1.
Thanks.

superbobry:
I'd appreciate if you guys could also have a look at my question in the JIRA ticket on the inefficiency of hdfsPread.

interma:
@superbobry 
Merged, please close this PR, thanks.

PR #1377, opened 2018-06-21
HAWQ-1627. Support setting the max protocol message size when talking

interma:
with HDFS

**Fix what?**
User can set the max protocol message size, e.g.
In _etc/hdfs-client.xml_:
```
	<property>
		<name>ipc.maximum.data.length</name>
		<value>67108864</value>
	</property>
```

**How to fix?**
Use `CodedInputStream.SetTotalBytesLimit()` set the max size.

**Test?**
No test, reason: 
It can be covered by former tests.

### DIFF
--- a/depends/libhdfs3/src/rpc/RpcChannel.cpp
+++ b/depends/libhdfs3/src/rpc/RpcChannel.cpp
@@ -33,6 +33,7 @@
 #include "WriteBuffer.h"
 
 #include <google/protobuf/io/coded_stream.h>
+#include <google/protobuf/io/zero_copy_stream_impl_lite.h>
 
 #define RPC_HEADER_MAGIC "hrpc"
 #define RPC_HEADER_VERSION 9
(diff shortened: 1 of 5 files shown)


depends/libhdfs3/src/rpc/RpcChannel.cpp
```diff
@@ -768,7 +771,15 @@ void RpcChannelImpl::readOneResponse(bool writeLock) {
...
 
-    if (!curRespHeader.ParseFromArray(&buffer[0], headerSize)) {
+    // use CodedInputStream around the buffer, so we can set TotalBytesLimit on it
+    ArrayInputStream ais(&buffer[0], headerSize);
+    CodedInputStream cis(&ais);
+    cis.SetTotalBytesLimit(maxLength, maxLength/2);
+
+    // use ParseFromCodedStream instead of ParseFromArray, so it can consume the above CodedInputStream
+    //
+    // if just use ParseFromArray, we have on chance to set TotalBytesLimit (64MB default)
```
linwen: "on" should be "no"?

interma: Thanks! I will fix.

linwen:
+1

interma:
@amyrazz44 @wengyanqing help to review, thanks.

amyrazz44:
LGTM +1

interma:
Merged

PR #1379, opened 2018-06-29
HAWQ-1622. Cache PXF proxy UGI so that cleanup of FileSystem cache doesn't have to be done on each request

lavjain:


benchristel:
What's our approach to testing this going to be? Also, the cache should be a separate class IMO since its behavior and data don't really cohere with `SecurityServletFilter`. Something like:

```java
class UserGroupInformationCache {
  public UserGroupInformation get(SegmentTransactionId session) { /* ... */ }
  public void removeStaleUGIs(long currentTimeMillis)  { /* ... */ }
}
```

This interface assumes we will have some other thread outside this class that periodically calls `removeStaleUGIs` with the current time. We could also have the cache start its own cleanup thread, but then we'd have to `sleep` in our tests to observe the UGIs getting cleaned up, which could make the tests slow and flaky.

shivzone:
@lavjain even if this is WIP please create a jira and provide some context for the UGI cache

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -42,8 +52,51 @@
 
     private static final Log LOG = LogFactory.getLog(SecurityServletFilter.class);
     private static final String USER_HEADER = "X-GP-USER";
-    private static final String MISSING_HEADER_ERROR = String.format("Header %s is missing in the request", USER_HEADER);
-    private static final String EMPTY_HEADER_ERROR = String.format("Header %s is empty in the request", USER_HEADER);
+    private static final String SEGMENT_INDEX_HEADER = "X-GP-SEGMENT-ID";
+    private static final String TRANSACTION_ID_HEADER = "X-GP-XID";
+    private static final String MISSING_HEADER_ERROR = "Header %s is missing in the request";
+    private static final String EMPTY_HEADER_ERROR = "Header %s is empty in the request";
+    private static Map<SegmentTransactionId, TimedProxyUGI> cache = new ConcurrentHashMap<>();
```
shivzone: Why not use apache commons JCS or guava's caching library as opposed to implementing it ourselves with timer ?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -42,8 +52,51 @@
...
+    private static DelayQueue<TimedProxyUGI> delayQueue = new DelayQueue<>();
+    private static long UGI_CACHE_EXPIRY = 15 * 60 * 1000L; // 15 Minutes
```
shivzone: Instead consume this from pxf-site and make it configurable.
{code}
Configuration config = new Configuration();
config.addResource("pxf-site.xml");
{code}

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -42,8 +52,51 @@
...
+            return Long.compare(this.getDelayMillis(), ((TimedProxyUGI)other).getDelayMillis());
+        }
+
+        private long getDelayMillis() {
+            return (startTime + UGI_CACHE_EXPIRY) - System.currentTimeMillis();
+        }
+    }
+
+    private static class SegmentTransactionId {
+        String segmentIndex;
```
shivzone: Would this mean the cache will not be shared across even segments from the same host accessing the JVM ?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
frankgh: I agree with @shivzone about using guava, especially considering that we are already using it in this project. https://github.com/google/guava/wiki/CachesExplained

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
denalex: I don't think we have pxf-site anymore, config properties are now managed explicitly via pxf-env.sh and are passed as Java properties via -D parameter to Tomcat.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
denalex: I think that's what we decided on the meeting, even though I don't remember why transactionID only is not enough. I believe we decided no need for close() call also.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
lavjain: @shivzone @frankgh Will take a look at Guava cache to see if it meets our used case.

lavjain:
@benchristel Will move the cache code and cleanup logic to a separate class as you suggested.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
shivzone: We could use pxf-env in that case

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
shivzone: Not having to explicitly invoke close() was based on our choice to rely on expiration and inUse status. @lavjain can you recollect why transactionID woudln't do ?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -89,32 +182,98 @@ public Boolean run() throws IOException, ServletException {
...
+                release(timedProxyUGI, fragmentIndex, fragmentCount);
             }
         } else {
             // no user impersonation is configured
             chain.doFilter(request, response);
         }
     }
 
+   private TimedProxyUGI getTimedProxyUGI(String user, SegmentTransactionId session) throws IOException {
+        synchronized (session.segmentTransactionId.intern()) {
```
denalex: synchronizing on interned strings might not be a good practice (https://stackoverflow.com/questions/133988/synchronizing-on-string-objects-in-java), basically due to:
- possibilities that other logic elsewhere does it as well and creates possibility of deadlock
- intern() has cost and might require internal synchronization
- static pool of interned instances will grow and then GC will reclaim them, so unclear lifecycle of objects
It would seem preferable to use local object we have full control over

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -89,32 +188,103 @@ public Boolean run() throws IOException, ServletException {
...
+        TimedProxyUGI timedProxyUGI = delayQueue.poll();
+        while (timedProxyUGI != null) {
+            closeUGI(timedProxyUGI);
+            LOG.info(timedProxyUGI.session.toString() + " Delay Queue Size = " + delayQueue.size());
+            timedProxyUGI = delayQueue.poll();
+        }
+    }
+
+    private void closeUGI(TimedProxyUGI timedProxyUGI) {
+        synchronized (timedProxyUGI.session.segmentTransactionId.intern()) {
```
denalex: this can deadlock -- if 2 cleanup attempts happen at the same time for session A and session B -- session A comes first and gets a lock on, say, "seg1:tx1" in release() method and then comes here and tries to get a lock on session B "seg2:tx1", while session B would get that loc already and will, in turn try to get lock on session A. This might happen when new requests for both sessions arrive simultaneously after a delay when their previous UGIs have expired. Granted the probability of this is low, but non-zero, meaning this is dangerous and cannot be ignored. In other words, to prevent deadlocks, do not acquire new locks while holding other locks if another thread can do the same in reverse order, which is exactly the case here.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
lavjain: Agreed. Using the lock on interned string helped me to test the logic without any explicit lock management.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
lavjain: Good catch. The locking should only operate on one lock at a time. This could be accomplished if the clean method is called by a reaper thread.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/SegmentTransactionId.java
```diff
@@ -0,0 +1,60 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+public class SegmentTransactionId {
+
+    private Integer segmentId;
+    private String transactionId;
```
frankgh: `transactionId` is not being used anywhere, it can be removed.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.security.UserGroupInformation;
+
+public class UGICache {
+
+    private static final Log LOG = LogFactory.getLog(UGICache.class);
+    private static Map<SegmentTransactionId, TimedProxyUGI> cache = new ConcurrentHashMap<>();
+    //private static DelayQueue<TimedProxyUGI> delayQueue = new DelayQueue<>();
+    private static DelayQueue<TimedProxyUGI>[] delayQueues = new DelayQueue<>[64];
+    public static long UGI_CACHE_EXPIRY = 15 * 1 * 1000L; // 15 Minutes
```
frankgh: Can `UGI_CACHE_EXPIRY` be parameterized?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public UGICache() {
+        for (int i = 0; i < delayQueues.length; i++) {
+            delayQueues[i] = new DelayQueue<>();
```
frankgh: delayQueues is static. Every time a UGICache is instantiated, the `delayQueues` array will be initialized. We may want to initialize in a static ctor.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.security.UserGroupInformation;
+
+public class UGICache {
+
+    private static final Log LOG = LogFactory.getLog(UGICache.class);
+    private static Map<SegmentTransactionId, TimedProxyUGI> cache = new ConcurrentHashMap<>();
+    //private static DelayQueue<TimedProxyUGI> delayQueue = new DelayQueue<>();
+    private static DelayQueue<TimedProxyUGI>[] delayQueues = new DelayQueue<>[64];
```
frankgh: Are we limiting to 64 segments?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private static boolean closeUGI(TimedProxyUGI expiredProxyUGI) {
+
+        SegmentTransactionId session = expiredProxyUGI.getSession();
+        Integer segmentId = session.getSegmentId();
+        //synchronized (delayQueues[segmentId]) {
+            String fsMsg = "FileSystem for proxy user = " + expiredProxyUGI.getProxyUGI().getUserName();
+            try {
+                // The UGI object is still being used by another thread
+                if (expiredProxyUGI.inProgress.get() != 0) {
```
benchristel: Should this be `expiredProxyUGI.getCounter()`? Maybe there should be a method `isInUse()` that does the comparison to `0` too...

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+    private static DelayQueue<TimedProxyUGI>[] delayQueues = new DelayQueue<>[64];
+    public static long UGI_CACHE_EXPIRY = 15 * 1 * 1000L; // 15 Minutes
+
+    public UGICache() {
+        for (int i = 0; i < delayQueues.length; i++) {
+            delayQueues[i] = new DelayQueue<>();
+        }
+    }
+
+    public TimedProxyUGI getTimedProxyUGI(String user, SegmentTransactionId session) throws IOException {
```
benchristel: I think there's a potential session hijacking vulnerability here. If you pass a `session` that exists in the cache—for any user—the `user` param is ignored when looking up the corresponding UGI. This means that if I know a transaction ID for one of your recent transactions I can authenticate as you and use your UGI.

At a minimum, I think there should be some information in the session that is more difficult to guess than a transaction ID. I'm not sure what else needs to be done to make this secure.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+
+        SegmentTransactionId session = expiredProxyUGI.getSession();
+        Integer segmentId = session.getSegmentId();
+        //synchronized (delayQueues[segmentId]) {
+            String fsMsg = "FileSystem for proxy user = " + expiredProxyUGI.getProxyUGI().getUserName();
+            try {
+                // The UGI object is still being used by another thread
+                if (expiredProxyUGI.inProgress.get() != 0) {
+                    LOG.info(session.toString() + " Skipping close of " + fsMsg);
+                    expiredProxyUGI.resetTime(); // update time so that it doesn't expire until release updates time again
```
benchristel: Is this really the logic we want? With the current implementation, I think we're checking each UGI every 15 minutes (approximately) and if it happens to be not-in-use, we're destroying it. This means that sometimes, a UGI will live for 15+ minutes after `inProgress` goes to 0, and sometimes it will be destroyed immediately—e.g. if another query happens to cause `closeUGI` to be called right after `inProgress` goes to 0. This effectively means the cache does not have a consistent TTL.

I think it's simpler (and easier to document and debug etc.) if we start the timer for each UGI at the moment its reference count becomes 0, and if it's still 0 when the timer goes off, destroy the UGI. When we increment the reference count, we stop the timer. This ensures that UGIs will always be cached for at least 15 minutes after their last use.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,129 @@
...
+                    cache.remove(session);
+                }
+                
+                if (!expiredProxyUGI.isCleaned()) {
+                    LOG.info(session.toString() + " Closing " + fsMsg +
+                            " (Cache Size = " + cache.size() + ")");
+                    FileSystem.closeAllForUGI(expiredProxyUGI.getProxyUGI());
+                    expiredProxyUGI.setCleaned();
+                }
+
```
benchristel: Should we be returning `true` here?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -64,21 +75,28 @@ public void init(FilterConfig filterConfig) throws ServletException {
...
         if (SecureLogin.isUserImpersonationEnabled()) {
 
             // retrieve user header and make sure header is present and is not empty
-            final String user = ((HttpServletRequest) request).getHeader(USER_HEADER);
-            if (user == null) {
-                throw new IllegalArgumentException(MISSING_HEADER_ERROR);
-            } else if (user.trim().isEmpty()) {
-                throw new IllegalArgumentException(EMPTY_HEADER_ERROR);
+            final String user = getHeaderValue(request, USER_HEADER);
+            String transactionId = getHeaderValue(request, TRANSACTION_ID_HEADER);
```
benchristel: We are not passing `required=true` to `getHeaderValue()` here—is it really valid for `transactionId` to be null?

If `transactionId` can be null, that means we're going to create `SegmentTransactionId`s with nothing but a segment ID in them, and we'll potentially get the wrong UGI from the cache if another request with no `transactionId` was recently served.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/SegmentTransactionId.java
shivzone: We might need it in the future. It doesn't hurt to leave in there for now.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
lavjain: w\ @benchristel Will change `DelayQueues` to be not static.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/TimedProxyUGI.java
```diff
@@ -0,0 +1,93 @@
...
+        return unit.convert(getDelayMillis(), TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public int compareTo(Delayed other) {
+        TimedProxyUGI that = (TimedProxyUGI) other;
+        return Long.compare(this.getDelayMillis(), that.getDelayMillis());
+    }
+
+    public long getDelayMillis() {
```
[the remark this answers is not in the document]
lavjain: `Delayed` just expects you to implement `compareTo(Delayed other)`

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,154 @@
...
+        UGICacheEntry ugi = null;
+        DelayQueue<UGICacheEntry> delayQueue = getDelayQueue(segmentId);
+        while ((ugi = delayQueue.poll()) != null) {
+            // Place it back in the queue if still in use and was not closed
+            if (!closeUGI(ugi)) {
+                delayQueue.offer(ugi);
+            }
+            LOG.debug("Delay Queue Size for segment " +
+                    segmentId + " = " + delayQueue.size());
```
denalex: wrap with isDebugEnabled()

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.mockito.stubbing.Answer;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+import static org.powermock.api.mockito.PowerMockito.when;
+
+public class UGICacheTest {
```
denalex: if you use @RunWith(MockitoJUnitRunner.class) then you can use @Mock annotations, and do not have to verify if you trained the mock with specific values, not any . We discussed it with Scala, which didn't have that support.

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+        session = new SessionId(0, "txn-id", "the-user");
+
+        cache = new UGICache(provider);
+    }
+
+    @Test
+    public void getUGIFromEmptyCache() throws Exception {
+        UGICacheEntry entry = cache.getTimedProxyUGI(session);
+        assertNotNull(entry.getUGI());
+        verify(provider).createProxyUGI("the-user");
```
denalex: do we want to assert ref count and expiration time set as a result of such get ? Do we want to see if the principal name under proxy ugi matches the user passed in ?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void getSameUGITwiceUsesCache() throws Exception {
+        UGICacheEntry entry1 = cache.getTimedProxyUGI(session);
+        UGICacheEntry entry2 = cache.getTimedProxyUGI(session);
+        assertEquals(entry1, entry2);
```
denalex: assertSame for ref comparison ?

denalex: assert ref count == 2 ?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+        assertEquals(proxyUGI2a, proxyUGI2b);
+        assertNotEquals(proxyUGI1a, proxyUGI2a);
+        verify(provider, times(1)).createProxyUGI("the-user");
+        verify(provider, times(1)).createProxyUGI("different-user");
+    }
+
+    @Test
+    public void getUGIWhenRequestedUserDoesNotExist() throws Exception {
+        // what does UserGroupInformation.createProxyUser() do in this scenario?
+        // how about getLoginUser()?
```
denalex: user existence is not required for proxy users, remove the method

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void anySegmentIdIsValid() throws Exception {
+        session = new SessionId(65, "txn-id", "the-user");
```
denalex: is this relevant anymore since magic number 64 is gone now ?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+    @Test
+    public void releaseWithoutForceClean() throws Exception {
+        UGICacheEntry proxyUGI1 = cache.getTimedProxyUGI(session);
+
+        cache.release(proxyUGI1, false);
+        // UGI wasn't cleaned up, so we can still get it
+        UGICacheEntry proxyUGI2 = cache.getTimedProxyUGI(session);
+        assertEquals(proxyUGI1, proxyUGI2);
+        verify(provider, times(1)).createProxyUGI("the-user");
```
denalex: check refcount still one and new expiration date is set

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void releaseWithForceClean() throws Exception {
+        UGICacheEntry proxyUGI1 = cache.getTimedProxyUGI(session);
+
+        cache.release(proxyUGI1, true);
+        UGICacheEntry proxyUGI2 = cache.getTimedProxyUGI(session);
+        assertNotEquals(proxyUGI1, proxyUGI2);
```
denalex: hmm.. they should be equal but not same, does actual UGI ref gets into equals() method on CacheEntry ?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+        verify(provider, times(2)).createProxyUGI("the-user");
+    }
+
+    @Test
+    public void releaseAndReacquireDoesNotFreeResources() throws Exception {
+        UGICacheEntry proxyUGI1 = cache.getTimedProxyUGI(session);
+
+        cache.release(proxyUGI1, false);
+        UGICacheEntry proxyUGI2 = cache.getTimedProxyUGI(session);
+        proxyUGI2.setExpired();
```
denalex: this method should be internal to cache and not exercisable by users

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    public void releaseAnExpiredUGIResetsTheTimer() throws Exception {
+        UGICacheEntry proxyUGI1 = cache.getTimedProxyUGI(session);
+        proxyUGI1.setExpired();
+        cache.release(proxyUGI1, false);
+        UGICacheEntry proxyUGI2 = cache.getTimedProxyUGI(session);
+        assertEquals(proxyUGI2, proxyUGI1);
+        verify(provider, never()).destroy(any(UserGroupInformation.class));
```
denalex: where does it assert that the timer has been reset ?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
```diff
@@ -0,0 +1,197 @@
...
+        // retains a reference to a cache entry after releasing it.
+        UGICacheEntry proxyUGI1 = cache.getTimedProxyUGI(session);
+
+        cache.release(proxyUGI1, true);
+        cache.release(proxyUGI1, true);
+        UGICacheEntry proxyUGI2 = cache.getTimedProxyUGI(session);
+        assertNotEquals(proxyUGI1, proxyUGI2);
+        verify(provider, times(2)).createProxyUGI("the-user");
+    }
+}
```
denalex: good set of tests, I wonder if we need to include tests that use real timer (instead of explicitly setting expired) to make sure "clean-on-get" works and if we need multithreaded testing to make sure there are no deadlocks and sync is correct.

benchristel: To the first point—we can probably mock the system time instead of using real time, but I agree. Multithreaded tests would be good.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICacheEntry.java
```diff
@@ -0,0 +1,94 @@
...
+import org.apache.hadoop.security.UserGroupInformation;
+
+public class UGICacheEntry implements Delayed {
+
+    private volatile long startTime;
+    private UserGroupInformation proxyUGI;
+    private SessionId session;
+    private boolean cleaned = false;
+    private AtomicInteger inProgress = new AtomicInteger();
+    private static long UGI_CACHE_EXPIRY = 15 * 60 * 1000L; // 15 Minutes
```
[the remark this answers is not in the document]
benchristel: For testing? Or for some other reason?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
benchristel: @lavjain pointed out that the default value for `required` is `true`. Maybe it's better to be explicit, though.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -89,32 +106,49 @@ public Boolean run() throws IOException, ServletException {
...
             // no user impersonation is configured
             chain.doFilter(request, response);
         }
     }
 
+
+    private Integer getHeaderValueInt(ServletRequest request, String headerKey, boolean required)
+            throws IllegalArgumentException {
+        String value = getHeaderValue(request, headerKey, required);
+        return value != null ? Integer.valueOf(value) : null;
```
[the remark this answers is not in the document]
benchristel: which is a subclass of `IllegalArgumentException`. Should we handle the `NumberFormatException` differently?

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
benchristel: The ref count and expiration time are internal details of the cache, so rather than directly test them, we have other tests that simulate time passing, acquire multiple references etc. and assert that the UGIs are destroyed at the right time.

I believe the mock verification `verify(provider).createProxyUGI("the-user")` is as close as we can get to verifying the principals of the UGI, because we're using a mock UGI here.

pxf/pxf-service/src/test/java/org/apache/hawq/pxf/service/UGICacheTest.java
benchristel: We used this test to prove that our switch to using a `ConcurrentHashMap` instead of an array fixed the problem. Maybe it can be deleted now (or we could change the number 65 to something crazier like 2^31).

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -52,69 +62,90 @@
...
-                try {
-                    if (proxyUGI != null) {
-                        LOG.debug("Closing FileSystem for proxy user = " + proxyUGI.getUserName());
-                        FileSystem.closeAllForUGI(proxyUGI);
-                    }
-                } catch (Throwable t) {
-                    LOG.warn("Error closing FileSystem for proxy user = " + proxyUGI.getUserName());
-                }
+                // Optimization to cleanup the cache if it is the last fragment
+                boolean forceClean = (fragmentIndex != null && fragmentIndex.equals(fragmentCount));
```
shivzone: Why would fragmentIndex matter here ? Are we certain that the requests are always receieved in the order of fragment index ?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/SessionId.java
```diff
@@ -65,12 +74,15 @@ public int hashCode() {
...
-        return this.sessionId.equals(that.sessionId);
+    public boolean equals(Object obj) {
+        if (obj == null) return false;
+        if (obj == this) return true;
+        if (obj.getClass() != getClass()) return false;
+
+        SessionId that = (SessionId) obj;
+        return new EqualsBuilder()
+                .append(sessionId, that.sessionId)
+                .isEquals();
```
lavjain: EqualsBuilder is not needed for comparing strings

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -129,90 +135,93 @@ public void release(Entry timedProxyUGI, boolean forceClean) {
...
-                // updates the time again
-//                expiredProxyUGI.resetTime();
                 return false;
             }
 
             // Expired UGI object can be cleaned since it is not used
             // Determine if it can be removed from cache also
             Entry cachedUGI = cache.get(session);
-            if (expiredProxyUGI == cachedUGI) {
+            if (toDelete == cachedUGI) {
```
lavjain: Not sure if this logic is still true. The cache could be holding a reference to a newer UGI. That is why we were comparing the reference. This is because the release is also getting it from cache.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -61,48 +64,51 @@
...
+        Entry timedProxyUGI = cache.get(session);
+
+        if (timedProxyUGI == null) return;
 
-        Integer segmentId = timedProxyUGI.getSession().getSegmentId();
         timedProxyUGI.resetTime();
         timedProxyUGI.releaseReference();
-        if (forceClean) {
-            synchronized (getExpirationQueue(segmentId)) {
+        if (cleanImmediatelyIfNoRefs) {
```
lavjain: cleanImmediately here does not know if there are any refs.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -61,48 +64,51 @@
...
+     * @param session                  the session for which we want to release the UGI.
+     * @param cleanImmediatelyIfNoRefs if true, destroys the UGI for the given session (only if it is
+     *                                 now unreferenced).
      */
-    public void release(Entry timedProxyUGI, boolean forceClean) {
+    public void release(SessionId session, boolean cleanImmediatelyIfNoRefs) {
+
+        Entry timedProxyUGI = cache.get(session);
+
+        if (timedProxyUGI == null) return;
```
lavjain: Is there a case where the above condition would hold true?

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -129,90 +135,93 @@ public void release(Entry timedProxyUGI, boolean forceClean) {
...
-     * Iterate through all the entries in the queue for the given segment
+     * Iterate through all the entries in the queue for the given session's segment
      * and close expired {@link UserGroupInformation}, otherwise it resets
      * the timer for every non-expired entry.
      *
-     * @param segmentId
+     * @param session
      */
-    private void cleanup(Integer segmentId) {
+    private void cleanup(SessionId session) {
```
lavjain: The cleanup here is confusing because it is using the current session with any expired UGI

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -129,90 +135,93 @@ public void release(Entry timedProxyUGI, boolean forceClean) {
...
 
         Entry ugi;
-        DelayQueue<Entry> delayQueue = getExpirationQueue(segmentId);
+        DelayQueue<Entry> delayQueue = getExpirationQueue(session.getSegmentId());
```
lavjain: It might be better to pass the delayQueue to this method instead of doing a lookup again.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -244,10 +246,10 @@ void setCleaned() {
         }
 
         /**
-         * @return the number of active requests using the {@link UserGroupInformation}.
+         * @return true if the UGI is being referenced by a session, false otherwise
          */
-        int countReferences() {
-            return referenceCount.get();
+        private boolean isInUse() {
```
lavjain: All function modified should be of the same type as they are being used in a similar fashion.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
benchristel: We'll change this to `assert timedProxyUGI != null`.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
lavjain: For the same segment, they are always called in order.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
denalex: I'd rather not have any assert statements in Java code, if that's what you mean.

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,318 @@
...
+     * Decrement reference count for the given session's UGI. Resets the time at which the UGI will
+     * expire to 15 minutes in the future.
+     *
+     * @param session                  the session for which we want to release the UGI.
+     * @param cleanImmediatelyIfNoRefs if true, destroys the UGI for the given session (only if it is
+     *                                 now unreferenced).
+     */
+    public void release(SessionId session, boolean cleanImmediatelyIfNoRefs) {
+
+        Entry timedProxyUGI = cache.get(session);
```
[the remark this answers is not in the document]
benchristel: We are changing it to entry

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,318 @@
...
+
+    /**
+     * Decrement reference count for the given session's UGI. Resets the time at which the UGI will
+     * expire to 15 minutes in the future.
+     *
+     * @param session                  the session for which we want to release the UGI.
+     * @param cleanImmediatelyIfNoRefs if true, destroys the UGI for the given session (only if it is
+     *                                 now unreferenced).
+     */
+    public void release(SessionId session, boolean cleanImmediatelyIfNoRefs) {
```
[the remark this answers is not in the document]
benchristel: I'd like the parameter names to allow someone reading this code to consider the cache in isolation from its callers, and from that perspective, `cleanImmediatelyIfNoRefs` or some variant seems like a good choice. The cache can't know or control whether it's the "lastCall" to `release`.

What if, instead of using a boolean parameter to change the behavior of this function, we had a separate method `destroyUnreferencedUGIs` that clients could call to clean things up? Then that method could be called in the `SecurityServletFilter` and the concept of "lastCall" could be represented there.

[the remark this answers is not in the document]
denalex: hmm ... not sure that would be better. We have dilemma here -- cache should not know about caller and caller should not care about internal cache implementation, such as that cleaning is required at all, etc. Here're we're trying to cheat and give cache a bit more context (such as that this is our last call) and let it decide to clean immediately instead of waiting for it to timeout.

Either way we slice it it is not elegant. The proper way to separate concerns is perhaps registering a listener into a session that listens to close() events and does this `destroyUnreferencedUGIs` call and having a client calling session.close() when it is last call. This way client only closes session, session only calls a listener and the listener, which is part of the cache does the cleaning.

I do realize, however, this is a total overkill for the problem at hand. Whatever we call this doesn;t matter that much, the discussion itself is more interesting :)

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/UGICache.java
```diff
@@ -0,0 +1,318 @@
...
+     * @return true if the UGI entry was cleaned, false when the UGI entry was still in use
+     * and cleaning up was skipped
+     */
+    private boolean closeUGI(SessionId session, Entry toDelete) {
+        // There is no need to synchronize this method because it should be called
+        // from within a synchronized block
+        String fsMsg = "FileSystem for proxy user = " + toDelete.getUGI().getUserName();
+        try {
+            // The UGI object is still being used by another thread
+            if (toDelete.isInUse()) {
```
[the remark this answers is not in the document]
frankgh: We are renaming it to `expiredUGI`

pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/servlet/SecurityServletFilter.java
```diff
@@ -19,94 +19,114 @@
...
     private static final String USER_HEADER = "X-GP-USER";
-    private static final String MISSING_HEADER_ERROR = String.format("Header %s is missing in the request", USER_HEADER);
-    private static final String EMPTY_HEADER_ERROR = String.format("Header %s is empty in the request", USER_HEADER);
+    private static final String SEGMENT_ID_HEADER = "X-GP-SEGMENT-ID";
+    private static final String TRANSACTION_ID_HEADER = "X-GP-XID";
+    private static final String FRAGMENT_INDEX_HEADER = "X-GP-FRAGMENT-INDEX";
+    private static final String FRAGMENT_COUNT_HEADER = "X-GP-FRAGMENT-COUNT";
+    private static final String MISSING_HEADER_ERROR = "Header %s is missing in the request";
+    private static final String EMPTY_HEADER_ERROR = "Header %s is empty in the request";
+    private static UGICache proxyUGICache;
```
benchristel: @denalex @lavjain What is the lifetime of the `SecurityServletFilter` instance? As I was thinking about making this non-static, I realized that if there's more than one instance per JVM, we might get redundant caches (and possible UGI resource leaks, if a `SecurityServletFilter` gets garbage-collected before all the UGIs in its cache are cleaned up).

benchristel:
Merged to master. Closing PR...

PR #1380, opened 2018-07-05
HAWQ-1636. Fix compile apache hawq failure due to unsupported syntax in libyarn on osx 10.11

oushu1wangziming1:
Fix compile issue due to unsupported syntax in libyarn on osx 10.11


huor:
+1

PR #1381, opened 2018-07-12
Remove autogeneration of version number from pxf/build.gradle

benchristel:
This enables us to use IntelliJ's code navigation and refactoring
features. Previously, IntelliJ would navigate to / refactor the
generated files rather than the source files since the generated files
were the ones that actually got compiled.

We also upgraded Gradle to 3.0 because we suspected IntelliJ might not have good support for older Gradle versions.

Co-authored-by: Lav Jain <ljain@pivotal.io>
Co-authored-by: Ben Christel <bchristel@pivotal.io>

### DIFF
--- a/pxf/gradle/wrapper/gradle-wrapper.properties
+++ b/pxf/gradle/wrapper/gradle-wrapper.properties
@@ -1,3 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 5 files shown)


pxf/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,23 +1,6 @@
...
-# 
-# Unless required by applicable law or agreed to in writing,
-# software distributed under the License is distributed on an
-# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-# KIND, either express or implied.  See the License for the
-# specific language governing permissions and limitations
-# under the License.
-
-#Wed Aug 05 16:07:21 PDT 2015
+#Mon Jul 09 15:02:34 PDT 2018
```
denalex: lack of license will trip license checker, unless you add this file to exclusion list, even better to preserve the license.

shivzone:
Until recently the version number was hardcoded in the code. generatedSources task does seem to be more pain than benefit. I'm fine with merging this change even in the mean time until we figure out an alternate means of fetching the version programatically

benchristel:
Closing this for now, as these changes are out of date and the issue hasn't been a pain point for us recently.

PR #1382, opened 2018-07-25
Relogin from keytab when security is enabled instead of using the TOKEN

benchristel:
- We are now supporting Kerberos login with GPDB, which unlike HAWQ does
  not pass the TOKEN parameter. Thus, we re-login from the local keytab
  to authenticate the user.

Co-authored-by: Lav Jain <ljain@pivotal.io>
Co-authored-by: Ben Christel <bchristel@pivotal.io>
Co-authored-by: Shivram Mani <smani@pivotal.io>


benchristel:
This pull request is stale. @shivzone @lavjain I'm closing it. Feel free to reopen if you want.

PR #1383, opened 2018-07-31
HAWQ-1644. Make delegation token optional in PXF

shivzone:
Make delegation token property optional for PXF when used against secure hadoop. Delegation token is a complementary means for authentication with hadoop along with Kerberos. There is not much value in using the delegation token with masterless and segment only PXF architecture, as each pxf jvm will need to establish the token anyway. Simply authenticating with Keberos should be good enough in such a deployment mode.


pxf/pxf-service/src/main/java/org/apache/hawq/pxf/service/utilities/SecuredHDFS.java
```diff
@@ -45,28 +45,33 @@
...
      */
-    public static void verifyToken(ProtocolData protData, ServletContext context) {
+    public static void verifyToken(UserGroupInformation userGroupInformation,
+                                   String tokenString,
+                                   ServletContext context) {
         try {
             if (UserGroupInformation.isSecurityEnabled()) {
                 /*
-                 * HAWQ-1215: The verify token method validates that the token sent from
+                 * The verify token method validates that the token sent from
```
divyabhargov: Was removing the HAWQ reference story from the comment accidental?

PR #1397, opened 2018-09-20
HAWQ-1660. Optimize parquet scan when bloom filter enabled.

kuien:


linwen:
It is a good optimization point. If a lot of columns will be projected, we can only fetch joinkey and do a bloomfilter check, if doesn't match, no need to fetch other columns. 

But in this PR, if bloomfilter is not enable, it will fetch joinkey in the first loop, and fetch other columns in the second loop, which needs a little refine further.

src/backend/cdb/cdbparquetrowgroup.c
```diff
@@ -322,6 +306,92 @@ ParquetRowGroupReader_ScanNextTuple(
...
+	{
+		/* it is not expensive to do twice in case of bloomfilter */
+		if(projs[i] == false)
+		{
+			nulls[i] = true;
+			continue;
+		}
+
+		/* skip those attributes not in given list */
+		if (attsList != NIL && list_find_int(attsList, i) >= 0)
```
kuien: typo, should be < 0

interma:
@kuien I do a perf test on your pr, two issues:
1. query result error
2. performance downgrade

Details see below, please check code, thanks.

TPCH1G data on my mac, master code
```
tpch=# select count (*) from part, lineitem where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX';
count
-------
  6088
(1 row)

Time: 3150.873 ms
tpch=# set hawq_hashjoin_bloomfilter to on;
SET
Time: 2.903 ms
tpch=# select count (*) from part, lineitem where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX';
count
-------
  6088
(1 row)

Time: 1512.782 ms
```
 your code
```
tpch=# select count (*) from part, lineitem where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX';
 count
-------
  6088
(1 row)

Time: 49466.999 ms #<-- result ok, but bad performance
tpch=# set hawq_hashjoin_bloomfilter to on;                                                                             SET
Time: 13.106 ms
tpch=# select count (*) from part, lineitem where p_partkey = l_partkey and p_brand = 'Brand#23' and p_container = 'MED BOX';
 count
-------
     0 #<-- result error
(1 row)

Time: 1888.176 ms 
```

interma:
@kuien 
Btw: If also test on mac, you can generate tpch data via my dbgen tools:
https://github.com/interma/misc/tree/master/hawq/dbgen_mac

PR #1416, opened 2019-03-16
HAWQ-1687. Fix travis broken due to missing apr-1-config

jiny2:
Create for hour, 

need to add apr-util libarary in hawq travis


jiny2:
I cannot set assignee, reviewers etc , so weird

stanlyxiang:
+1

PR #1417, opened 2019-03-16
fix misc issues before orc format support

jiny2:
Raise this PR for hour

fix misc issues before orc format support, including typos, bugs, license


stanlyxiang:
+1

PR #1427, opened 2019-04-11
HAWQ-1697. fix bug of window func unsafe push down

Librago:
window function was pushed down to scan node, causing the plantree messed up.
fix it to choose safe expression push down.


huor:
LGTM +1

ztao1987:
+1

PR #1454, opened 2019-06-02
HAWQ-1722. Fix core dump due to lock is not released before reporting errors when exceeding MaxAORelSegFileStatus

huor:


jiny2:
+1 LGTM

ztao1987:
+1

PR #1476, opened 2019-08-15
HAWQ-1748. Idle executors keep exist instead of quit even when gp_vmem_idle_resource_timeout specified timer expires

huor:


jiny2:
+1 LGTM

ztao1987:
+1

PR #1483, opened 2019-11-07
JIRA-1753. add datalocality comments

Librago:
location number of external table should less than bucketnum of target table

description:

when load data from external table like 
insert into table from exttable.
bucket number of internal table must be less than location number. but the error message is "can not allocate memory."
This is confused to user, so change it.

issue in jira
https://issues.apache.org/jira/projects/HAWQ/issues/HAWQ-1753?filter=allopenissues


stanlyxiang:
+1 

please add JIRA num to PR title and use a more clearly commit message. It's enhancement for comments instead of code fix.

jiny2:
Please make this pr bound to the issue.

jiny2:
+1 LGTM

PR #1484, opened 2019-11-13
HAWQ-1754. add comment for max segment number count for random table.

Librago:
max segment number for random table is adjusted by limit of virtual segment number per segment. add comment for it.


Librago:
https://issues.apache.org/jira/browse/HAWQ-1754 JIRA issue

huor:
LGTM +1

PR #1485, opened 2019-11-17
HAWQ-1755. add comment for enforce virtual segment number.

Librago:
add comment for enforce virtual segment number.
which is the ultimate factor to determine virtual segment number.


Librago:
https://issues.apache.org/jira/browse/HAWQ-1755

jiny2:
LGTM +1

huor:
LGTM +1

PR #1487, opened 2019-11-21
HAWQ-2762. add comment to vseg_to_splits_map.

Librago:
key is global host Index, value is list of virtual segment id. this map is key to calculate datalocality of virtual segments.

https://issues.apache.org/jira/browse/HAWQ-1757


huor:
LGTM +1

stanlyxiang:
+1

PR #1497, opened 2020-10-20
HAWQ-1769. Fix DirectoryIterator of libhdfs

qinzuoyan:
The DirectoryIterator of libhdfs has some problems now:
* even if the `getListing` RPC returns "remainingentries = 0", the following invoking of DirectoryIterator::hasNext() may still trigger an unnecessary `getListing` RPC to the server.
* moreover, in the above case, if the server returns a non-empty list (so it is a server bug because the server should return an empty list) when the client send an unnecessary `getListing` RPC, then DirectoryIterator::hasNext() will return true, and we will enter endless loop.

### DIFF
--- a/depends/libhdfs3/src/client/DirectoryIterator.cpp
+++ b/depends/libhdfs3/src/client/DirectoryIterator.cpp
@@ -28,17 +28,17 @@
 namespace Hdfs {
 
 DirectoryIterator::DirectoryIterator() :
-    needLocations(false), filesystem(NULL), next(0) {
+    needLocations(false), hasMore(true), filesystem(NULL), next(0) {
 }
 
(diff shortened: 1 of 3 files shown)


huor (approved):

huor:
LGTM +1

ztao1987 (approved):

huor:
Great to have the fix for DirectoryIterator! The fix looks good. However, you need to follow below process to raise pull request:
1. create an issue in hawq jira
2. change your commit message to "HAWQ-XXXX. Fix DirectoryIterator" where XXXX is the jira number for the issue you created

Please let me know if you have more questions.

qinzuoyan:
Done

PR #1505, opened 2021-02-04
HAWQ-1782. Fix failed to read EXTERNAL TABLE of GPFDIST protocol

chiyang10000:
- revise CREATE EXTERNAL TABLE default to READABLE EXTERNAL TABLE

- fix read from EXTERNAL TABLE of GPFDIST protocol

The documented behaviour of CREATE EXTERNAL TABLE indicates a READABLE
TABLE but used to be violated.

Referring to https://curl.se/libcurl/c/curl_multi_fdset.html,
```
If no file descriptors are set by libcurl, max_fd will contain -1 when
this function returns. Otherwise it will contain the highest descriptor
number libcurl set. When libcurl returns -1 in max_fd, it is because
libcurl currently does something that isn't possible for your
application to monitor with a socket and unfortunately you can then not
know exactly when the current action is completed using select(). You
then need to wait a while before you proceed and call curl_multi_perform
anyway. How long to wait? Unless curl_multi_timeout gives you a lower
number, we suggest 100 milliseconds or so, but you may want to test it
out in your own particular conditions to find a suitable value.
```
it is not an error of `max_fd == -1`. Moreover, the situation of
`max_fd == -1` seems common on macOS.

### DIFF
--- a/.github/workflows/scripts/gtest_filter_negative
+++ b/.github/workflows/scripts/gtest_filter_negative
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 export GTEST_FILTER_NEGATIVE=\
-TestErrorTable.TestErrorTableAll:\
 TestCommonLib.TestHdfsConfig:\
 TestExtOrc.TestNormalPath:\
 TestExtOrc.BoolTypeTest:\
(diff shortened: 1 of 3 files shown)


huor:
+1 LGTM

PR #1510, opened 2021-03-16
HAWQ-1788: set default value for dfs.domain.socket.path

zhouyuan:
Instead of setting empty string, an example value from HDFS doc should
be better to follow.

https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-hdfs/ShortCircuitLocalReads.html

Signed-off-by: Yuan Zhou <yuan.zhou@intel.com>


zhouyuan:
@chiyang10000 thanks for the help!

